PR #12, opened 2016-07-12
Add Label Badge to Line Chart.

Adondriel:
Added an optional Label option to Line Annotation.
Updated the documentation to display this.


Chart.Annotation.js
```diff
@@ -148,14 +149,16 @@ var AnnotationPlugin = Chart.PluginBase.extend({
 		}
 	},
 
-	afterDraw: function(chartInstance, easingDecimal) {
+	beforeDraw: function(chartInstance, easingDecimal) {
```
etimberg: probably should be `beforeDatasetsDraw` per my comment in the other issue

src/line.js
```diff
@@ -1,51 +1,71 @@
 // Line Annotation implementation
-module.exports = function(Chart) {
```
etimberg: did the whitespace change here?

Adondriel: It looks like it must have... not sure why github sees it as one big change. But that line was also readded below.

src/line.js
```diff
@@ -1,51 +1,71 @@
...
+            ctx.lineTo(view.x2, view.y2);
+            ctx.stroke();
+            //Verify that label has been specified, so that we don't add a null label.
+            //TODO: Added customizable badge options, like: where to display the badge, left/right/center. or, Badge Border Color and Badge Background color. 
+            //I would also like to add an option for different directional offsets, horizontalOffset, and VertOffset for example.
+            //Current issue with background color is that it likes to disappear when you hover over any other element on the chart.
+            if (options.label) {
+                //draw the background of the badge.
+                ctx.fillStyle = 'white';
+                ctx.fillRect((chartArea.right / 2), pixel - 5, 35, 15);
```
etimberg: Sorry that this took a while. I went to merge this and do some testing and I noticed this is incorrect. The width of the fill rect should be the width of the text. The text can be measured using `ctx.measureText`. To use this effectively, the font should be configurable and would need to be set before calling `measureText`.

The height of the rect would be the font size in px.

src/line.js
```diff
@@ -1,51 +1,71 @@
...
+                //Draw the border around the badge.
+                ctx.fillStyle = view.borderColor;
+                ctx.strokeRect((chartArea.right / 2), pixel - 5, 35, 15);
+                //Draw the text of the badge.
+                ctx.fillStyle = 'black';
```
etimberg: This colour should probably be configurable as well

src/line.js
```diff
@@ -1,51 +1,71 @@
...
+                ctx.textAlign = 'left';
+                ctx.fillText(options.label ? options.label : '', (chartArea.right / 2) + 5, pixel - 5);
```
etimberg: the extra check for label being present is not necessary. How will this location work for vertical lines? There should probably be a check for that case and change the position

src/line.js
```diff
@@ -1,51 +1,71 @@
...
+    var horizontalKeyword = 'horizontal';
+    var verticalKeyword = 'vertical';
+
+    var LineAnnotation = Chart.Element.extend({
+
+        draw: function (chartInstance, options) {
+            var view = this._view;
+            var ctx = chartInstance.chart.ctx;
+            var scale = chartInstance.scales[options.scaleID];
+            var pixel = scale ? scale.getPixelForValue(options.value) : NaN;
```
etimberg: do we need to call this again? would we be able to use the `view`?

compwright:
It doesn't appear that this works correctly with vertical lines yet.

petewalker:
I've added a follow-up to this pull request that addresses the configurability issues, and the compatibility with vertical lines: #31

compwright:
@etimberg This one should be closed if you're planning to accept #31.

PR #27, opened 2016-09-29
Typo in box example

compwright:
I couldn't figure out why my box was as tall as the scale, even though I set `yMin` and `yMax`.


etimberg:
thanks @compwright

PR #39, opened 2016-11-22
Include annotation values in scale min/max range

### LINKED ISSUE #23: Chart scales are not adjusted with annotations
The automatic scale does not account for the annotations. If I set an annotation with value greater than the data scale the annotation is drawn outside the chart area.

compwright:
Resolves #23


compwright:
@etimberg Could you look over this and give me your feedback? It does have a bug that I can't seem to find - whenever a scale min or max is changed, a line annotation will not show up. If another line is added that does not require a scale adjustment, both lines then appear.

You can see this behavior in the updates.html sample.

src/plugin.js
```diff
@@ -80,19 +80,74 @@ function buildAnnotations(configs) {
...
+				var scale = chartInstance.scales[scaleId];
+				decorate(scale, 'determineDataLimits', function(previous) {
+					previous();
+					var dataMin = scale.min;
+					var dataMax = scale.max;
+					var range = getAnnotationDataLimits(scaleId, chartInstance.annotations, scale.min, scale.max);
+					scale.min = range.min;
+					scale.max = range.max;
+					if (scale.min < dataMin) {
+						console.log('Adjusted', scaleId, 'min from', dataMin, 'to', scale.min);
```
etimberg: might not be a good idea to log this in production.

src/types/box.js
```diff
@@ -21,20 +23,31 @@ module.exports = function(Chart) {
...
 			if (xScale) {
-				min = isValid(options.xMin) ? xScale.getPixelForValue(options.xMin) : chartArea.left;
-				max = isValid(options.xMax) ? xScale.getPixelForValue(options.xMax) : chartArea.right;
+				min = helpers.isValid(options.xMin) ? xScale.getPixelForValue(options.xMin) : chartArea.left;
+				max = helpers.isValid(options.xMax) ? xScale.getPixelForValue(options.xMax) : chartArea.right;
 				left = Math.min(min, max);
 				right = Math.max(min, max);
+
+				model.ranges[options.xScaleID] = {
+					min: xScale.getValueForPixel(left),
```
etimberg: I was going to comment that you could use `xScale.min` and `xScale.max` but I realized I don't think we explicitly make those public so good call doing it this way

src/plugin.js
compwright: It won't be, this is temporary.

etimberg:
@compwright i noticed that bug during update. I think you need to `decorate` in the `beforeUpdate` plugin call so that the axis updates with the correct sizes but actually configuring the annotations needs to happen in `afterUpdate` otherwise the pixel values will be wrong.

PR #43, opened 2016-11-30
Support mouse events on annotations

compwright:
Add an array of the events that should be watched to the `annotation` config object, and a corresponding `on{Event}` function to each annotation. Event callbacks are passed the event object as the first argument, and the target element is bound to `this`.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -17,6 +17,8 @@ Chart.Annotation.drawTimeOptions = {
 	beforeDatasetsDraw: DRAW_BEFORE_DATASETS
 };
 
+Chart.Annotation.Element = require('./element.js')(Chart);
+
 var annotationTypes =
 Chart.Annotation.types = {
 	line: require('./types/line.js')(Chart),
@@ -27,6 +29,7 @@ Chart.Annotation.types = {
 var annotationDefaults =
 Chart.Annotation.defaults = {
 	drawTime: DRAW_AFTER,
+	events: [],
 	annotations: []
 };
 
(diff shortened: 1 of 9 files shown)


src/types/line.js
```diff
@@ -1,36 +1,59 @@
...
+				endPixel = helpers.isValid(this.options.endValue) ? scale.getPixelForValue(this.options.endValue) : pixel;
+			}
+
+			if (isNaN(pixel)) {
+				return;
 			}
 
-			var chartArea = chartInstance.chartArea;
-			var ctx = chartInstance.chart.ctx;
+			var chartArea = this.chartInstance.chartArea;
```
etimberg: [minor] if you capture `this` into a local variable there will be better minification

src/types/box.js
```diff
@@ -2,42 +2,65 @@ var helpers = require('../helpers.js');
...
+		getArea: function() {
+			return this.getWidth() * this.getHeight();
+		},
+		setRanges: function() {
 			var model = this._model = this._model || {};
 
-			var xScale = chartInstance.scales[options.xScaleID];
-			var yScale = chartInstance.scales[options.yScaleID];
-			var chartArea = chartInstance.chartArea;
+			var xScale = this.chartInstance.scales[this.options.xScaleID];
```
etimberg: same comment about `this`

src/plugin.js
```diff
@@ -138,21 +128,41 @@ var annotationPlugin = {
 				});
 			});
 		});
+
+		// Detect and intercept events that happen on an annotation element
+		var config = chartInstance.options.annotation || {};
+		if (config.events) {
```
etimberg: [info] just a quick bit of info here. This will become a lot easier once some of the planned Chart.js v2.5 work is done. The goal is to have plugins notified about events so that you don't have to add your own listeners.

compwright: Any idea when v2.5 will be ready?

kurkle:
@compwright looks like you introduced the dblclick event here, was it requested and/or did/do you have a use case for it?

Cc @stockiNail

PR #44, opened 2016-12-01
Support hover events

compwright:
Also fixes a click event duplication when a `dblclick` event happens.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,9 +36,8 @@ To configure the annotations plugin, you can simply add new config options to yo
 		drawTime: 'afterDraw', // (default)
 
 		// Mouse events to enable on each annotation.
-		// Similar to the `events` common chart configuration option
-		// http://www.chartjs.org/docs/#chart-configuration-common-chart-configuration
-		// Should be an array of: mousemove, mousedown, mouseup, click, dblclick, contextmenu, wheel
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -124,6 +123,12 @@ Vertical or horizontal lines are supported.
 
 	// Mouse event handlers - be sure to enable the corresponding events in the
 	// annotation events array or the event handler will not be called.
+	// See https://developer.mozilla.org/en-US/docs/Web/Events for a list of
+	// supported mouse events.
+	onMouseenter: function(e) {},
```
etimberg: Would it be a good idea to support touch events as well?

src/events.js
```diff
@@ -0,0 +1,148 @@
...
+							if (typeof options[eventHandlerName] === 'function') {
+								eventHandlers.push([ options[eventHandlerName], hoverEvent, element ]);
+							}
+						});
+					}
+				});
+			}
+		}
+
+		// Suppress duplicate click events during a double click
```
etimberg: I like this comment. Very detailed :)

README.md
compwright: Eventually, yes, but it will need some care to do it right in a way that works cross-browser and also works for the user. Probably need some hit radius logic added, and integration with Hammer.js. That's more than I can get into right now.

PR #45, opened 2016-12-03
Rewrote the annotation update handler, bugfixes

compwright:
* Fixes a bug where calling `chartInstance.update()` broke the referential integrity of the annotations, preventing further updates
* Fixes a bug where the `hovering` element flag was getting reset when calling `chartInstance.update()` from inside the `onMouseover` handler
* Annotations now use an unambiguous unique ID rather than an array index to map configs to element instances
* Properly hooked in the scale adjuster using the `afterDataLimits` scale option
* Refactored the plugin object to its own file
* Refactored helpers.js to use a consistent module format
* Pointed the `main` property in `package.json` back to the `src` directory to deal with Webpack warnings
* Tighter plugin code and var names


src/index.js
```diff
@@ -0,0 +1,64 @@
+// Get the chart variable
+var Chart = require('chart.js');
+Chart = typeof(Chart) === 'function' ? Chart : window.Chart;
+
+// Configure plugin namespace
+Chart.Annotation = Chart.Annotation || {};
+
+Chart.Annotation.dblClickSpeed = 350; // ms
```
etimberg: did we document this property in the readme?

src/index.js
```diff
@@ -0,0 +1,64 @@
...
+Chart.Annotation.Element = require('./element.js')(Chart);
+
+Chart.Annotation.types = {
+	line: require('./types/line.js')(Chart),
+	box: require('./types/box.js')(Chart)
+};
+
+var helpers = require('./helpers.js')(Chart);
+
+Chart.Annotation.ScaleAdjuster = function(scale) {
```
etimberg: Is the intent of putting this at `Chart.Annotation.ScaleAdjuster` to allow it to be replaced easily by a consumer of this lib?

src/annotation.js
```diff
@@ -0,0 +1,117 @@
...
+				ns.firstRun = false;
+			}
+
+			var elementIds = [];
+
+			// Add new elements, or update existing ones
+			ns.config.annotations.forEach(function(annotation) {
+				var element;
+				if (typeof annotation._id === 'undefined' && !!annotationTypes[annotation.type]) {
+					annotation._id = helpers.objectId();
```
compwright: That's an interesting idea. I wouldn't want to be required to define it, and while I can't think of any reason why I'd want to set a specific ID, it's better to err on the side of flexibility. It would be a simple change here.

src/annotation.js
```diff
@@ -0,0 +1,117 @@
...
+					var cls = annotationTypes[annotation.type];
```
compwright: `jshint` doesn't complain. I'd rather have it lowercase, because convention is that newable _classes_ are capital. `cls` is a variable class reference, not the class definition itself. I think this subtle difference makes the code more readily understandable.

src/index.js
compwright: Not replaced necessarily (although that would be possible), but used or reused in other places. Originally did it this way because I thought that I would have to have users manually add it to their `scales` options objects, but then I found a way to add it automatically at the right time, so this doesn't need to be public anymore.

Issue #63, opened 2017-04-26
JS-Error in use with chart doughnut on the same page


PureDude:
If I render a doughnut-chart on the same page with annotation-plugin I get the following JS-Error and the chart is not shown: **TypeError: t.scales is undefined**
When I remove the annotation plugin chart is rendered. You can see it here:
https://codepen.io/PureDude/pen/JNbVqL
Any idea what this can be?

etimberg:
This is likely because the doughnut chart only has one axis and doesn't create a `Chart.scales` property. Happy to look at a PR fixing this

PureDude:
Thank you for your quick response! I don't want to use the annotations in a doughnut chart but in a line chart on the same page. So there is currently no way or workaround to use the annotation plugin in a line chart on the same page with a donut chart without annotations?

etimberg:
@PureDude if you're using Chart.js v2.5.0, you could add the plugin to only that charts that need it rather than for all charts on the page https://github.com/chartjs/Chart.js/blob/master/docs/developers/plugins.md#using-plugins

PureDude:
@etimberg Thx for this info! I know it's off topic but do you have an example for me how I can include the annotation plugin in this way? All my attempts ended with errors...

etimberg:
That's a good question. You may need a fork because the global registration happens automatically :/ @compwright thoughts?

chrisyaxley:
Any update on the fix for this? I haven't been able to get the plugin include working either.

compwright:
@PureDude can you verify if https://github.com/chartjs/chartjs-plugin-annotation/pull/69 fixes you issue?

H/T @chrisyaxley

compwright:
Duplicate of #49

PureDude:
@compwright I have tested and #69 (already included in Version 0.5.5) fixes the issue! Many thanks!

PR #50, opened 2016-12-29
Annotation plugin improvements (Fixes #3, #35, #47, #49, #51, #53, #63, #66, and #68)

yusufozturk:
Check Scales on ChartOptions to avoid null errors.

Fixes #49

PR #69, opened 2017-05-18
Adding a fix for when doughnut chart is on the page

chrisyaxley:
This seems to fix the console error for doughnut charts (#63).


compwright:
Duplicate of #50

PR #65, opened 2017-05-09
Fix chartArea not yet available when calling setDataLimits

bendem:
```
angular.js:14199 TypeError: Cannot read property 'left' of undefined
    at ChartElement.setDataLimits (https://localhost:8003/bower_components/chartjs-plugin-annotation/chartjs-plugin-annotation.js:517:91)
    at ChartElement.initialize (https://localhost:8003/bower_components/chartjs-plugin-annotation/chartjs-plugin-annotation.js:139:9)
    at ChartElement.Chart.Element (https://localhost:8003/bower_components/chart.js/dist/Chart.js:4899:19)
    at ChartElement [as constructor] (https://localhost:8003/bower_components/chart.js/dist/Chart.js:5193:14)
    at new ChartElement (https://localhost:8003/bower_components/chart.js/dist/Chart.js:5193:14)
    at https://localhost:8003/bower_components/chartjs-plugin-annotation/chartjs-plugin-annotation.js:69:20
    at Array.forEach (native)
    at Object.beforeUpdate (https://localhost:8003/bower_components/chartjs-plugin-annotation/chartjs-plugin-annotation.js:63:27)
    at Object.notify (https://localhost:8003/bower_components/chart.js/dist/Chart.js:7346:17)
    at Chart.Controller.update (https://localhost:8003/bower_components/chart.js/dist/Chart.js:4136:16)
```

PR #80, opened 2017-08-02
Fix error on creating box annotation with missing min or max param

patotoma:
You can see the issue here: https://codepen.io/anon/pen/EvggZy.

Basically when creating an annotation like this:
```
{
            drawTime: "beforeDatasetsDraw",
            type: "box",
            yScaleID: "y-axis-0",
            yMin: 0,
            // yMax: -10, <------ uncomment this to make it work!
            backgroundColor: "rgba(101, 33, 171, 0.5)",
}
```

without specifying for example the `yMax` parameter, it throws an error on missing chartInstance object. I understand that there may be another problem why chartInstance is missing but this fixes the error and makes it work.

PS: I've already merged PR #65 in this one but needed to update the code position. Thanks

### DIFF
--- a/src/types/box.js
+++ b/src/types/box.js
@@ -12,12 +12,12 @@ module.exports = function(Chart) {
 			var yScale = chartInstance.scales[options.yScaleID];
 			var chartArea = chartInstance.chartArea;
 
+			// Set the data range for this annotation
+			model.ranges = {};
+			
 			if (!chartArea) {
 				return;
 			}
-
-			// Set the data range for this annotation
-			model.ranges = {};
 			
 			var min = 0;
 			var max = 0;


patotoma:
Hello @compwright, I don't think it is necessary to apply this to **line annotation** too because unlike in **box**, `chartArea` is not accessed in `setDataLimits` of `LineAnnotation`.

compwright:
Good point! Thanks

BorysIatsun:
Hello @compwright, when you plan to release this fix and update in npm? It would help much to me in issue I'm facing. Thanks

compwright:
@patotoma @boiat Just released v0.5.7

boiat:
@compwright,
 Great, thank you

PR #81, opened 2017-08-03
Added typings and updated  package.json types 

nik-john:
Added typings and updated  package.json types as per latest Typescript specs

Please refer to https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html for more details


styfle:
Where is the `.d.ts` file? I don't see it.

styfle:
I don't see any content in the `index.d.ts`. Should this be part of the build step: generate the `.d.ts` file?

PR #90, opened 2017-10-18
Create bower.json

oscarotero:
Please, accept this PR. We need a bower.json file in order to load the compiled javascript file instead the src.
Thanks in advance!

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,6 @@
+{
+  "name": "chartjs-plugin-annotation",
+  "description": "Annotation plugin for Chart.js",
+  "license": "MIT",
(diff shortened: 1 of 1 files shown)


styfle:
I thought bower was deprecated. They wrote a blog post about [migrating away from bower](https://bower.io/blog/2017/how-to-migrate-away-from-bower/).

compwright:
@styfle it is deprecated, and I would hesitate to add but it sounds like from the blog post the migration plugin would still need a bower.json so that the right file gets loaded.

oscarotero:
Version is not needed in the bower.json, because it uses the tag (this is the only thing I think bower did better than npm). I've added the author.

PR #93, opened 2017-11-06
More flexible positioning of boxes

MickMickMickMick:
Allows to configure box coordinates with a two tuple array. The
box will be drawn at the average pixel position. This allows the
more flexible placement of boxes in a category scale.


kurkle:
This would need to be rebased, doubt you are interested after 3 years :)

PR #107, opened 2018-01-30
eslint fixes

benmccann:
These are fixes suggested by eslint. There was too much to make all the changes and enable eslint all in one PR. Will send a separate PR to enable eslint

The `case` statement formatting seems potentially funny to me, but otherwise the style rules it suggested make sense

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -8,7 +8,9 @@ module.exports = function(Chart) {
 
 	function setAfterDataLimitsHook(axisOptions) {
 		helpers.decorate(axisOptions, 'afterDataLimits', function(previous, scale) {
-			if (previous) previous(scale);
+			if (previous) {
+				previous(scale);
+			}
 			helpers.adjustScaleRange(scale);
 		});
 	}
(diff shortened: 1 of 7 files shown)


simonbrunel:
do you mean: `case view.mode === horizontalKeyword && view.labelPosition === 'right':`?

benmccann:
I mean that before the `case` statements were indented one level from the `switch` whereas eslint wants them not to be indented. It made more sense to me to have them be indented

simonbrunel:
IMO, this `switch-case` indent level is useless because it increases the line length without making the code more readable since your code will be indented under each `case`.

```javascript
switch (foo) {
	case 0:
		switch (bar) {
			case 0:
				// ...
				break;
			default:
				// ...
				break;
		}
		break;
	default:
		// ...
		break;
}

// - vs -

switch (foo) {
case 0:
	switch (bar) {
	case 0:
		// ...
		break;
	default:
		// ...
		break;
	}
	break;
default:
	// ...
	break;
}
```

benmccann:
I don't have strong feelings. Just wanted to make sure it was something we'd considered and we weren't using a default that we were all against.

Would you be able to merge this PR for me? Thanks!

PR #109, opened 2018-01-31
Pass the index of value datapoint to getPixelValue

holmari:
If the value is not a data object, the value will be undefined, so
current behavior is retained. This fixes an issue where the line would be
positioned on the first element with the given datapoint's value, which
is not necessarily the case if the axis values repeat.

An example situation where the bug would show up would be a case
where you want to create a scrub line using a vertical line. If the x axis
is labelled only with e.g. "MMM DD", a multi-year chart would have repeating
values and the scrub line would be always incorrectly positioned due
to this missing index.


holmari:
Thanks for the review @arnasledev. Could this be merged in? I realized there were a couple of jshint issues, which I fixed just now.

PR #122, opened 2018-05-25
Avoid 'Unable to preventDefault inside passive event listener invocation.' message

anxorl:
Avoid 'Unable to preventDefault inside passive event listener invocation.' message.

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -78,7 +78,6 @@ module.exports = function(Chart) {
 					options.onClick.call(element, e);
 				}, dblClickSpeed);
 				e.stopImmediatePropagation();
-				e.preventDefault();
 				return;
 			} else if (e.type === 'dblclick' && element.clickTimeout) {
 				clearTimeout(element.clickTimeout);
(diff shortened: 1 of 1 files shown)


Techn1x:
I am getting this as well..

benmccann:
@compwright it looks like this was added in https://github.com/chartjs/chartjs-plugin-annotation/commit/5d14e2632cc956e588a90a5a9ea14d8f171fea2e. Do you know if it's safe to remove?

KumarNarendra:
I am also getting this. My suggestion is, Instead of removing it, we should add `{ passive: false }` to the event.

benmccann:
@KumarNarendra I haven't spent a lot of time understanding this code and haven't worked with passive event listeners before. I agree that `{ passive: false }` would also fix the error, but I'm not sure why one solution would be better than the other. Can you explain why that solution might be preferable?

benmccann:
An alternative fix has been proposed in https://github.com/chartjs/chartjs-plugin-annotation/pull/160. Please take a look and provide feedback as to which is the better approach

kurkle:
The alternative fix was merged

Issue #114, opened 2018-03-09
Read options from dedicated plugins section


alllex:
According to [Chart.js docs](http://www.chartjs.org/docs/latest/developers/plugins.html#plugin-options) all plugins should read options from dedicated `options.plugins.<plugin_name>` section. However, this plugin reads its options straight from `options.<plugin_name>`.

This makes it difficult to work with the plugin from TypeScript where in [declared types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/chart.js/index.d.ts) it is not reasonable to add options per plugin. There is, however, a convenient `plugins?: any;`.

I guess it is not possible to stop reading options from `options.annotation` due to compatibility issues. One of possible workarounds is to check if `options.plugins.annotation` is present and **also** read from there.

KayHS:
Im also having trouble using @types/chart.js with this plugin now.

` error TS2345: Argument of type '{ type: string; data: { labels: string[]; datasets: { fill: true; lineTension: number; pointRadiu...' is not assignable to parameter of type 'ChartConfiguration'.
  Types of property 'options' are incompatible.
    Type '{ plugins: { datalabels: { display: boolean; }; }; legend: { display: false; }; annotation: { ann...' is not assignable to type 'ChartOptions'.
      Object literal may only specify known properties, but 'annotation' does not exist in type 'ChartOptions'. Did you mean to write 'rotation'?`

sfentress:
Any chance of getting a release bump, so we can pull this in with npm?

benmccann:
I don't think I have permission to `npm publish`. Perhaps you could ask someone in the #dev slack channel to cut a release

KayHS:
Has this been fixed?

PR #130, opened 2018-08-03
Read plugin options from plugins.annotation

holmari:
As a fallback, read the options directly from options.annotation, which
is the old behavior.

Fixes https://github.com/chartjs/chartjs-plugin-annotation/issues/114


KayHS:
Hey when will this be added?

benmccann:
@holmari I just noticed that it looks like there are a couple places that read the options directly only from the old place instead of using `ns.options` which contains the correct options:
https://github.com/chartjs/chartjs-plugin-annotation/blob/master/src/annotation.js#L21
https://github.com/chartjs/chartjs-plugin-annotation/blob/master/src/annotation.js#L113

Would you be able to take a look?

Issue #101, opened 2017-12-05
Suggestion: partial line annotations for bar charts


LeoDupont:
Hello,

I made a small Chart.js plugin which draws "partial lines" (not full-height or full-width)  on a horizontal bar chart (easily adaptable to vertical bars). Here is a sample: 

![chrome_2017-12-05_12-00-51](https://user-images.githubusercontent.com/29914053/33603782-08253ada-d9b4-11e7-9168-ea6978ff6a5a.png)

I'm just sharing it in case you think it could be included in **chartjs-plugin-annotation**, in which case we could start working on adapting it to your guidelines.

The plugin:

```javascript
const verticalLinePlugin = {
	renderVerticalLine: function(chartInstance, line) {
		
		let datasetMeta = chartInstance.getDatasetMeta(0);
		let context = chartInstance.chart.ctx;

		// === Coordinates ===
		
		// y-range (depending on which bar to draw the line for) : 
		let datasetModel = datasetMeta.data[line.dataIndex]._model;
		const yBarCenter = datasetModel.y
		const barHeight = datasetModel.height;
		const yStart = yBarCenter - (barHeight / 2) - 4; // `4` offset could be customizable
		const yEnd = yBarCenter + (barHeight / 2) + 4;

		// x-coordinate (line's value) :
		const xAxisID = datasetMeta.xAxisID;
		const xCoordinate = chartInstance.scales[xAxisID].getPixelForValue(line.xCoordinate);
		
		// === Drawing the line ===
		// (could be more customizable: dashed line, width, etc...)
		
		context.beginPath();
		context.strokeStyle = line.color;
		context.moveTo(xCoordinate, yStart);
		context.lineTo(xCoordinate, yEnd);
		context.stroke();

		// === Writing label under line's end ===
		// (could be customizable)

		if (typeof line.label !== typeof undefined) {
			context.textAlign = 'center';
			context.fillStyle = line.color;
			context.fillText(line.label, xCoordinate, yEnd + 7);
		}
	},

	// I could not find a better working hook... 
	// this one's get called multiple times in a row, which is not ideal
	afterDatasetsDraw: function(chart, easing) { 
		if (chart.data.datasets.length < 1) {
			// No dataset, no line :
			return;
		}
		if (chart.config.lines) {
			chart.config.lines.map(line => {
				this.renderVerticalLine(chart, line);
			});
		}
	}
};

Chart.plugins.register(verticalLinePlugin);
```

Usage: 

[...]

PR #132, opened 2018-09-24
Introduce ability to annotate a single bar

lukasbischof:
This PR introduces the ability to annotate only a single bar (implementing #101):
image

It also supports horizontal bars:
image

The PR adds two new options to the line annotation: 
* `onlyForDataIndex` which specifies the data index which should be annotated
* `linePadding` which adds a customisable padding to the start and end of the line
* `datasetIndex` which specifies on which dataset the data should be annotated


sammeel:
Hi, I came across your answer as I was searching for the same thing. I noticed that your label is always aligned to the graph, not to the line. Is this the intended behavior? For a line across the whole chart, this is the intended behavior, but I would have expected the label to be aligned to line.

For example a center aligned label for dataindex 0 gives me this chart:
image

lukasbischof:
Hi, thanks for your input. However, I'm not able to reproduce the problem. Centering should actually be supported and if I try to reproduce, it yields this result:
image

Could you point out how exactly you received this behaviour and where variables on window are set?

sammeel:
mmm, I must have done something wrong. I just tried reproducing it again, and now it works fine. 

My appologies :)

lukasbischof:
@ygwain Actually it should be already like that. Could you provide your code to reproduce?

README.md
```diff
@@ -100,6 +100,17 @@ Vertical or horizontal lines are supported.
 	// https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineDashOffset
 	borderDashOffset: 5,
 
+	// Only draw a line for a specific data at the specified index. Currently only supported for bar graphs
```
benmccann: data is plural, but it sounds like this option applies only to a single element, so it might be better to say "specific data element" instead

benmccann:
I haven't looks super closely at this PR, but it seems quite specific. Would https://github.com/chartjs/chartjs-plugin-annotation/pull/141 also allow you to accomplish the same thing (in perhaps a more general manner)?

faraonkicu:
Hello, when you are planning to merge this feature?

Obleo33:
I would love to be able to use this feature. Please merge soon.

benmccann:
I would rather merge #141 instead of this PR unless there's a reason that approach won't work. Please see my comment just above https://github.com/chartjs/chartjs-plugin-annotation/pull/132#issuecomment-457332787

Obleo33:
> I would rather merge #141 instead of this PR unless there's a reason that approach won't work. Please see my comment just above [#132 (comment)](https://github.com/chartjs/chartjs-plugin-annotation/pull/132#issuecomment-457332787)

I am a bit new to all this and this solution looked a bit easier to implement. If you think [#141](https://github.com/chartjs/chartjs-plugin-annotation/pull/141) can perform the same function I am open to learning.

lukasbischof:
I've just looked briefly at #141 and I guess the main difference between this PR is that it allows you to annotate just one single bar at a specific index in a specific dataset while #141 hides an entire annotation.

benmccann:
I went back and read through the thread at https://github.com/chartjs/chartjs-plugin-annotation/pull/141 again. The author of that PR and I had decided to a `hoverColor` option would be a better solution for what he's trying to accomplish. That's fairly different from the original problem being discussed, so sorry for the confusion. Let's consider these PRs separately

benmccann (dismissed):
I think this is too specific to the bar chart. Could you pass a start and end tick and then use `scale.getPixelForTick` instead of using the bar controller and model?

kurkle (approved):

kurkle (dismissed):

kurkle (approved):

stockiNail:
@kurkle I have seen that these new features are not available on version 1 of the plugin.
Make sense to think implementing them in version1 as well?

kurkle:
There is a discussion about this: https://github.com/chartjs/chartjs-plugin-annotation/discussions/484

I would not like to add bar specific configuration when its quite easily achieved already.

kurkle:
But samples would be good.

stockiNail:
@kurkle thank you ! I haven't seen the discussion.

stockiNail:
> But samples would be good.

I think makes sense. Do you already a list of additional samples (apart this one) you'd like to have in the samples doc section?

kurkle:
> > But samples would be good.
> 
> I think makes sense. Do you already a list of additional samples (apart this one) you'd like to have in the samples doc section?

No, I don't have a list

Issue #95, opened 2017-11-10
Hide/Show annotations?


Techn1x:
Hi, is there any way to hide & show the individual annotations with a property? (Rather than removing the object from the `annotations` array?)

For example, Chart.JS allows you to define a `display`: true/false/callback on things like the legend

Thanks!

corentingi:
I am looking for the same thing
To actually emulate this, it is needed to delete and recreate the annotations...

compwright:
Not currently. If you'd like to take a crack at it, we'll entertain a PR! Thanks!

Techn1x:
I might take a look at doing a PR at some point. Any guidance?

I've taken a look at the chartjs-plugin-datalabels project which uses a `display` attribute (there's a global one, and a per-data one), all of the logic for it is in this file https://github.com/chartjs/chartjs-plugin-datalabels/blob/a56a4d1bf733976507205046d189d5fd6c4f3005/src/plugin.js

Just having trouble envisioning how I might apply the same to this project

Hopefully when I take a deeper look later it might become more apparent to me

yusufozturk:
@Techn1x  @corentingi 
You can make this by adding a "display" option into your annotation. Or you can hide them entirely, or you can just hide annotations individually.

I'm going to make a pull request for this. But for now, you can make following changes.

First, add a new option called "display" into default options:

![image](https://user-images.githubusercontent.com/928600/34516201-57a5bc3c-f076-11e7-9659-1019b5e57297.png)

Then add "canHide" option to check them individually:

![image](https://user-images.githubusercontent.com/928600/34516241-7f8cfb98-f076-11e7-88a9-a58dc4784740.png)

So if you add "canHide: true" into your annotation and if you leave "display: false", then these annotations will not be visible.

How can you enable/disable visibility of the annotations? Just enable "display" option by setting it to "display: true/false" and then update your chart. Like in example:

`chart.options.annotation.display = true;
chart.update();`

That's it :)

yusufozturk:
Or another alternative:

![image](https://user-images.githubusercontent.com/928600/34517210-660333be-f07a-11e7-9c61-c89f2a0de224.png)

So you can also hide them individually by setting "isHidden:true" option. I used "isHidden" instead of "display" so nobody will have issue after version upgrade. Otherwise everybody has to update their annotation config which is not very pleasant.

GevaertB:
will this be in the next release?

Kibo007:
[...]

PR #141, opened 2018-11-22
Add hoverColor option

Techn1x:
Discussed here: https://github.com/chartjs/chartjs-plugin-annotation/issues/95

`display` is set to true by default. Setting to false will hide either all annotations, or the individual annotation. Hiding the annotation essentially just makes it invisible, so that events will still trigger (useful for hover events, for example)


Techn1x:
Thanks, fixed now

benmccann:
@Techn1x this PR will need to be rebased

amiedavis:
Hi, I'm about to add chartjs-plugin-annotation to a project and I would like this functionality (I need to conditionally add a line depending on where the chart is rendered from), is this PR likely to be merged soon?

src/helpers.js
```diff
@@ -91,8 +91,9 @@ module.exports = function(Chart) {
 	function initConfig(config) {
 		config = chartHelpers.configMerge(Chart.Annotation.defaults, config);
 		if (chartHelpers.isArray(config.annotations)) {
-			config.annotations.forEach(function(annotation) {
+			config.annotations = config.annotations.map(function(annotation) {
 				annotation.label = chartHelpers.configMerge(Chart.Annotation.labelDefaults, annotation.label);
+				return chartHelpers.configMerge(Chart.Annotation.annotationDefaults, annotation);
```
benmccann: `configMerge` is being deprecated https://github.com/chartjs/Chart.js/pull/6022. Can you switch these to just use `helpers.merge` instead, which is more appropriate here

benmccann:
It seems unintuitive to me that you can hover over a hidden object to make it appear. I think it'd be confusing if things just started magically appearing. How would you know where you're supposed to hover?

Techn1x:
I use it to highlight intervals in a Band Chart, without needing to use a billion ticks

untitled project
(ignore the slowness in the gif, it's due to the RDP session I was recording through, not the chart itself)

What about something like...
`display: none/hidden/visible`

Techn1x:
@amiedavis Conditionally adding and removing annotations can currently be done by adding/removing them from the annotations array, it doesn't necessarily need this PR

benmccann:
That feature is pretty cool, but I think it's probably a bit outside the scope of the plugin, so I'm going to close this PR for now

benmccann:
To clarify, I'm fine with the display option, but I think if an annotation is hidden it should not fire events. We can reopen if you're interested in pursuing that

Techn1x:
Fair enough. So we're agreed that at least a `display: true/false` (either shows or removes annotation completely) is a good idea

With regards to hidden annotations, would you support either of these approaches?
1. `transparent: true/false` (annotation is always in chart, but if true then all colours used in the annnotation are switched to transparent)
OR
2. `hoverColor: rgba()` etc. (each color-related attribute has a related hover attr, which is the color used when the annotation is hovered)

Setting an annotation to transparent could be done externally by setting each color, but is tedious if managing several annotations which could have different colors to remember & reapply each time you want to change the transparency

I think approach 2 is interesting because it could allow implementations other than my own, and is somewhat consistent with chart.js

benmccann:
`hoverColor` sounds fine to me

I'll reopen this PR and you can decide whether to update this one or start a new one. Feel free to close it again in the latter case

benmccann:
@Techn1x will you have a chance to update this PR?

src/helpers.js
matthewhegarty: I've picked up making this change, but switching to `helpers.merge()` causes the annotation not to render, presumably because there is specific logic in `configMerge()` for handling scale.
Is there any guidance / best practice / examples on how to migrate from `configMerge()` to `helpers.merge()`?

kurkle (approved):

Issue #140, opened 2018-11-22
onMouseleave and onMouseout not called when mousing onto different annotation


Techn1x:
I have a lot of annotations on my chart, and I noticed that mouseout/mouseleave events do not occur until ALL annotations are moused away from. Once the mouse leaves all annotations, then all of the mouseout/mouseleave events occur immediately

Techn1x:
The cause is this code here https://github.com/chartjs/chartjs-plugin-annotation/blob/master/src/events.js#L49-L64 which only causes the hover-off code to trigger when zero elements are found at the cursor location (i.e. `if (!element)...`)

The solution is to remember the last hovered element, then trigger the hover-off logic for it, when a new element is hovered (i.e. the code above what I linked)

Working on a PR for it now

vemuez:
Yup, I'm having this exact same issue. An example of this: https://jsfiddle.net/cb84xonk/
I've increased the vertical padding to ensure some overlap happens to get the idea across. 

You'll notice if you move back and forth between the same annotations, the triggers do not re-invoke but resets after you leave the annotations as a whole

kurkle:
This was fixed in #239

PR #142, opened 2018-11-23
Hover off events now fire when new annotation is hovered

Techn1x:
Fixes https://github.com/chartjs/chartjs-plugin-annotation/issues/140

Added the `applyHover` method, which is just a refactor really.
The key to the fix is this part;
```javascript
// end hover on the last hovered element
if(lastHoveredElement && element !== lastHoveredElement) {
  applyHover(lastHoveredElement,e,eventHandlers,false);
}
```


benmccann:
@Techn1x can you rebase this PR?

benmccann:
@Techn1x just a reminder that this PR will need to be rebased

src/events.js
```diff
@@ -37,29 +65,16 @@ module.exports = function(Chart) {
 		// Detect hover events
 		if (e.type === 'mousemove') {
 			if (element && !element.hovering) {
+				// end hover on the last hovered element
+				if(lastHoveredElement && element !== lastHoveredElement) {
```
benmccann: the formatting here doesn't quite match what's expected. please run `gulp lint` after rebasing to check whether there are any style issues

src/events.js
```diff
@@ -2,6 +2,7 @@ module.exports = function(Chart) {
 	/* eslint-disable global-require */
 	var chartHelpers = Chart.helpers;
 	var helpers = require('./helpers.js')(Chart);
+	var lastHoveredElement;
```
benmccann: please move this down a line below the `eslint global-require` comments

Techn1x:
@benmccann Rebased and fixed lint issues, should be good to go

src/events.js
```diff
@@ -25,6 +26,33 @@ module.exports = function(Chart) {
 		return filteredEvents;
 	}
 
+	function applyHover(element, e, eventHandlers, beginHover) {
```
benmccann: I think it might be slightly cleaner to break this into two functions: one for starting hover and one for ending it

src/events.js
```diff
@@ -25,6 +26,38 @@ module.exports = function(Chart) {
...
 	}
 
+	function startHover(element, e, eventHandlers) {
+		var options = element.options;
+		if (!element.hovering) {
+			// fire hover events
+			['mouseenter', 'mouseover'].forEach(function(eventName) {
+				var handlerName = helpers.getEventHandlerName(eventName);
+				var hoverEvent = helpers.createMouseEvent(eventName, e); // recreate the event to match the handler
+				element.hovering = true;
```
benmccann: For `endHover` the lines setting `element.hovering` and `lastHovered` are outside the `forEach`. Is it possible to do the same here for consistency?

benmccann:
Thanks for your patience! This looks good. I just have one last comment

src/events.js
Techn1x: I think the reason I did it that way (this was back in November) was because that's how the original code was (looking at the diff, it had `element.hovering = true;` inside the `forEach` when starting hover, and then outside the `forEach` for ending the hover)

I can't immediately see why it would be that way, so I'll make the change and test it when I get some time and then get back to you

Techn1x: Since there aren't any tests in this library, back in November I was probably just trying to refactor code without messing with the logic more than I needed to

kurkle (approved):
lgtm

PR #153, opened 2019-02-23
Fix how-to-use guide and put the "annotation" inside options

saeedesmaili:
The "annotation" should be inside the "options", not "plugins", as same as the doc's codepen example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,8 @@ Annotations work with line, bar, scatter and bubble charts that use linear, loga
 To configure the annotations plugin, you can simply add new config options to your chart config.
 
 ```javascript
-{
-	plugins: {
+
+	options: {
 		annotation: {
 			// Defines when the annotations are drawn.
 			// This allows positioning of the annotation relative to the other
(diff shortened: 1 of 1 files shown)


benmccann:
It looks like the code in master does read from `options.plugins`, but hasn't been released.

They really should be under `options.plugins`: https://www.chartjs.org/docs/latest/developers/plugins.html#plugin-options

So really the best thing to do is to cut a new release instead. However, I'm not quite sure that it's completely working, so we should probably fix that first: https://github.com/chartjs/chartjs-plugin-annotation/pull/130#issuecomment-474411821

rwwagner90:
Is there any chance that new release will be coming soon?

benmccann:
You could ask on the #dev Slack channel. I've actually never used this plugin and don't have permission to cut a new release

rwwagner90:
@benmccann I am not in the Slack. Could you please ping whoever is in charge of releases?

kurkle:
The correct place for it is in `options.plugins`. This is fixed in the master (Chart.js v3) branch.

Issue #159, opened 2019-05-02
[BUG]: Unable to preventDefault inside passive event listener invocation


xr0master:
Events are Mouseenter and Mouseleave
Chrome 73
osx
Chartjs 2.8

PR #160, opened 2019-05-02
fixed prevent default

xr0master:
Avoid 'Unable to preventDefault inside passive event listener invocation.' message.
Support compatibility with old browsers

Closes https://github.com/chartjs/chartjs-plugin-annotation/issues/159 https://github.com/chartjs/chartjs-plugin-annotation/pull/122

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -167,6 +167,24 @@ module.exports = function(Chart) {
 			})
 			.slice(0, 1)[0]; // return only the top item
 	}
+	/**
+	 * Copied from https://github.com/chartjs/Chart.js/blob/4f722ab619be071e90dd9ea67db27032dc9edd2a/src/platforms/platform.dom.js#L109
+	 */
+	var supportsEventListenerOptions = (function() {
+		var supports = false;
+		try {
+			var options = Object.defineProperty({}, 'passive', {
+				// eslint-disable-next-line getter-return
+				get: function() {
+					supports = true;
+				}
+			});
+			window.addEventListener('e', null, options);
+		} catch (e) {
+			// continue regardless of error
+		}
+		return supports;
+	}());
 
 	return {
 		initConfig: initConfig,
@@ -179,7 +197,8 @@ module.exports = function(Chart) {
 		adjustScaleRange: adjustScaleRange,
 		getNearestItems: getNearestItems,
 		getEventHandlerName: getEventHandlerName,
-		createMouseEvent: createMouseEvent
+		createMouseEvent: createMouseEvent,
+		supportsEventListenerOptions: supportsEventListenerOptions
 	};
 };
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 
 .DS_Store
+.idea/
 
 node_modules/*
 custom/*
--- a/src/events.js
+++ b/src/events.js
@@ -95,7 +95,11 @@ module.exports = function(Chart) {
 
 		if (eventHandlers.length > 0) {
 			e.stopImmediatePropagation();
-			e.preventDefault();
+
+			if (!helpers.supportsEventListenerOptions) {
+				e.preventDefault();
+			}
+
 			eventHandlers.forEach(function(eventHandler) {
 				// [handler, event, element]
 				eventHandler[0].call(eventHandler[2], eventHandler[1]);


xr0master:
@benmccann Used the same logic as Chart.js

benmccann:
@xr0master can you point me to where Chart.js does this logic? I didn't see it

xr0master:
@benmccann https://github.com/chartjs/Chart.js/blob/master/src/platforms/platform.dom.js#L127

benmccann:
Could we call [helpers.addEvent](https://github.com/chartjs/Chart.js/blob/master/src/platforms/platform.dom.js#L457) instead of copying the code into this repo?

xr0master:
Wut? You do it in your code already. This plugin creates the eventHandlers in the dispatcher method and binds it to chart by helpers.addEvent.

src/helpers.js
```diff
@@ -168,6 +168,22 @@ module.exports = function(Chart) {
 			.slice(0, 1)[0]; // return only the top item
 	}
 
+	var doesSupportPassiveEvent = (function() {
```
benmccann: let's call this `supportsEventListenerOptions` for consistency with the main repo

maybe also add a comment:
```
/**
 * Copied from https://github.com/chartjs/Chart.js/blob/4f722ab619be071e90dd9ea67db27032dc9edd2a/src/platforms/platform.dom.js#L109
 */
```

xr0master:
The issue is that Chart.js uses passive events if possible. Therefore, we check if it is possible too, then we don't call the preventDefault method. You cannot call this method for a passive event, this is the meaning of being passive.

benmccann:
Yeah, sorry, looked too quickly. This looks good to me. Just one minor comment. If you don't mind updating quickly I can merge this

src/helpers.js
xr0master: OK, no problem. Just I like to do question variable for boolean value

benmccann:
@xr0master I noticed there's another pending PR open for this. Do you have thoughts about @KumarNarendra's alternate suggested fix in https://github.com/chartjs/chartjs-plugin-annotation/pull/122#issuecomment-469690185 ?

xr0master:
His suggestion is to reduce performance to fix the useless call of preventDefault method. But how can you do it? The event listeners are in Chart.js code.

benmccann:
What's the reason for calling `e.preventDefault()` in old browsers in this PR vs never doing it like https://github.com/chartjs/chartjs-plugin-annotation/pull/122 is proposing?

xr0master:
God knows why the author of this plugin uses the preventDefault method. Perhaps that was the reason. You need to decide which solution you like more.

PS My opinion is that this plugin should be completely rewritten for new solutions. Maybe when the Chart.js new version 3 comes out.

laurynasn:
Are there any plans to release either this or #160 in the near future?

Borzenko:
@benmccann are you going to release this fix?

stockiNail:
@benmccann I have the same issue.

benmccann:
I'm not really the maintainer for this plugin and it's not clear to me what the right fix is. I've merged a few PRs just to clear out the review queues on some of the plugins, but I didn't author this plugin, don't use it, and am not particularly interested in taking it over. Even if I did merge one of these PRs, I don't have permission to cut new releases. If there's anyone with a history of contributing to Chart.js projects that wants to contribute, please let us know in the #dev channel on Slack

stockiNail:
@benmccann apologize. I didn't want to push you to find out a solution.
I have added a post just to track this PR.
Of course, the expectation is that someone will take care and fix. But I'm also aware that it's not so easy.

It's difficult for me to support the development of this project due to my poor knowledge of javascript.

Nevertheless I have tested the sample setting as comments the llines ([here](https://github.com/chartjs/chartjs-plugin-annotation/blob/7037eea71314613a8af83d2298241841cbc34c4f/src/events.js#L83) and [here](https://github.com/chartjs/chartjs-plugin-annotation/blob/7037eea71314613a8af83d2298241841cbc34c4f/src/events.js#L98)) where `preventDefault()` is called and I can confirm that  it works without any message.

I dont' know if it can help

loganknecht:
Hello! I've looked at the PR and this seems sane. Is there a chance this can be merged? I understand that there isn't really a maintainer for this, but it would be nice to have this merged to fix this bug.

If not, no problem - it is a bit of a strange issue!

Ribeiro-Tiago:
I would imagine this PR also closes https://github.com/chartjs/chartjs-plugin-annotation/issues/179

kurkle (approved):

PR #161, opened 2019-05-13
Draw labels on top of lines. Fixes  #94.

### LINKED ISSUE #94: label text is overwritten by annotation lines
We're adding vertical "date markers" to our charts. It's all going well apart from where the markers are close enough for the lines to overlap the labels.

In these cases the line cuts through the labels - is there any way to bring the labels in front of the lines?

krusynth:
I've split out the label from the line drawing, and added a second loop for this phase of drawing.

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -32,7 +32,6 @@ module.exports = function(Chart) {
 				element.transition(easingDecimal).draw();
 			});
 			elements.forEach(function(element) {
-				element.configure();
 				element.transition(easingDecimal).drawLabel();
 			});
 		};
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -21,14 +21,20 @@ module.exports = function(Chart) {
 		return function(chartInstance, easingDecimal) {
 			var defaultDrawTime = chartInstance.annotation.options.drawTime;
 
-			helpers.elements(chartInstance)
+			var elements = helpers.elements(chartInstance)
 				.filter(function(element) {
 					return drawTime === (element.options.drawTime || defaultDrawTime);
-				})
-				.forEach(function(element) {
-					element.configure();
-					element.transition(easingDecimal).draw();
 				});
+
+			// Draw our box/line first, then the labels, so the labels will be on top.
+			elements.forEach(function(element) {
+				element.configure();
+				element.transition(easingDecimal).draw();
+			});
+			elements.forEach(function(element) {
+				element.configure();
+				element.transition(easingDecimal).drawLabel();
+			});
 		};
 	}
 
--- a/src/types/box.js
+++ b/src/types/box.js
@@ -141,6 +141,9 @@ module.exports = function(Chart) {
 			ctx.strokeRect(view.left, view.top, width, height);
 
 			ctx.restore();
+		},
+		drawLabel: function() {
+			// Box doesn't draw labels yet, but we add this for consistency of interfaces.
 		}
 	});
 
--- a/src/types/line.js
+++ b/src/types/line.js
@@ -243,6 +243,29 @@ module.exports = function(Chart) {
 			ctx.lineTo(view.x2, view.y2);
 			ctx.stroke();
 
+			ctx.restore();
+		},
+		drawLabel: function() {
+			var view = this._view;
+			var ctx = this.chartInstance.chart.ctx;
+
+			if (!view.clip) {
+				return;
+			}
+
+			// Canvas setup
+			ctx.beginPath();
+			ctx.rect(view.clip.x1, view.clip.y1, view.clip.x2 - view.clip.x1, view.clip.y2 - view.clip.y1);
+			ctx.clip();
+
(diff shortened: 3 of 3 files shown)


src/annotation.js
```diff
@@ -21,14 +21,20 @@ module.exports = function(Chart) {
...
-					element.transition(easingDecimal).draw();
 				});
+
+			// Draw our box/line first, then the labels, so the labels will be on top.
+			elements.forEach(function(element) {
+				element.configure();
+				element.transition(easingDecimal).draw();
+			});
+			elements.forEach(function(element) {
+				element.configure();
```
benmccann: do we need to call `configure` a second time?

krues8dr: 🤔 Good question! I don't know?

src/types/line.js
```diff
@@ -243,6 +243,29 @@ module.exports = function(Chart) {
 			ctx.lineTo(view.x2, view.y2);
 			ctx.stroke();
 
+			ctx.restore();
+		},
+		drawLabel: function() {
```
benmccann: Is the label also being drawn in `draw` and then we're drawing it again here on top a second time? It looks like there's a lot of code duplication here

benmccann:
Chart.js 2.9.0 will have a concept of z-index: https://github.com/chartjs/Chart.js/pull/6262. It would be nice to make this utilize the same concept

src/types/line.js
krues8dr: I believe I have all of the label code pulled into `drawLabel.` However, there is some setup duplication as I'm not 100% confident in what is needed for drawing given that we're doing this in two passes (that will be broken up by a separate drawing operation, potentially resetting our values).

benmccann: At the very least I'd rather have `draw` call `drawLabel` rather than duplicate the code. But even that seems a little redundant to me

What values are you worried about being reset? I've never actually used this plugin, but just help out a bit in getting the code reviewed

krues8dr:
Given the way callbacks are currently being handled, I'm not sure that refactoring the entire draw loop is in scope for this issue - you'd have to delegate all drawing back to the chart controller.

src/types/line.js
krues8dr: That would re-introduce the original issue here. The problem is that lines and labels are currently drawn at the same time, in order - so any following annotation is drawn on top of the previous ones. We need a separate loop so we draw all lines first, and all labels second, so lines are not atop labels. Since we're iterating twice, all of our values (position, size etc.) will be stomped in `ctx` by further iterations of the loop.

benmccann: There's no reason to copy/paste code though. You could call `drawLabel` from `draw` and still call `drawLabel` a second time from another for loop afterwards as you're doing now.

krues8dr: Then you're just duplicating the draw action to no benefit.

benmccann: I don't understand. Aren't you already duplicating it by copy/pasting all the code? The benefit would be drawing the labels on top of the lines...

krues8dr: The duplication is because you need the same context at a different iteration of the loop where the context is lost.  You could create a new setup function and use it both times to mutate the context but that seems like a lot of extra overhead and passing of state for little benefit.  DRY seems less important than readability in this case to me.

benmccann: Resolving this comment. I got a big confused by the way GitHub was displaying the diff originally.

src/types/box.js
```diff
@@ -141,6 +141,9 @@ module.exports = function(Chart) {
 			ctx.strokeRect(view.left, view.top, width, height);
 
 			ctx.restore();
+		},
+		drawLabel: function() {
```
benmccann: Maybe cleaner to instead check `if (element.drawLabel)` in `annotation.js` instead?

krues8dr: Labels for boxes is [on the roadmap](https://github.com/chartjs/chartjs-plugin-annotation#to-do-items), so this frames it out for that.

src/annotation.js
benmccann: Ok. Can you test if it works without the duplicate calls here? Let's remove if possible

benmccann:
@kurkle you might be interested in this one given your work in the main repo

src/annotation.js
krues8dr: Confirmed, `draw()` doesn't mutate the state of `model` as used in `configure`.  Updated in 44a5432f483c3b707b16dc5e5471ff64a0aa0e70

krues8dr:
Hold off on this please. I've just noticed a side-effect where axis values aren't being redrawn when datasets are toggle on or off.

krues8dr:
Ok, phew, fixed that.

kurkle:
This is a step towards layers, looks quite good. Should work without the 2nd transition.

firebird631:
Hi, I except the same enhancement. Is it still in progress ?

PR #170, opened 2019-07-29
fix onDestroy of undefined in react when accessing annotation

jonassunandar:
Hello!

I love your project, but I encountered an error ``TypeError: Cannot read property 'onDestroy' of undefined`` in my React project when added annotation in my chart state.

I solved it by simply checking if current chartInstance is exist, then it can pop it.

Btw this is my first pull request :D please go easy on me.

Thank you !

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -123,7 +123,11 @@ module.exports = function(Chart) {
 			}
 		},
 		destroy: function(chartInstance) {
-			var deregisterers = chartInstance.annotation.onDestroy;
+			var deregisterers; 
+			if (chartInstance)
+				deregisterers= chartInstance.annotation.onDestroy;
+			else return;
+			
 			while (deregisterers.length > 0) {
 				deregisterers.pop()();
 			}
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -123,10 +123,14 @@ module.exports = function(Chart) {
 			}
 		},
 		destroy: function(chartInstance) {
-			var deregisterers; 
-			if (chartInstance){
-				deregisterers = chartInstance.annotation.onDestroy;
-			}else{
+			var deregisterers;
+			if (chartInstance) {
+				if (chartInstance.annotation) {
+					deregisterers = chartInstance.annotation.onDestroy;
+				} else {
+					return;
+				}
+			} else {
 				return;
 			}
 			while (deregisterers.length > 0) {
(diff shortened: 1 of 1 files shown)


src/annotation.js
```diff
@@ -123,7 +123,16 @@ module.exports = function(Chart) {
 			}
 		},
 		destroy: function(chartInstance) {
-			var deregisterers = chartInstance.annotation.onDestroy;
+			var deregisterers;
+			if (chartInstance) {
```
benmccann: It would be cleaner to write this as:
```
if (!chartInstance || !chartInstance.annotation) {
  return;
}
deregisterers = chartInstance.annotation.onDestroy;
```

jonassunandar:
Thank you for fast response !

To be honest, I have no idea why. I only tried to make a horizontal line as explained [here](https://stackoverflow.com/a/42799913/6503216)

In my code, I only use it once in my react state

```
import {annotation} from 'chartjs-plugin-annotation'

...
state:{

portofolioChartOptions: {
   responsive: true,
      maintainAspectRatio: false,
      annotation: { //draw horizontal line
        annotations: [{
          type: 'line',
          mode: 'horizontal',
          scaleID: 'y-axis-0',
          value: 105,
          borderColor: 'rgb(75, 192, 192)',
          borderWidth: 4,
          label: {
            enabled: false,
            content: 'Initial Portofolio'
          }
        }]
      },
      tooltips: {
        mode: 'index',
        intersect: true
      },
      scales: {
        xAxes: [
          {
            ticks: {
              autoSkip: true,
              maxTicksLimit: 10,
            },
          },
        ],
        yAxes: [{
          ticks: {
              beginAtZero: true
          }
        }]
      },
    }
```
I used it here
```
<Portofolio data={this.state.portofolioChartData} options={this.state.portofolioChartOptions} />
```

And here is my Portofolio component
```
import { Line } from 'react-chartjs-2';
export class Portofolio extends React.Component{
    constructor(props){
        super(props);
    }
    render(){
        return(
            <div>
                <Line data={this.props.data} options={this.props.options}/>
            </div>
        )
    }
}

```

Do you know the reason?

jonassunandar:
I think the problem lies in hot loading. I clearly saw ``TypeError: Cannot read property 'onDestroy' of undefined``. But when I delete node_modules and run ``` npm install ``` again without changing anything in your package, it run smoothly.

Sorry for taking your time. Cheers!

Svetomechc:
@jonassunandar btw,  deleting node_modules and running npm install did not solve this exact issue for me. But I'm running on Angular though.

PR #178, opened 2019-08-26
Add rotation option for line labels

philippemilink:
Hi,

I added a parameter `rotate` for line labels. As it is named, this parameter allows to rotate the label. I needed this feature to display the label of a vertical line along the line (ie the label is vertical too).

I updated the sample _bar_ to add a demo of this feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,10 @@ Vertical or horizontal lines are supported.
 		enabled: false,
 
 		// Text to display in label - default is null. Provide an array to display values on a new line
-		content: "Test label"
+		content: "Test label",
+
+		// Rotation of label, in degrees, default is 0
+		rotation: 90
 	},
 
 	// Mouse event handlers - be sure to enable the corresponding events in the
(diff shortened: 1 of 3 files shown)


joshvillbrandt:
This is exactly what I needed! I'd love this to be merged and released. Thanks @philippemilink and @benmccann.

benmccann:
I merged it, but don't have permission to cut a release

joshvillbrandt:
Thanks for the quick response, @benmccann!!!

Who can we politely pester to push out a release for us? :)

benmccann:
It looks like @compwright is the only person who's ever cut a release in the past. Perhaps @etimberg would be able to help if @compwright is unavailable

etimberg:
@benmccann do we know what the release procedure is on this repo? Looks different than elsewhere as there is no release branch. Looks like everything is manual at the moment. Would be best to try and move to the system from the main repo where the built files are not checked in at all

benmccann:
Yeah, it looks like it was done manually to me. I'm fine with removing the built files from the repo.

We might be able to copy the release scripts from one of the other plugins. E.g. https://github.com/chartjs/chartjs-adapter-luxon/blob/master/scripts/release.sh

Issue #213, opened 2020-04-18
Most of the Samples seem to be broken 


robin22d:
Most of the sample seem to be missing:

- randomScalingFactor()

- window.chartColors

Add line function does not work in updates.html

robin22d:
https://github.com/chartjs/chartjs-plugin-annotation/pull/214

PR #214, opened 2020-04-18
Samples fixes

robin22d:
### Issue
Most of the Samples seem to be broken #213

### Fixes

- Added window.chartColors

- Added randomScalingFactor function

- updates.html addLine function now updates the lines useing window.myScatter.options.annotation.annotations instead of line


kurkle:
These are now fetched from `chartjs.org/samples/master`.

PR #241, opened 2020-10-27
Fix destroying annotations and beforeDatasetsDraw

joshkel:
JS Maps use `delete`, not `remove`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/delete

The code mistakenly used beforeDatasetDraw (singular), which caused the annotations to be drawn for every dataset, instead of being drawn once before all datasets.

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -24,7 +24,7 @@ export default {
 		updateElements(chart, options, args.mode);
 	},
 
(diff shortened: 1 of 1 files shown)


kurkle (approved):

kurkle:
Thanks @joshkel

PR #239, opened 2020-10-24
Align event hooks with to datalabels plugin

kurkle:
TODO: 

- [x] add migration guide

PR #242, opened 2020-10-27
Fixes for demos

joshkel:
The mouseover event didn't work, because Chart.js doesn't fire mouseover events by default. If I understand correctly, mouseover isn't really what we want; that's fired once when the mouse moves over the chart as a whole, but we need to evaluate whether it moves over individual annotation elements.

Change the demo line value from 120 to 10. 120 is usually (always?) outside of the range of the randomly generated data, so it wasn't visible.

Accessing `element.chartInstance.chart.canvas` failed, but accessing `element.chartInstance.canvas` works.


kurkle:
These things are also fixed in #239 (among other things)

joshkel:
Thank you.

Other than the fix for the "a another test label" typo in samples/labels.html, it looks like this PR is no longer needed. Would you like for me to close it?

kurkle:
Fixed the label there too.
This might be good against v2 branch though. There are couple of issues there too to fix, before pushing e release.

kurkle:
ok, does not look like it would be worth it to rebase.

joshkel:
Sounds good. Thanks for the quick feedback and for your work to get this plugin updated for Chart.js 3.x.

PR #246, opened 2020-10-30
`display` setting for annotation

### LINKED ISSUE #244: Hiding annotations along with datasets
I have some annotations that are closely tied with particular datasets, so I'd like to hide them whenever the user hides the corresponding dataset by clicking on the dataset's legend item.

I can think of a couple of possible ways of doing this:

1. Use a legend onClick handler (similar to Chart.js's [multi series pie](https://www.chartjs.org/samples/master/charts/multi-series-pie.html) example) to programmatically show and hide the annotation items. Unfortunately, there's currently no direct way of doing that (#221). (Temporarily setting the colors to transparent then restoring them when done would probably work, but it's awkward.)
2. Give users the ability to specify an optional datasetIndex property on each annotation object (similar to the datasetIndex properties that Chart.js maintains for its legend items and tooltip items); then chartjs-plugin-annotation can use that to check if the corresponding dataset is visible and skip it if not.

Would you be interested in a PR implementing either a direct way to hide annotations (via an `enabled` or `display` or `hidden` boolean property) or implementing the datasetIndex approach? The datasetIndex approach makes a lot of sense to me, but if that's too special-purpose to want to support in chartjs-plugin-annotation, then that's fine.

Thank you.

stockiNail:
Fixes #244.

Adds display property to the annotation element options (both box and line). The default is `true`.

```javascript
plugins: {
   annotation: {
      annotations: [{
         type: 'line',
         display: false,  // <-- new option, default is true
         scaleID: 'y',
         value: 10
      }]
   }
}
```

It can also set by a callback in order to hide the annotation using a logic related to chart or datasets status.

```javascript
plugins: {
   annotation: {
      annotations: [{
         type: 'line',
         display(chart, options) {
            // your logic, for instance the annoatation is shown 
            // only if a dataset (in this case the first one) is not hidden
            return chart.isDatasetVisible(0);
         },
         scaleID: 'y',
         value: 10
      }]
   }
}
```

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -76,13 +76,6 @@ const directUpdater = {
 	update: Object.assign
 };
 
-function resolveDisplay(state, options) {
-	const annotations = options.annotations || [];
-	annotations.forEach(annotation => {
-		console.log(annotation);
-	});
-}
-
 function resolveAnimations(chart, animOpts, mode) {
 	if (mode === 'reset' || mode === 'none' || mode === 'resize') {
 		return directUpdater;
@@ -99,7 +92,7 @@ function updateElements(chart, state, options, mode) {
 	const annotations = options.annotations || [];
 	const count = annotations.length;
 	const start = elements.length;
-	
+
 	if (start < count) {
 		const add = count - start;
 		elements.splice(start, 0, ...new Array(add));
@@ -114,8 +107,8 @@ function updateElements(chart, state, options, mode) {
 			el = elements[i] = new elType();
 		}
 		const display = typeof annotation.display === 'function' ? callCallback(annotation.display, [chart, annotation], this) : valueOrDefault(annotation.display, true);
-		el._display = typeof display === 'boolean' ? display : false;
-		
+		el._display = !!display;
+
 		const properties = calculateElementProperties(chart, annotation, elType.defaults);
 		animations.update(el, properties);
 	}
@@ -182,7 +175,7 @@ function draw(chart, options, caller) {
 	clipArea(ctx, chartArea);
 	for (let i = 0; i < elements.length; i++) {
 		const el = elements[i];
-		if ((el.options.drawTime || options.drawTime || caller) === caller && el._display) {
+		if (el._display && (el.options.drawTime || options.drawTime || caller) === caller) {
 			el.draw(ctx);
 		}
 	}
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -248,4 +248,3 @@ function getScaleLimits(scale, annotations) {
 	});
 	return {min, max};
 }
-
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -1,5 +1,5 @@
 import {Animations} from 'chart.js';
-import {clipArea, unclipArea, isFinite, merge, valueOrDefault} from 'chart.js/helpers';
+import {clipArea, unclipArea, isFinite, merge, valueOrDefault, callback as callCallback} from 'chart.js/helpers';
 import {handleEvent, updateListeners} from './events';
 import BoxAnnotation from './types/box';
 import LineAnnotation from './types/line';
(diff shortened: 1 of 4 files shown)


src/annotation.js
```diff
@@ -25,15 +25,17 @@ export default {
...
 	beforeUpdate(chart, args, options) {
+		const annotationOptions = options || args;
 		if (!args.mode) {
-			bindAfterDataLimits(chart, options);
+			bindAfterDataLimits(chart, annotationOptions);
 		}
 	},
 
 	afterUpdate(chart, args, options) {
+		const annotationOptions = options || args;
```
kurkle, line 35: why the `options || args`?

src/annotation.js
```diff
@@ -74,6 +76,13 @@ const directUpdater = {
 	update: Object.assign
 };
 
+function resolveDisplay(state, options) {
```
kurkle, line 79: looks like this function is unused

src/annotation.js
```diff
@@ -90,7 +99,7 @@ function updateElements(chart, state, options, mode) {
 	const annotations = options.annotations || [];
 	const count = annotations.length;
 	const start = elements.length;
-
+
```
kurkle, line 102: extra whitespace

src/annotation.js
```diff
@@ -104,6 +113,9 @@ function updateElements(chart, state, options, mode) {
 		if (!el || !(el instanceof elType)) {
 			el = elements[i] = new elType();
 		}
+		const display = typeof annotation.display === 'function' ? callCallback(annotation.display, [chart, annotation], this) : valueOrDefault(annotation.display, true);
+		el._display = typeof display === 'boolean' ? display : false;
```
kurkle, line 117: could do `el._display = !!display` to force boolean

src/annotation.js
```diff
@@ -104,6 +113,9 @@ function updateElements(chart, state, options, mode) {
 		if (!el || !(el instanceof elType)) {
 			el = elements[i] = new elType();
 		}
+		const display = typeof annotation.display === 'function' ? callCallback(annotation.display, [chart, annotation], this) : valueOrDefault(annotation.display, true);
```
kurkle, line 116: I'm working on getting core handle these kinds of things, so plugin development is simpler and the interface would be consistent between plugins etc. But its work in progress so this is fine for now.

src/annotation.js
```diff
@@ -170,7 +182,7 @@ function draw(chart, options, caller) {
 	clipArea(ctx, chartArea);
 	for (let i = 0; i < elements.length; i++) {
 		const el = elements[i];
-		if ((el.options.drawTime || options.drawTime || caller) === caller) {
+		if ((el.options.drawTime || options.drawTime || caller) === caller && el._display) {
```
kurkle, line 185: Should probably check the `_display` first, might save a cpu cycle there :)

src/annotation.js
stockiNail: because if I use options (therefore the third argument) Ive got the following exception (using the samples):`

```javascript
Uncaught TypeError: Cannot read property 'annotations' of undefined
    at getAnnotationOptions (chartjs-plugin-annotation.js:636)
    at adjustScaleRange (chartjs-plugin-annotation.js:640)
    at LinearScale.scale.afterDataLimits (chartjs-plugin-annotation.js:626)
    at LinearScale.update (chart.js:5459)
    at fitBoxes (chart.js:2839)
    at Object.update (chart.js:2951)
    at Chart._updateLayout (chart.js:7123)
    at Chart.update (chart.js:7104)
    at new Chart (chart.js:6881)
    at window.onload (bar.html:67)
```
I think this is related to this PR https://github.com/chartjs/Chart.js/pull/7949 where `args` argument has been added but I don't think it's included into chart.js version in node_modules.

```html
<script src="../node_modules/chart.js/dist/chart.js"></script>
```

In fact, if you are using CHART.JS dist/master it works.

```html
<script src="https://www.chartjs.org/dist/master/chart.min.js"></script>
```

src/annotation.js
stockiNail: @kurkle I have some doubts about the following code, to compute the display options (annotation.,js):

```javascript
const display = typeof annotation.display === 'function' ? callCallback(annotation.display, [chart, annotation], this) : valueOrDefault(annotation.display, true);
el._display = typeof display === 'boolean' ? display : false;
```

I don't know if I could write it better. Maybe there are some other better ways to do it. Otherwise, that's fine!

src/annotation.js
stockiNail: forgot to remove it....

src/annotation.js
kurkle: Right, I've been using it with master (`npm link` in chart.js and `npm link chart.js` in plugin to get master in node_moduled)

src/annotation.js
stockiNail: pushed updated PR

src/annotation.js
stockiNail: To be removed when new CHART.JS will be published

kurkle (approved):

PR #248, opened 2020-10-31
Remove annotations array, using an object instead and min/max on value and endValue options

### LINKED ISSUE #247: Unbale to set default annotations
Due to the annotations is an Array, the default options can not be merged correctly,  see https://github.com/chartjs/Chart.js/issues/7754.
Therefore even if you set annotation into the default annotations array, you can not retrieve it at chart level.

I would suggest to remove the usage of array into configuration and going to an object, where every key id the annotation id, like the following:

```javascript
{
    plugins: {
        annotation: {
            drawTime: 'afterDatasetsDraw', 
            dblClickSpeed: 350, 
            enter: function(context) {
            },
            annotations: {
                myLine: { // <----------------- annotation id, key of "annotations" object
                    drawTime: 'afterDraw',
                    type: 'line',
                    mode: 'horizontal',
                    scaleID: 'y',
                    value: '25',
                    borderColor: 'red',
                    borderWidth: 2,
                    click: function(context) {
                    }
                }
            }
        }
    }
}
```

stockiNail:
Fixes #247 and #249

The annotations node of setting is not longer an array  but now all annotations are managed as properties of an object where their key is their ID as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,26 +41,28 @@ To configure the annotations plugin, you can simply add new config options to yo
                 // context: {chart, element}
             },
 
-            // Array of annotation configuration objects
+            // Object with the annotation configuration objects, one for each key
             // See below for detailed descriptions of the annotation options
-            annotations: [{
-                drawTime: 'afterDraw', // overrides annotation.drawTime if set
-                type: 'line',
-                mode: 'horizontal',
-                scaleID: 'y',
-                value: '25',
-                borderColor: 'red',
-                borderWidth: 2,
+            annotations: {
+                myLine: {
+                    drawTime: 'afterDraw', // overrides annotation.drawTime if set
+                    type: 'line',
+                    mode: 'horizontal',
+                    scaleID: 'y',
+                    value: '25',
+                    borderColor: 'red',
+                    borderWidth: 2,
 
-                // Event hooks can be defined per annotation (`click` below) or
-                // at chart level (`enter` above). If defined in both places,
-                // the one defined per annotation takes precedence.
+                    // Event hooks can be defined per annotation (`click` below) or
+                    // at chart level (`enter` above). If defined in both places,
+                    // the one defined per annotation takes precedence.
 
-                // Fires when the user clicks this annotation on the chart
-                click: function(context) {
-                    // context: {chart, element}
+                    // Fires when the user clicks this annotation on the chart
+                    click: function(context) {
+                       // context: {chart, element}
+                    }
                 }
-            }]
+            }
         }
     }
 }
--- a/README.md
+++ b/README.md
@@ -41,10 +41,10 @@ To configure the annotations plugin, you can simply add new config options to yo
                 // context: {chart, element}
             },
 
-            // Object with the annotation configuration objects, one for each key
+            // Array of annotation configuration objects 
+            // or Object with the annotation configuration objects, one for each key
             // See below for detailed descriptions of the annotation options
-            annotations: {
-                myLine: {
+            annotations: [{
                     drawTime: 'afterDraw', // overrides annotation.drawTime if set
                     type: 'line',
                     mode: 'horizontal',
@@ -61,8 +61,7 @@ To configure the annotations plugin, you can simply add new config options to yo
                     click: function(context) {
                        // context: {chart, element}
                     }
-                }
-            }
+             }]
         }
     }
 }
--- a/README.md
+++ b/README.md
@@ -45,23 +45,23 @@ To configure the annotations plugin, you can simply add new config options to yo
             // or Object with the annotation configuration objects, one for each key
             // See below for detailed descriptions of the annotation options
             annotations: [{
-                    drawTime: 'afterDraw', // overrides annotation.drawTime if set
-                    type: 'line',
-                    mode: 'horizontal',
-                    scaleID: 'y',
-                    value: '25',
-                    borderColor: 'red',
-                    borderWidth: 2,
+                drawTime: 'afterDraw', // overrides annotation.drawTime if set
+                type: 'line',
+                mode: 'horizontal',
+                scaleID: 'y',
+                value: '25',
+                borderColor: 'red',
+                borderWidth: 2,
 
-                    // Event hooks can be defined per annotation (`click` below) or
-                    // at chart level (`enter` above). If defined in both places,
-                    // the one defined per annotation takes precedence.
+                // Event hooks can be defined per annotation (`click` below) or
+                // at chart level (`enter` above). If defined in both places,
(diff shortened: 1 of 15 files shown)


samples/updates.html
```diff
@@ -151,19 +153,19 @@
 		};
 
 		function addLine() {
-			window.myScatter.options.plugins.annotation.annotations.push({
+			window.myScatter.options.plugins.annotation.annotations.myHorizontlLine = {
```
kurkle, line 156: This does not work as it used to

src/annotation.js
```diff
@@ -26,6 +26,19 @@ export default {
...
+		if (!!annotationOptions.annotations) {
+			Object.keys(annotationOptions.annotations).forEach(key => {
+				let value = annotationOptions.annotations[key];
+				if (isObject(value)) {
+					value.id = key;
+					array.push(value);
+				}
+			});
+		}
+		annotationOptions.annotations = array;
```
kurkle, line 40: Still internally an array, so we could as well support both ways?
What is the ID used for? (I did not need it in my use cases)

The array is easier, when you need to dynamically add new annotations (like the updates example).

kurkle (commented):
Looks good, but can we have it both ways (array or object)?

samples/updates.html
stockiNail: I have changed the logic, my fault.

I'm gonna change building a id based on a counter.

src/annotation.js
stockiNail: > Still internally an array, so we could as well support both ways?

If you use the array (both defaults and chart), the config is not merged, you have only the chart, if I'm not wrong.
If you use the array on defaults and a object in a chart (or viceversa), you have only the chart settings.
Sounds not doable.

> The array is easier, when you need to dynamically add new annotations (like the updates example).

Yes, you're right, it's easier. Nevertheless using a counter you can do it (done on the updates sample) but it's not possible to use the defaults with arrays, not doable. 
I don't know what's better. In my use cases, I didn't have any need to add an annotation at runtime and if needed, I'd create an id. But, that's only my personal opinion.

stockiNail: Anyway, let me check. Because I think we can implement both ways, with limitation that the defaults are not merged.

stockiNail: Now you can manage the annotations both by an Array (as it was before) and by an Object.
The default is an object instead of an array. ok?.

stockiNail:
@kurkle  In last push, you can use an array (if provided) or an object. The plugin creates an internal array, anyway. In this way whoever wants to use the settings as array, can do it with the current constraint that can not use the defaults settings. 
If you need to set defaults, in order to create a common annotations that they can be overriden in the chart options, you should use the object.

The default of annotations options is an object.
In the README I left the sample with the array, specifying that you can use an object. Maybe the doc on README should be changed, adding more info...

stockiNail:
@kurkle  `breaking change` label can be removed because it works with the previous settings as well.

src/annotation.js
```diff
@@ -62,7 +75,7 @@ export default {
 	defaults: {
 		drawTime: 'afterDatasetsDraw',
 		dblClickSpeed: 350, // ms
-		annotations: [],
+		annotations: {},
```
kurkle, line 78: I think this makes it a breaking change still

kurkle (commented):
> If you need to set defaults, in order to create a common annotations that they can be overriden in the chart options, you should use the object.

How exactly are you using this?
Are you aware that you can use `defaults.elements.lineAnnotation` and `defaults.elements.boxAnnotation` for element specific defaults?

src/annotation.js
stockiNail: > 
> 
> I think this makes it a breaking change still

You're right! Sorry

stockiNail: > 
> 
> I think this makes it a breaking change still

Do you think it's an issue?

kurkle: It's ok. Someone will file an bug, if it breaks something that we don't want to break. Will do a beta release at some point.

README.md
```diff
@@ -41,26 +41,27 @@ To configure the annotations plugin, you can simply add new config options to yo
...
-                // Event hooks can be defined per annotation (`click` below) or
-                // at chart level (`enter` above). If defined in both places,
-                // the one defined per annotation takes precedence.
-
-                // Fires when the user clicks this annotation on the chart
-                click: function(context) {
-                    // context: {chart, element}
-                }
-            }]
+                    drawTime: 'afterDraw', // overrides annotation.drawTime if set
```
kurkle, line 48: Some new spaces here, do you thing those are needed?

src/annotation.js
```diff
@@ -26,6 +26,19 @@ export default {
...
+		if (isObject(annotationOptions.annotations)) {
+			const array = new Array();
+			Object.keys(annotationOptions.annotations).forEach(key => {
+				let value = annotationOptions.annotations[key];
+				if (isObject(value)) {
+					value.id = key;
+					array.push(value);
+				}
+			});
+			annotationOptions.annotations = array;
```
kurkle, line 39: This modification here requires that the provided options object is not the one that user provided.
I think its currently that way and will probably stay like that (because its merged with defaults).

BTW, the options || args can be removed, since beta.5 of chart.js is out?

kurkle (commented):
Some NITs, I'll wait your comments on these before merging.

stockiNail:
> 
> 
> > If you need to set defaults, in order to create a common annotations that they can be overriden in the chart options, you should use the object.
> 
> How exactly are you using this?
> Are you aware that you can use `defaults.elements.lineAnnotation` and `defaults.elements.boxAnnotation` for element specific defaults?

No, I wasn't... Usually I use defaults.plugins.annotation. In that case , you could sets the defaults on elements nevertheless this could be only 1, and if you need more lines or boxes, you have the same constraint, I guess

README.md
stockiNail: updating...

src/annotation.js
stockiNail: > BTW, the options || args can be removed, since beta.5 of chart.js is out?

Yes, I think it could be removed. Nevertheless I'm not familiar with npm packages... To set BETA5 I think it could be easy, I'm gonna change the files, but how to update the node modules? just `npm update`?

kurkle: That should do it. But no need to do that in this PR.

kurkle (approved):

src/annotation.js
stockiNail: > 
> 
> That should do it. But no need to do that in this PR.

OK! I don't do it

PR #254, opened 2020-11-02
Enable Auto rotation of labels (#250)

### LINKED ISSUE #250: Automatic rotation of labels
This is an enhancement about the possibility to rotate automatically the labels using the angle of the line when is oblique.

The `rotation` options can set a fixed rotation for label but it's quite difficult to set a right value when you want that the label is following the direction of line, because you don't know the angle of the line in advance.

The proposal is to add new boolean options to label element, called `autoRotation`, which is ignored if the `rotation` options has been or if the line is not oblique.

Now, without `rotation` and oblique lines, with following label configuration:

```javascript
label: {
    backgroundColor: 'red',
    content: 'Test Label',
    enabled: true
,
```

labelRotationNow

With the parameter  `autoRotation` :

```javascript
label: {
    autoRotation: true, // <--- set auto Rotation
    backgroundColor: 'red',
    content: 'Test Label',
    enabled: true
,
```

labelRotationNext

I'm gonna submit a PR (**already ready**, just to submit), if agreed.

stockiNail:
Fixes #250

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,26 +41,28 @@ To configure the annotations plugin, you can simply add new config options to yo
                 // context: {chart, element}
             },
 
-            // Array of annotation configuration objects
+            // Object with the annotation configuration objects, one for each key
             // See below for detailed descriptions of the annotation options
-            annotations: [{
-                drawTime: 'afterDraw', // overrides annotation.drawTime if set
-                type: 'line',
-                mode: 'horizontal',
-                scaleID: 'y',
-                value: '25',
-                borderColor: 'red',
-                borderWidth: 2,
+            annotations: {
+                myLine: {
+                    drawTime: 'afterDraw', // overrides annotation.drawTime if set
+                    type: 'line',
+                    mode: 'horizontal',
+                    scaleID: 'y',
+                    value: '25',
+                    borderColor: 'red',
+                    borderWidth: 2,
 
-                // Event hooks can be defined per annotation (`click` below) or
-                // at chart level (`enter` above). If defined in both places,
-                // the one defined per annotation takes precedence.
+                    // Event hooks can be defined per annotation (`click` below) or
+                    // at chart level (`enter` above). If defined in both places,
+                    // the one defined per annotation takes precedence.
 
-                // Fires when the user clicks this annotation on the chart
-                click: function(context) {
-                    // context: {chart, element}
+                    // Fires when the user clicks this annotation on the chart
+                    click: function(context) {
+                       // context: {chart, element}
+                    }
                 }
-            }]
+            }
         }
     }
 }
--- a/README.md
+++ b/README.md
@@ -150,11 +150,8 @@ Vertical or horizontal lines are supported.
         // Text to display in label - default is null. Provide an array to display values on a new line
         content: "Test label",
 
-        // Rotation of label, in degrees, default is 0
+        // Rotation of label, in degrees, or 'auto' to use the degrees of the line, default is 0
         rotation: 90,
-        
-        // if set to true and rotation is missing, use the degrees of the oblique line to rotate the label, default is false 
-        autoRotation: false
     },
 
     // Event hooks - context: {chart, element}
--- a/README.md
+++ b/README.md
@@ -30,22 +30,21 @@ To configure the annotations plugin, you can simply add new config options to yo
             // See http://www.chartjs.org/docs/#advanced-usage-creating-plugins
             drawTime: 'afterDatasetsDraw', // (default)
 
-            // Mouse events to enable on each annotation.
-            // Should be an array of one or more browser-supported mouse events
-            // See https://developer.mozilla.org/en-US/docs/Web/Events
-            events: ['click'],
-
             // Double-click speed in ms used to distinguish single-clicks from
             // double-clicks whenever you need to capture both. When listening for
             // both click and dblclick, click events will be delayed by this
             // amount.
             dblClickSpeed: 350, // ms (default)
 
+            // Fires when an annotation is hovered.
+            enter: function(context) {
+                // context: {chart, element}
+            },
+
             // Array of annotation configuration objects
             // See below for detailed descriptions of the annotation options
             annotations: [{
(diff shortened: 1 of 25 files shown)


src/annotation.js
```diff
@@ -174,7 +176,8 @@ function calculateElementProperties(chart, options, defaults) {
 		y2,
 		width: x2 - x,
 		height: y2 - y,
-		options: merge(Object.create(null), [defaults, options])
+		options: merge(Object.create(null), [defaults, options]),
+		_obliqueLine: oblique
```
kurkle, line 180: I'd contain the logic completely to the `line` element, because it only applies to line, right?

README.md
```diff
@@ -151,7 +151,10 @@ Vertical or horizontal lines are supported.
         content: "Test label",
 
         // Rotation of label, in degrees, default is 0
-        rotation: 90
+        rotation: 90,
```
kurkle, line 154: How about making the `rotation` accept `'auto'` instead of adding another option and needing complex if statements to figure out what to do (and ending up ignoring the rotation option)?

src/types/line.js
```diff
@@ -113,10 +127,12 @@ function drawLabel(ctx, line) {
 
 	const {width, height} = measureLabel(ctx, label);
 	const pos = calculateLabelPosition(line, width, height);
+	const rotation = line._obliqueLine && !label.rotation && !!label.autoRotation ? calculateAutoRotation(line, chartArea) : label.rotation * PI / 180;
```
kurkle, line 130: So this could be `const rotation = label.rotation == 'auto' ? calculateAutoRotation(line, chartArea) : toRadians(label.rotation);`

src/types/line.js
```diff
@@ -101,7 +102,20 @@ LineAnnotation.defaults = {
 	}
 };
 
-function drawLabel(ctx, line) {
+function calculateAutoRotation(line, chartArea) {
+	const {x, y, x2, y2} = line;
+	let cathetusAdjacent, cathetusOpposite;
+	if (line.options.mode === 'horizontal') {
+		cathetusAdjacent = y2 > y ? chartArea.width : -chartArea.width;
```
kurkle, line 109: should not need `chartArea` to calculate the angle between 2 points?
Something like (not tested):
```js
return Math.atan2(x - x2, y - y2);
```

kurkle (commented):
Some thoughts

README.md
stockiNail: > 
> 
> How about making the `rotation` accept `'auto'` instead of adding another option and needing complex if statements to figure out what to do (and ending up ignoring the rotation option)?

You're right!!! I really didn't think to implement in this way... I'm gonna change it as suggested

src/annotation.js
stockiNail: I have added some logic over there (instead of in line) because annotation has got already the logic there but having a look better, I can put it in line . Let me try

src/types/line.js
stockiNail: > 
> 
> should not need `chartArea` to calculate the angle between 2 points?
> Something like (not tested):
> 
> ```js
> return Math.atan2(x - x2, y - y2);
> ```

Let me check

stockiNail: Doesn't work but let me take some time to check better

autoRot1

kurkle: probably need to swap the axes: `Math.atan2(y - y2, x - x2);`

stockiNail: already done, but it doesn't still work (for horizontal line)

stockiNail: Can I do with my method only without chartArea?

stockiNail:
```javascript
function calculateAutoRotation(line) {
	const {x, y, x2, y2} = line;
	let cathetusAdjacent, cathetusOpposite;
	if (line.options.mode === 'horizontal') {
		cathetusAdjacent = y2 > y ? x2 - x : -(x2 - x);
		cathetusOpposite = Math.abs(y - y2);
	} else {
		cathetusAdjacent = Math.abs(x - x2);
		cathetusOpposite = x2 > x ? y2 - y : -(y2 - y);
	}
	return Math.atan(cathetusOpposite / cathetusAdjacent);
}
```

kurkle: Lines drawn from origin (middle) out:
image

https://codepen.io/kurkle/pen/eYzrxEL

stockiNail: `vert` and `horz` are using the method `calc1` implemented and they are correct.

Instead `atan2`, by `calc2` method,  doen't work correctl, as far as I see, because the label is upside down. Am I wrong?

kurkle: I think most will agree with you. Though Annotation lines would most likely be going in the directiond on the right side of that image (where all solutions seem ok). I'd expect a line going right to left (backward)  have the label positioned differently compared to lime going left to right. But as said thats unlikely on annotations and more likely be an error in config. And you can always do the rotation manually. So I'm fine with either.

stockiNail: Fine! I agree with you. Let me know if anything is missing from your standpoint.
I have just a question: for many months (I don't want to say years) this plugin was stuck, therefore I think they are many missing things, like callbacks on many options.
Maybe, additional improvements (I have got some in my mind) could be postponed after the release of next version of this plugin.
What do you think?

kurkle (approved):

src/types/line.js
kurkle: I had some reason I'd like to push a release for v2 before going for first beta for v3.
Soonish.

kurkle:
One more thought. Should we make 'auto' the default? (Would match the PR description better :))

stockiNail:
> 
> 
> One more thought. Should we make 'auto' the default? (Would match the PR description better :))

Yes, we could (not should !! ;)). The only point of attention to do it is ,changing the default, the label will follow always the line and currently the default is to have the label horizontally (rotation 0). 
Some users could not appreciate that.

etimberg (approved):

PR #261, opened 2020-11-13
Add initial types

kurkle:
Requires: https://github.com/chartjs/Chart.js/pull/8048

Closes: https://github.com/chartjs/Chart.js/issues/7945

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -370,6 +370,12 @@
       "integrity": "sha512-P8BjAsXvZS+VIDUI11hHCQEv74YT67YUi5JJFNWIqL235sBmjX4+qx9Muvls5ivyNENctx46xQLQ3aTuE7ssaQ==",
       "dev": true
     },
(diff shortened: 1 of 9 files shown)


stockiNail:
@kurkle I see you have changed the callback invocation for `display` property.
Be aware that it's still not 100% correct because the context used for other callbacks are passing the element instead of annotation.
See https://github.com/chartjs/chartjs-plugin-annotation/pull/256
I can resubmit, removing `this` as you did but passing the correct context.

kurkle:
> @kurkle I see you have changed the callback invocation for `display` property.
> Be aware that it's still not 100% correct because the context used for other callbacks are passing the element instead of annotation.
> See #256
> I can resubmit, removing `this` as you did but passing the correct context.

merged & rebased

stockiNail:
@kurkle THANK YOUUU!

PR #270, opened 2020-11-27
Line labels position options (not center) don't work

### LINKED ISSUE #269: Label position not equals to 'center' doesn't work
In the master (therefore in new version and only there) if, configuring a line annotation for the label, you set the `position` , to `top, bottom, left or right`, the label is not shown.
It works only with `center`.

It seems an issue on interpolation calculation in the `calculateLabelPosition` of line annotation.

https://github.com/chartjs/chartjs-plugin-annotation/blob/da07bf955e00523bc8912e5017a36af7cd8c68f9/src/types/line.js#L8-L9

stockiNail:
Fixes #269


kurkle:
this too needs rebase

stockiNail:
rebased

kurkle:
This seems to include the font merging or part of it.

stockiNail:
> 
> 
> This seems to include the font merging or part of it.

Yes, let me remove it.

stockiNail:
Merge of label font with chart font has been removed.

src/annotation.js
```diff
@@ -121,28 +132,19 @@ function updateElements(chart, state, options, mode) {
 		if (!el || !(el instanceof elType)) {
 			el = elements[i] = new elType();
 		}
-		const properties = calculateElementProperties(chart, el, annotation, elType.defaults);
-		animations.update(el, properties);
-
-		const display = typeof annotation.display === 'function' ? callCallback(annotation.display, [{chart, element: el}]) : valueOrDefault(annotation.display, true);
-		el._display = !!display;
+		el.options = merge(Object.create(null), [elType.defaults, annotation]);
```
kurkle, line 135: This is from the font merge. Should not set options directly, its going to skip animations.

samples/autoRotation.html
```diff
@@ -97,13 +99,15 @@
...
 								myVertical: {
 									type: 'line',
-									mode: 'vertical',
 									scaleID: 'x',
 									value: 'February',
 									endValue: 'January',
 									borderColor: 'black',
 									borderWidth: 5,
 									label: {
+										font: {
```
kurkle, line 108: These too are from the font merge

src/types/line.js
```diff
@@ -88,6 +89,7 @@ export default class LineAnnotation extends Element {
 				y = min;
 				y2 = max;
 			}
+			options._horizontal = !scale.isHorizontal();
```
kurkle, line 92: lets have this kind of things directly on the element

src/annotation.js
stockiNail: > Should not set options directly, its going to skip animations.

Does that mean that the options of element must be created by `animations.update();` ?
I have done in this way in order to do not have `_display` property on element but the `display` into options.

src/types/line.js
stockiNail: oki

samples/autoRotation.html
stockiNail: oki

src/annotation.js
stockiNail: Another doubt, @kurkle: I have seen the animation config but in my opinion the elements of this plugin are not animated.
Drawing the elements directly into plugin, how could they be animated?
Am I wrong?

stockiNail:
@kurkle applied all updates, suggested above

src/annotation.js
```diff
@@ -98,6 +98,18 @@ function resolveAnimations(chart, animOpts, mode) {
 	return new Animations(chart, animOpts);
 }
 
+function initElement(chart, element, elementType, options, animations) {
```
kurkle, line 101: Sorry, can you revert this whole thing? Lets not move the `_display` to `options.display` or name this function `initElement`, because its done multiple times to same element, so its not initializing. Also lets keep it below `updateElements`, for easier reading of the file.

src/types/point.js
```diff
@@ -72,7 +73,7 @@ PointAnnotation.defaultRoutes = {
 };
 
 function pointInCircle(p, point) {
-	const {width} = point;
+	const width = point.width;
```
kurkle, line 76: +4 bytes to bundle size, lets revert this one too.

src/types/line.js
```diff
@@ -141,11 +143,7 @@ function calculateAutoRotation(line) {
 function drawLabel(ctx, line) {
 	const label = line.options.label;
 
-	ctx.font = fontString(
-		label.font.size,
-		label.font.style,
-		label.font.family
-	);
+	ctx.font = toFontString(label.font);
```
kurkle, line 146: Lets keep this one, although it does not really reflect anything in the PR title.

For future, try to keep a PR small and only do the thing title says. Its much easier to review that way. 
Also when you can, don't put them on top of each other (I know this is because I asked you to split them, but you could have done it differently), it also makes it easier to review and avoids the rebase after every merge.

Sorry for nagging.

src/types/line.js
```diff
@@ -26,7 +26,7 @@ export default class LineAnnotation extends Element {
 	}
 
 	isOnLabel(x, y) {
-		const {labelRect} = this;
+		const labelRect = this.labelRect;
```
kurkle, line 29: This one also only adds weight.

src/types/line.js
```diff
@@ -88,6 +88,7 @@ export default class LineAnnotation extends Element {
 				y = min;
 				y2 = max;
 			}
+			this._horizontal = !scale.isHorizontal();
```
kurkle, line 91: Lets not do two things in this function (resolve and set one of the properties, while returning the rest).
You can add the `_horizontal` to the returned object instead. And you might want to move this function out of the class to avoid the //eslint-disable.

kurkle (changes requested):

src/annotation.js
stockiNail: OKI! Done! I'm gonna push in next minutes

src/types/point.js
stockiNail: OKI! Done! I'm gonna push in next minutes.
I didn't know if there is a standard approach for that. I thought `const {properties} = instance;` could be used when you have more than 1 property to get, and not based on amount of bytes.

src/types/line.js
stockiNail: @kurkle you're right and apologize for that. Sometimes it's difficult to submit more than 1 PR without any dependencies among that. In the future I'm gonna submit 1 PR at once in order to do not have this cascade.
Again, sorry...

src/types/line.js
stockiNail: OKI! Done! I'm gonna push in next minutes.
I didn't know if there is a standard approach for that. I thought const {properties} = instance; could be used when you have more than 1 property to get, and not based on amount of bytes.

src/types/line.js
stockiNail: > Lets not do two things in this function (resolve and set one of the properties, while returning the rest).
> You can add the `_horizontal` to the returned object instead.

Oki even if I thought that all annotations have got own logic to resolve the elements options and the result objetc could be different. But ok!

> You can add the `_horizontal` to the returned object instead. And you might want to move this function out of the class to avoid the //eslint-disable.

It could be static method (or a function of element type). Ok for you?

src/types/line.js
kurkle: I don't think there is a standard, except to try to have the files as small as possible. If we were using babel to support older browsers, then this one would be better, because the destructuring assignment is not supported by old browsers.

src/annotation.js
kurkle: Take a look at the `animation.html` sample :)

src/annotation.js
stockiNail: @kurkle Now I remember why I have created another function.
ESLINT:
```javascript
101:1  error  Function 'updateElements' has a complexity of 11. Maximum allowed is 10  complexity
```

I can remove the check if `_display` is `true` but in my opinion that's not correct. 
What do you think?

src/types/line.js
kurkle: I don't think it can be static, and actually can't be moved out of the class, because its accessed through the class. Lets just leave it as is.

src/annotation.js
kurkle: oh, we have that strict rules. I'll update the eslint config.

src/types/line.js
stockiNail: I have tried something like that (here LineAnnotation but it's the same for all other types):
```javascript
LineAnnotation.resolveElementProperties = function(chart, options) {
	const scale = chart.scales[options.scaleID];
	let {top: y, left: x, bottom: y2, right: x2} = chart.chartArea;
	let min, max, _horizontal;

	if (scale) {
		min = scaleValue(scale, options.value, NaN);
		max = scaleValue(scale, options.endValue, min);
		if (scale.isHorizontal()) {
			x = min;
			x2 = max;
		} else {
			y = min;
			y2 = max;
		}
		_horizontal = !scale.isHorizontal();
	}
	return {
		x,
		y,
		x2,
		y2,
		width: x2 - x,
		height: y2 - y,
		_horizontal
	};
};
```

and in the annoation.js I invoke it by the element type:

```javascript
const properties = elType.resolveElementProperties(chart, annotation);
```

Make sense? OTherwise I'll revert to previous method of the class.

kurkle: I don't think you can use `elType` to call a static. You would need to use `LineAnnotation`. I disabled the `methods-use-this` rule, so those comments can be removed.
You need to rebase (and `npm i` to get a eslint-plugin-es) to get the new settings.

kurkle: (I might be wrong about the static thing, but I don't really see a reason going that way)

stockiNail: > 
> 
> (I might be wrong about the static thing, but I don't really see a reason going that way)

Nop, I'll revert to the method of the class.

stockiNail:
Requested changes  should be done.

src/types/line.js
```diff
@@ -1,10 +1,10 @@
 import {Element, defaults} from 'chart.js';
-import {isArray, fontString, toRadians} from 'chart.js/helpers';
+import {isArray, toFontString, toRadians} from 'chart.js/helpers';
 import {scaleValue, roundedRect, inTriangle} from '../helpers';
 
 const pointInLine = (p1, p2, t) => ({x: p1.x + t * (p2.x - p1.x), y: p1.y + t * (p2.y - p1.y)});
-const interpolateX = (y, p1, p2) => pointInLine(p1, p2, Math.abs((y - p1.y) / (p2.y - p1.y))).x;
-const interpolateY = (x, p1, p2) => pointInLine(p1, p2, Math.abs((x - p1.x) / (p2.x - p1.x))).y;
+const interpolateX = (y, p1, p2) => pointInLine(p1, p2, y / (p2.y + p1.y)).x;
```
kurkle, line 6: oh, now I realize this does not work.

Say the Y of p1 is 50, and p2 is 100. Now if we want to interpolate the x at Y=50, we need to have the 3rd parameter be 0. `50 / (100 + 50)` is not 0. `(50-50)/(100-50)` however is. Also at Y=100, the 3rd parameter needs to be 1. 100/150 is not 1.

Where did the original go wrong?

src/types/line.js
```diff
@@ -88,14 +88,16 @@ export default class LineAnnotation extends Element {
...
 		}
 		return {
 			x,
 			y,
 			x2,
 			y2,
 			width: x2 - x,
-			height: y2 - y
+			height: y2 - y,
+			_horizontal
```
kurkle, line 100: `_horizontal: !scale.isHorisontal()`

stockiNail: `scale` could be `undefined` (see previous if statement) and therefore I'll get an error, with your code.

kurkle: Ah, right!

src/types/line.js
stockiNail: Well, I'm sure that I have found the previous interpolation didn't work... But now, I don't remember that use case...
Let me revert  to previous interpolation. If  I'll find the issue again, I'll open an issue.

stockiNail: Ignore my previous post... The reason of the change was in the issue #269

stockiNail: In the master (therefore in new version and only there) if, configuring a line annotation for the label, you set the `position` , to `top, bottom, left or right`, the label is not shown.
It works only with `center`.

It seems an issue on interpolation calculation in the `calculateLabelPosition` of line annotation.

https://github.com/chartjs/chartjs-plugin-annotation/blob/da07bf955e00523bc8912e5017a36af7cd8c68f9/src/types/line.js#L8-L9

That's the reason and with new "interpolation" I'm quite sure that works.

stockiNail: It works with original interpolation. Let me revert to it.

kurkle (approved):

PR #267, opened 2020-11-27
 Adds cornerRadius options to box annotation

stockiNail:
adds `corneRadius` options to box annotation.

cornerRadius

PR #274, opened 2020-12-09
Various fixes

joshkel:
This fixes a few mistakes I found:

* Fix moveListened vs. listened: If a chart's only hooks were move hooks on individual annotations, then when the chart is updated, `state.listened` was cleared at the beginning of `updateListeners`, but since `state.moveListened` was still set, the `annotations.forEach` loop didn't re-evaluate move listeners and re-set `state.listened`.
* Update plugin hooks for the [new IPlugin interface](https://www.chartjs.org/docs/master/getting-started/v3-migration#iplugin-interface): This fixes problems with `handleEvent` not seeing the correct event type and with the drawTime option not being honored.
* Fix ESLint warnings: ESLint was giving a [no-invalid-this](https://eslint.org/docs/rules/no-invalid-this) warning. The annotations.display callback isn't documented as having a `this` parameter, and I don't see any way that `this` could work in this context, so I removed it.

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -119,7 +119,7 @@ function updateElements(chart, state, options, mode) {
 		const properties = calculateElementProperties(chart, annotation, elType.defaults);
 		animations.update(el, properties);
 
-		const display = typeof annotation.display === 'function' ? callCallback(annotation.display, [{chart, element:el}], this) : valueOrDefault(annotation.display, true);
+		const display = typeof annotation.display === 'function' ? callCallback(annotation.display, [{chart, element: el}]) : valueOrDefault(annotation.display, true);
 		el._display = !!display;
 	}
 }
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -48,21 +48,21 @@ export default {
 		updateElements(chart, state, options, args.mode);
 	},
 
-	beforeDatasetsDraw(chart, options) {
+	beforeDatasetsDraw(chart, _args, options) {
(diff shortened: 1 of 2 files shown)


stockiNail:
Last item about _this_ of display callback was already addressed by PR #267.

joshkel:
Thanks. After I submitted my PR, I also noticed that #261 addresses the `this` display callback change. Do you want me to remove that from my PR? Or leave it for now and let it be resolved later if a merge conflict appears?

stockiNail:
I think you can leave it and let's see the result of merge. Probably there will be a conflict with the other PRs where I moved also the calls to display callback.
Let's trust GIT ;)

stockiNail:
>     * Update plugin hooks for the [new IPlugin interface](https://www.chartjs.org/docs/master/getting-started/v3-migration#iplugin-interface): This fixes problems with `handleEvent` not seeing the correct event type and with the drawTime option not being honored.

Maybe you could add also the `afterBuildTicks` hook and remove the following code:

https://github.com/chartjs/chartjs-plugin-annotation/blob/da07bf955e00523bc8912e5017a36af7cd8c68f9/src/annotation.js#L40-L42

and move/adapt the function

https://github.com/chartjs/chartjs-plugin-annotation/blob/da07bf955e00523bc8912e5017a36af7cd8c68f9/src/annotation.js#L195-L213

into the hook: 
```javascript
afterBuildTicks?(chart: Chart, args: { scale: Scale }, options: O): void;
```

kurkle (approved):

Issue #243, opened 2020-10-28
Labels with drawTime: beforeDatasetsDraw


joshkel:
I wanted my datasets to be drawn on top of my line and box annotations, so I used `drawTime: 'beforeDatasetsDraw'`. However, that causes the datasets to overwrite the lines' and boxes' labels, which doesn't look good at all:

![image](https://user-images.githubusercontent.com/476338/97492629-300eb580-193a-11eb-8747-22584bc5433d.png)

It seems to me that, even if annotation lines and boxes are drawn below the datasets, the labels for those annotations should be drawn on top of datasets. (At least, I can't think of any circumstances where you'd want a text label but are okay with it being partially obscured by your data...)

I saw similar questions and discussions (#94, #147, #161), although my request is a bit different - I'm proposing moving label-drawing to a separate loop that's always after the datasets (ignoring `drawTime`), instead of #161's approach of moving label-drawing to a separate loop within the configured `drawTime`.

kurkle:
Would it be good to have separated draw Time for labels?
There is some work towards z-index already in chart.js, but I think it would need a new hook and/or breaking the old ones (can't really have both z and afterDatasetsDraw)

joshkel:
Yes, I'd be happy with a separate draw time for labels. (I really can't think of a time when you wouldn't want them after the datasets - but just because I can't think of something doesn't mean it's not useful.)

I'm happy to submit submit a PR if you like. (Possibly as an update to #161? It makes sense to handle the two together.)

I saw comments about a z-index for Chart.js but wasn't clear on how far along it was.

Thank you.

kurkle:
I'm happy to review and merge a pr for that. Can be based on #161, but does not have to be. Pick the route that is easier for you.

joshkel:
Done in #275

PR #275, opened 2020-12-09
Implement a separate labelDrawTime option

joshkel:
Implementation notes:

Since not all annotation types implement labels, the `drawLabel` method is optional.

If the `labelDrawTime` option isn't specified, then it checks the individual annotation's `drawTime` option, then the chart-wide annotation plugin's `labelDrawTime` option, then the chart-wide annotation plugin's `drawTime` option. This means that you can't override an individual annotation's `drawTime` without also affecting its `labelDrawTime`, but I think it's the more intuitive and backward-compatible behavior.

Since the option should default to `drawTime` if not set, I did not add it to the `defaults` object. Please let me know if I've misunderstood how plugins' default options should be implemented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,12 +41,6 @@ To configure the annotations plugin, you can simply add new config options to yo
                 // context: {chart, element}
             },
 
-            label: {
-                // Defines when annotations' labels are drawn.  Defaults to
-                // drawTime.
-                drawTime: 'afterDatasetsDraw', // (default)
-            },
-
             // Array of annotation configuration objects
             // or Object with the annotation configuration objects, one for each key
             // See below for detailed descriptions of the annotation options
@@ -92,9 +86,6 @@ Vertical or horizontal lines are supported.
     // optional drawTime to control layering, overrides global drawTime setting
     drawTime: 'afterDatasetsDraw',
 
-    // optional drawTime to control labels' layering; defaults to this element's drawTime
-    labelDrawTime: 'afterDatasetsDraw',
-
     // ID of the scale to bind onto
     scaleID: 'y',
 
@@ -121,6 +112,9 @@ Vertical or horizontal lines are supported.
     label: {
         // Background color of label, default below
         backgroundColor: 'rgba(0,0,0,0.8)',
+
+        // optional drawTime to control labels' layering; defaults to this element's drawTime
+        drawTime: 'afterDatasetsDraw',
 
         font: {
             // Font family of text, inherits from global
--- a/README.md
+++ b/README.md
@@ -30,6 +30,10 @@ To configure the annotations plugin, you can simply add new config options to yo
             // See http://www.chartjs.org/docs/#advanced-usage-creating-plugins
             drawTime: 'afterDatasetsDraw', // (default)
 
+            // Defines when annotations' labels are drawn.  Defaults to
+            // drawTime.
+            labelDrawTime: 'afterDatasetsDraw', // (default)
+
             // Double-click speed in ms used to distinguish single-clicks from
             // double-clicks whenever you need to capture both. When listening for
             // both click and dblclick, click events will be delayed by this
@@ -41,7 +45,7 @@ To configure the annotations plugin, you can simply add new config options to yo
                 // context: {chart, element}
             },
 
-            // Array of annotation configuration objects 
+            // Array of annotation configuration objects
             // or Object with the annotation configuration objects, one for each key
             // See below for detailed descriptions of the annotation options
             annotations: [{
@@ -85,6 +89,9 @@ Vertical or horizontal lines are supported.
 
     // optional drawTime to control layering, overrides global drawTime setting
     drawTime: 'afterDatasetsDraw',
+
+    // optional drawTime to control labels' layering; defaults to this element's drawTime
+    labelDrawTime: 'afterDatasetsDraw',
 
     // ID of the scale to bind onto
     scaleID: 'y',
@@ -214,7 +221,7 @@ The 4 coordinates, xMin, xMax, yMin, yMax are optional. If not specified, the bo
 
     // Fill color
     backgroundColor: 'green',
-    
+
     // Radius of box rectangle, default below
     cornerRadius: 0,
 
@@ -304,7 +311,7 @@ The 2 coordinates, xValue, yValue are optional. If not specified, the point is e
 
     // Stroke width
     borderWidth: 2,
-    
+
     // Radius of the point, default below
     radius: 10,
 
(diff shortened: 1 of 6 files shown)


joshkel:
This should fix #243 and #94. It supersedes #161.

README.md
```diff
@@ -30,6 +30,10 @@ To configure the annotations plugin, you can simply add new config options to yo
             // See http://www.chartjs.org/docs/#advanced-usage-creating-plugins
             drawTime: 'afterDatasetsDraw', // (default)
 
+            // Defines when annotations' labels are drawn.  Defaults to
+            // drawTime.
+            labelDrawTime: 'afterDatasetsDraw', // (default)
```
stockiNail, line 35: @joshkel may I ask you why to use a property at element level instead of using `drawTime` property at `label` node?

src/annotation.js
```diff
@@ -180,15 +180,19 @@ function calculateElementProperties(chart, options, defaults) {
...
-		const el = elements[i];
-		if (el._display && (el.options.drawTime || options.drawTime || caller) === caller) {
+	elements.forEach(el => {
+		if ((el.options.drawTime || options.drawTime || caller) === caller) {
 			el.draw(ctx);
 		}
-	}
+	});
+	elements.forEach(el => {
+		if ('drawLabel' in el && (el.options.labelDrawTime || el.options.drawTime || options.labelDrawTime || options.drawTime || caller) === caller) {
```
stockiNail, line 192: @joshkel using `drawTime` at `label` level, can the statement be the following?

```javascript
if ('drawLabel' in el && (el.options.label.drawTime || el.options.drawTime || options.drawTime || caller) === caller) {
```

stockiNail (commented):
I would suggest to use `drawTime` options into `label` node instead of `labelDrawTime` at option level.

joshkel:
Thanks for the suggestion, @stockiNail. I hadn't considered using a `drawTime` property at the `label` node. I do like that better, although I have a question about how it should be designed: Should setting a chart-wide default drawTime for labels be permitted? That would mean that the chart's `options.plugins.annotation` has a `label` sub-object with only one usable property. (The line annotation plugin's labels implementation doesn't otherwise pull any label defaults from the chart's `options.plugins.annotation` - maybe that's worth raising separately, if it's an issue for anyone. Or maybe there's a way to better merge in defaults automatically; I'm still learning my way around the Chart.js APIs.)

I noticed that text annotations were listed under "To-do Items" (and it's probably something that our project will eventually need). I'm guessing that a hypothetical text annotations feature would use the same `label` sub-object as lines (to keep a consistent interface) and so whatever we decide here would work for that, too.

stockiNail:
> Should setting a chart-wide default drawTime for labels be permitted? That would mean that the chart's `options.plugins.annotation` has a `label` sub-object with only one usable property. (The line annotation plugin's labels implementation doesn't otherwise pull any label defaults from the chart's `options.plugins.annotation` - maybe that's worth raising separately, if it's an issue for anyone. Or maybe there's a way to better merge in defaults automatically; I'm still learning my way around the Chart.js APIs.)

@joshkel you would like to set the label `drawTime` default into annotation plugin options in order that all labels of all annotation can inherit from that. In this case the `labelDrawTime` should be set, agree with you. Do you think this is the good use case?
In this case, forgive because it makes sense to me as well, I didn't think about that case even if usually this decoupled drawing should be an exception, I guess, and then could be configurable for a single annotation.

> I noticed that text annotations were listed under "To-do Items" (and it's probably something that our project will eventually need). I'm guessing that a hypothetical text annotations feature would use the same label sub-object as lines (to keep a consistent interface) and so whatever we decide here would work for that, too.

I  was thinking about that (also to have label on the box and all other annotation types) and maybe an abstract label class is needed and every annotation could extend and implement special labeling.
This is valid (in my opinion) also for `TextAnnotation` (not implemented yet).
But I didn't do yet because I'm working on other stuff, sorry.

stockiNail:
@joshkel apologize, I forgot an important thing on my previous answer.
@kurkle taught me that the annotations are now Chart.js `elements` and therefore you can configure the annotation (by its type) also in `Chart.defaults.elements` and/or `chart.options.elements`, like point, bar, arc or line (default elements of Chart.js).

Therefore if you want to configure all line annotations (but this is valid for all annotation types) with a specific `drawTime` (and then let's assume to use `drawTime` for `label `as well), you can do something like that:

```javascript
options: {
   elements: {
       lineAnnotation: {
         drawTime: 'beforeDatasetsDraw',
         label:  {
            drawTime: 'afterDraw'
         },
   }
   ....
}
```

You can do the same into `Chart.defaults` as well.

In this way you can configure the `drawTime` default option for all labels of all line annotations.

Unfortunately a piece of code is currently missing into plugin but I think it makes sense to add:

https://github.com/pepstock-org/chartjs-plugin-annotation/blob/a46915f3e409928281f1e0c9d550bc524c1bddc3/src/annotation.js#L136

```javascript
el.options = merge(Object.create(null), [elType.defaults, chart.options.elements[elType.id], annotation]);
```
This line is of my branch where the options are calculated differently for your code (when merged and after conflicts resolving, it should work changing that).

In this way you can get the `drawTime` also into your label as default for all line annotation and then the drawing can use `drawTime` (instead of `labelDrawTime`) for the label as well.

If you agree on that (to use `drawLine` for `label` and use the elements for default configuration), I can add the update to my another branch and submit another PR.

kurkle:
I think I like the `label.drawTime` more. Lets go with that. 

I'm planning to provide some utilities for this in the core, no need to write the resolution path in every plugin (and extension):

> el.options = merge(Object.create(null), [elType.defaults, chart.options.elements[elType.id], annotation]);

stockiNail:
> I'm planning to provide some utilities for this in the core, no need to write the resolution path in every plugin (and extension):

Sounds good!

joshkel:
Apologies for the delayed reply.

I changed to `label.drawTime`, as suggested, since that definitely seems like a cleaner design.

Do you want me to do anything with merging options now, or wait until other PRs and/or core work?

src/types/line.js
```diff
@@ -130,6 +134,7 @@ function drawLabel(ctx, line) {
 
 	line.labelRect = {x: pos.x, y: pos.y, width, height};
 
+	ctx.beginPath();
```
kurkle, line 137: For the current use cases I'd move this in the `roundedRect` function (should remove from beginPath() from box.js then)

kurkle (commented):
Other than the beginPath() and merging README.md with master, this seems good to go.

src/types/line.js
joshkel: Done. Thanks.

kurkle (approved):

PR #300, opened 2021-01-09
Initialize always all elements even if do not show 

### LINKED ISSUE #298: `display` callback and lazy initialization of elements
The current implementation of the `display` property creates new, empty annotation objects as needed, but it only populates those objects if the element is displayed. This creates a couple of issues:

* When the chart is first shown, the element parameter that's passed to the `display` callback is empty, so the callback doesn't have access to the information it needs to determine if the element should be shown.
* If an annotation is hidden (via `display: false` or returning false from a callback), then the chart is updated (via a call to `chart.update()`), then chartjs-plugin-annotation tries to process the empty annotation object and generates errors. See https://codepen.io/joshkel/pen/GRjBMQM for an example.

https://github.com/chartjs/chartjs-plugin-annotation/blob/3741a30d230e201fc8e9f961e31e650a5245387f/src/annotation.js#L112-L123

stockiNail:
Fixes #298

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -113,14 +113,12 @@ function updateElements(chart, state, options, mode) {
     if (!el || !(el instanceof elType)) {
       el = elements[i] = new elType();
     }
+    const properties = el.resolveElementProperties(chart, annotation);
+    properties.options = merge(Object.create(null), [elType.defaults, annotation]);
+    animations.update(el, properties);
+
     const display = typeof annotation.display === 'function' ? callCallback(annotation.display, [{chart, element: el}]) : valueOrDefault(annotation.display, true);
     el._display = !!display;
-
-    if (el._display) {
-      const properties = el.resolveElementProperties(chart, annotation);
-      properties.options = merge(Object.create(null), [elType.defaults, annotation]);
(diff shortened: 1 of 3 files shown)


kurkle (dismissed):

kurkle:
It would be good to add a test each time a bug is fixed. Would you like to add one?

stockiNail:
I'm not so skilled on test framework but let me try.
Probably time is needed to me to learn it a bit.
Ok?

kurkle:
Sure!

You can do something like that: https://github.com/chartjs/Chart.js/pull/8291/files/3b2df9da0d1ddaff3be1add3a05442b2ad703639..47780b433a7e6492f0c62a8c69d1c8660524da70

Just with the chart config from the issue :)

Why I linked the example is: You need to wrap the test in a function, when expecting `toThrow()` or `not.toThrow()`.

stockiNail:
Thank you very much @kurkle

stockiNail:
@kurkle I should have been able to create the test case.

kurkle (approved):

kurkle:
Thanks, @stockiNail

Issue #271, opened 2020-11-29
Label is cut due to rotation


stockiNail:
When you configure the `rotation` to the line label, the `x` and `y` (calculated interpolating the points on the line) are producing the following result:

![download](https://user-images.githubusercontent.com/11741250/100551367-40d78180-3280-11eb-8ea9-0851bcfb4737.png)

As you can see the label of red line is cut (due to the clip area).
The `x` and `y` calculation (by  `height` and `width` and `calculateLabelPosition` function) is assuming that the label has got rotation 0 but with different rotation,`height` and `width` of rotated label are changed and in cascade also`x` and `y`. 

With PR #266 which is calculating the vertex of label, it's possible to adjust the label position.
**I'm gonna submit a PR to fix it.**

Next enhancement could be to allow the user to draw outside of clip area (currently set to `chartArea`), maybe by an options (for instance `clip: true` (default)) in order that  issue #258 could be addressed as well.

kurkle:
Thanks for all the pull requested. I'll try to get beta.7 of Chart.js out and when that is out, will focus here. So please be patient :)

stockiNail:
@kurkle yes, no problem. I see you are deeply engaged to beta7 (I have read all discussion about plugin hooks, really interesting).
I have already fixed part of the this issue but I need more time as well to address all possible cases about label positions. 
I'm gonna wait your feedback before pushing anything.

Thank youuuu

stockiNail:
@kurkle FYI, I fixed and now the labels are shown correctly. I'll take time to "optimized" my poor JS.

Issue #258, opened 2020-11-10
Feature Request: bracket annotation


danatcofo:
![image](https://user-images.githubusercontent.com/418147/98617700-2c2d4c80-22c5-11eb-9f7c-d0633ab7d01b.png)

I would love to have the ability to do a bracket annotation as I am showing here in with the orange brackets in this terrible visual.  basically like the vertical or horizontal line in function except rather than a line, we get a caret or symbol of some type on top|bottom|left|right within the area outside the plot.

This would be really useful when there are a lot of series that need drawing and the annotation is easily lost.

stockiNail:
@danatcofo just for my curiosity, could make sense to use a char (a char annotation), where with font size, color and rotation you can set whatever char you want (for instance curly or squared or round brackets) ?
Of course, using the canvas drawing, it's quite difficult to give the freedom to the user to use the symbol he/she wants. With chars could be easier.

stockiNail:
@danatcofo is something like that?

![Issue258](https://user-images.githubusercontent.com/11741250/100747997-0e926500-33e3-11eb-98fa-e99590e45ff7.PNG)

This is done using 2 line annotations with the same value and the labels (1 on top and 1 on bottom) with y adjust.
If sounds good, you don't need to new annotation but you can use line annotation one with labels.

Nevertheless this is doable if the clipping on chart area will be configurable. At the moment it is not.
You also should pay attention that when you are drawing outside of chart area, you can override other elements (legend, title, ticks) or go out of the canvas (in this case you should set the `layout.padding` of chart.js options).

Annotation plugin options, used for the above sample (config of plugin for CHART.JS 3):

```javascript
plugins: {
   annotation: {
      drawTime: 'afterDatasetsDraw',
      annotations: {
         top: {
            type: 'line',
            scaleID: 'x',
            value: 'March',
            borderColor: 'transparent',
            label: {
               position: 'top',
               font: {
                  size: 48,
                  color: 'orange'
               },
               backgroundColor: 'white',
               borderWidth: 0,
               yPadding: 0,
               xPadding: 0,
               yAdjust: -22,
               rotation: 90,
               content: '}',
               enabled: true
            }
         },
         bottom: {
            type: 'line',
            scaleID: 'x',
            value: 'March',
            borderColor: 'transparent',
            label: {
               position: 'bottom',
               font: {
                  size: 48,
                  color: 'orange'
               },
               backgroundColor: 'white',
               borderWidth: 0,
[...]

PR #305, opened 2021-01-12
Enable line annotation using two scales

kurkle:
Changes:

- `position`: 'start', 'end' or 'center'
- alternative xScaleID, yScaleID, xMin, yMin, xMax, yMax options for `line` annotation
- label hit testing by rotating the point (similar to datalabels plugin)
- simpler label position adjustment by calculating the parameter `t` for `pointInLine`


kurkle:
| `dist/chartjs-plugin-annotation.esm.js` | 6.07 kB | -505 B (-8%) | ✅ |
| `dist/chartjs-plugin-annotation.js` | 6.18 kB | -506 B (-8%) | ✅ |
| `dist/chartjs-plugin-annotation.min.js` | 4.17 kB | -447 B (-10%) | 👏 |

@stockiNail take a look, this touches few of the contributions you've made.

stockiNail:
@kurkle thank you very much!
I'm gonna have a look tomorrow.

stockiNail:
@kurkle first feedback:

When the label is rotated an close to boundaries of chart area, the label is cut.,

This was an issue #271 that was solved re-positioning the label.

cut

I'm still testing, more feedbacks in next hours

kurkle:
Thanks @stockiNail, I failed to consider that situation. Current code does move that label down so it fits the chart area. But that can get you only so far:

image

same with this pr: (note the text direction, that might be the next feedback :))

image

In this case, removing the clipping would give a good result, but it does not help if there is no room outside the chart area.
I'm wondering if it would be better to consider this when setting the scale ranges?

stockiNail:
> same with this pr: (note the text direction, that might be the next feedback :))

I have already done :D. See my previous note in the review close to the code.

stockiNail:
> In this case, removing the clipping would give a good result, but it does not help if there is no room outside the chart area.
I'm wondering if it would be better to consider this when setting the scale ranges?

Pr #282 solved this issue recalculating the label position in order to stay inside the chart area.

I know that it's not so simple (that's why I added chart reference into element in that PR).

Removing the cliparea could be configurable in order to address the issue #258 as well, something like that:

```javascript
label: {
	clipEnabled: false, // <-- defaults to true
	position: 'start',
	rotation: -45,
	backgroundColor: 'red',
	content: 'This is a long test label',
	enabled: true
},
```

In this way you can delegate the user to have enough space for the label.

To change the scales limits (as is done now for the values) could be a solution but this could change the chart going to a picture that probably you don't like with additional space in chart area that you want to avoid.
Honestly, I don't like also the current implementation which is changing the min/max of scales when the values are out of scale, therefore we can maintain as is.
If this will be the solution, we must calculate the rotation and then the necessary space for the label before the element drawing is invoked, I guess.

stockiNail:
Other feedback (not so important). All samples are still using the previous `position` option values (left, right, top and bottom).
If it can help, I can submit a PR to your branch, if you don't have time to do it.

kurkle:
> Other feedback (not so important). All samples are still using the previous `position` option values (left, right, top and bottom).
> If it can help, I can submit a PR to your branch, if you don't have time to do it.

Found only one sample using that. And it turs out it does not work correctly with this PR:

pr: (start)
image

master: (top)
image

stockiNail:
@kurkle It seems that `return Math.max(x, y)` (in `calculateTAdjust` method) is wrong because is returning the maximun value instead it should return x (if line is horizontal) and y (if line is vertical)

stockiNail:
Not sure if it could be correct but the return could be

 return w > h ? x : y;

if width of line if greater than height means that the line is horizontal

kurkle:
Last commit solves it like this (test label with long text is `positon: 'end'`:

image

stockiNail:
The commit "Fix calculateTAdjust" doesn't fix the issue. 

fixcjpa

kurkle:
Ok, another try. 

Adjust the label position in 2 phases, first along the line, up to 25% from the desired end, then with slightly modified current method.

image

image

kurkle:
| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/chartjs-plugin-annotation.esm.js` | 6.61 kB | +35 B (+1%) |
| `dist/chartjs-plugin-annotation.js` | 6.73 kB | +41 B (+1%) |
| `dist/chartjs-plugin-annotation.min.js` | 4.48 kB | -142 B (-3%) |

etimberg (commented):
Didn't notice anything wrong. is the change to `start`, `center`, `end` considered breaking?

stockiNail:
@etimberg in my opinion, yes. You don't have errors because if the position is not start or end it will use center, without any issue.
Nevertheless who is using now left, right, top or bottom will have the label in the center, therefore I think it should be considered a breaking change.
Am I wrong?

kurkle:
@etimberg thats the reason I labeled this breaking.

stockiNail:
@kurkle apologize for delay of my feedback but I had other stuff to do it.
But finally I had time to test it and it sounds working correctly!
Thank you very much!

etimberg (dismissed):

PR #311, opened 2021-01-14
Auto register/unregister annotation types

kurkle:
Annotation types are not really designed to be shakeable, so it makes sense to automatically register them as elements when the plugin is registered.

This is one step towards removing the automatic registration from `UMD` build, for consistency.

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -1,4 +1,4 @@
-import {Animations} from 'chart.js';
+import {Animations, Chart} from 'chart.js';
 import {clipArea, unclipArea, isFinite, merge, valueOrDefault, callback as callCallback, isObject} from 'chart.js/helpers';
 import {handleEvent, updateListeners} from './events';
 import BoxAnnotation from './types/box';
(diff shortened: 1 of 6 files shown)


simonbrunel (commented):
It looks good, however the TS definition seems incorrect now since it [still exports everything](https://github.com/chartjs/chartjs-plugin-annotation/blob/master/types/index.d.ts#L12) and doesn't have any default export (which should be the plugin).

kurkle:
Thanks, changed the export to default for the plugin.

I'm not sure if there is anything private in the types, so did not touch the other exports.

The failing test made me realize the plugin would need to be registered to work now. Eg. the `plugins: [Annotation]` use case might not work.

simonbrunel:
> The failing test made me realize the plugin would need to be registered to work now. Eg. the `plugins: [Annotation]` use case might not work as expected (missing defaults).

If I understand your point, it looks like this issue impacts all plugins. Actually, the fact that the plugin `defaults` are now part of the `plugin` object breaks the possibility to register a plugin locally.

simonbrunel (approved):

PR #316, opened 2021-01-22
Enables the callbacks for "values" options

### LINKED ISSUE #284: Set line annotation at specified data point.
I want to set the value of a line annotation equal to a specified data point, can I use a callback function which has the context containing the chart when set the value item in annotation config?

stockiNail:
Enables the user to configure the "values options" (`value`, `endValue`, `xMin`, `xMax`, `yMin`, `yMax`, `xValue`, `yValue`) in order to define the values of annotations at runtime.

Fixes #284

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,17 +88,52 @@ To configure the annotations plugin, you can simply add new config options to yo
     scaleID: 'y',
 
     // Data value to draw the line at
+    // A callback can also be used:
+    //   value(context) {
+    //     // context: {chart, element}
+    //     return value;
+    //   },
     value: 25,
 
     // Optional value at which the line draw should end
+    // A callback can also be used:
+    //   endValue(context) {
+    //     // context: {chart, element}
+    //     return value;
+    //   },
     endValue: 26,
 
     // Alternative configuration of values, using two scales
     xScaleID: 'x',
     yScaleID: 'y',
+    
+    // Optional value for x point of starting line, in units along the x axis
+    // A callback can also be used:
+    //   xMin(context) {
+    //     // context: {chart, element}
+    //     return value;
+    //   },
     xMin: 'January',
+    // Optional value for y point of starting line, in units along the x axis
+    // A callback can also be used:
+    //   yMin(context) {
+    //     // context: {chart, element}
+    //     return value;
+    //   },
     yMin: 25,
+    // Optional value for x point of ending line, in units along the x axis
+    // A callback can also be used:
+    //   xMax(context) {
+    //     // context: {chart, element}
+    //     return value;
+    //   },
     xMax: 'February',
+    // Optional value for y point of ending line, in units along the x axis
+    // A callback can also be used:
+    //   yMax(context) {
+    //     // context: {chart, element}
+    //     return value;
+    //   },
     yMax: 26,
 
     // Line color
@@ -202,15 +237,35 @@ The 4 coordinates, xMin, xMax, yMin, yMax are optional. If not specified, the bo
     yScaleID: 'y',
 
     // Left edge of the box. in units along the x axis
+    // A callback can also be used:
+    //   xMin(context) {
+    //     // context: {chart, element}
+    //     return value;
+    //   },
     xMin: 25,
 
     // Right edge of the box
+    // A callback can also be used:
+    //   xMax(context) {
+    //     // context: {chart, element}
+    //     return value;
+    //   },
     xMax: 40,
 
     // Top edge of the box in units along the y axis
+    // A callback can also be used:
+    //   yMax(context) {
+    //     // context: {chart, element}
+    //     return value;
+    //   },
     yMax: 20,
 
     // Bottom edge of the box
+    // A callback can also be used:
+    //   yMin(context) {
+    //     // context: {chart, element}
+    //     return value;
+    //   },
     yMin:  15,
 
     // Stroke color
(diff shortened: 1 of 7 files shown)


stockiNail:
@kurkle I think there is a bug in the `adjustScaleRange` that I discovered by this PR.

The method doesn't use the "new" options that plugin `afterDataLimits` hook is providing and then if the user has changed the "values" options, these are ignored and the max and min scales are using previous value.

stockiNail:
> @kurkle I think there is a bug in the `adjustScaleRange` that I discovered by this PR.
> 
> The method doesn't use the "new" options that plugin `afterDataLimits` hook is providing and then if the user has changed the "values" options, these are ignored and the max and min scales are using previous value.
> 
> Shall I open an issue or it can be solved in this PR?

The initialization of elements is performed into `afterUpdate` which is called after `afterDataLimits` hook.
That mean that inside the `afterDataLimits` hook we could use the elements but only the options.
Nevertheless I'm gonna try to move the initialization of elements in `beforeUpdate`. I don't know why updateElements is invoked after the update.

stockiNail:
>  I don't know why updateElements is invoked after the update.

It seems that `chartArea` of chart instance is available only `afterUpdate`. 
In this case the `adjustScaleRange` must use the options and not the elements..

stockiNail:
@kurkle in this implementation, the element instance can not be passed in the context because both does not exist or is not initialized. 
Maybe the scriptable context should contain the annotation options instead of annotation element.

kurkle:
> @kurkle I think there is a bug in the `adjustScaleRange` that I discovered by this PR.
> 
> The method doesn't use the "new" options that plugin `afterDataLimits` hook is providing and then if the user has changed the "values" options, these are ignored and the max and min scales are using previous value.
> 
> Shall I open an issue or it can be solved in this PR?

I think it would be better to separate. I have some concerns about the current approach in this PR (if the value is string for example, it would not work). This should also mean that the annotations are not updated when options are changed?

stockiNail:
> I think it would be better to separate.

I can split PR to solve the issue of `adjustScaleRange` and the callbacks implementation.

> I have some concerns about the current approach in this PR (if the value is string for example, it would not work).

I can't understand well what you meant. This PS is not changing anything about if the value (or the result of callback) is a string or not. The logic is the same there was before, I guess. Therefore I think if the value as string doesn't work with this PR, it should work with the previous code. Anyway my doubt about string is that we should check if the value is really a label of the chart. If not, the annotation should be disabled. 
 
> This should also mean that the annotations are not updated when options are changed?

With former code, when you change the plugin annotation in the chart and you will update the chart to use new values, it used the elements in the state which are not updated because the `updateElements` is performed afterwards, in the `afterUpdate` plugin hook because to update the elements you need the `chartArea`.

If you agree, I'm gonna leave the PR (with callbacks) as is and I'll create new one with a possible solution for this `adjustScaleRange`. Ok for you?

stockiNail:
Splitted. The issue related the `adjustScaleRange` is in this PR  #323

stockiNail:
@kurkle I was reviewing this PR.

Even if I think that the values could be set by a callback as scriptable options because this can enable some use cases like to place the annotation based on data (to calculate average, std deviation, linear regression), I have some doubts about the consistency of this implementation.

First of all, the context passed to a callback. The context is defined as an object with 2 options, chart and element (which contains the annotation options). Unfortunately the elements are created in the `afterUpdate` hook but the callbacks for value must be executed before that hook, in the `afterDataLimits` one. Therefore the callbacks can pass only the options.
And for the same reason, also the callback for `display` option must be executed before the elements creation therefore also that callback could not have element.
My first doubt is if make sense to pass an element (which is not always available) or instead to pass the options (always available).
Of course we can have different context for different callbacks.

Another topic is the callbacks are called twice, during both `afterDataLimits` and `afterUpdate`. For performance reason it could be better to invoke once and store the result in an object. The best could be the element, but as said above, if not available.
Therefore I was thinking if make sense to create all elements (empty, with the options) in `beforeUpdate` hook. 
But in this case, being currently the element filling the object by the animation, it is still not clear to me the role of the `animation` object because, having a look to all samples, I can not see any example where the elements is drawing following the different phases of animation because the annotation is drawn always with the final position.

Sorry for long post.. I need your hints to understand if this PR makes sense as is or if must be changed or  removed in favor of another implementation
Thanks a lot in advance.

stockiNail:
> But in this case, being currently the element filling the object by the animation, it is still not clear to me the role of the `animation` object because, having a look to all samples, I can not see any example where the elements is drawing following the different phases of animation because the annotation is drawn always with the final position.

Now I remember, the sample animation.html even if I'd like that the drawing of animations will be animated but I don't know how to enable it.

stockiNail:
I did some tests splitting the updateElements method in 2 in order to revert part of PR #323 and then to use elements also in `adjustScaleRange` method.

The `beforeUpdate` plugin hook will engaged new method, `createElements`, like the following (just a prototype therefore the code must be improved):

```javascript
function createElements(chart, state, options) {
  const annotations = options.annotations || [];
  const elements = resyncElements(state.elements, annotations);

  for (let i = 0; i < annotations.length; i++) {
    const annotation = annotations[i];
    let el = elements[i];
    const elType = annotationTypes[annotation.type] || annotationTypes.line;
    }
    el.options = merge(Object.create(null), [chart.options.elements[elType.id], annotation]);
    el._display = isAnnotationVisible(chart, annotation, el);
  }
}
```

This method is creating the elements, setting the options.
In this way, the `afterDataLimits` hook, which is calling `adjustScaleRange` method and after the `beforeUpdate`, can leverage on elements, avoiding to apply the same logic to filter options.

And then, afterUpdate will remain the same but the updateElements is updating the animations and resolve the element properties (just a prototype therefore the code must be improved):

```javascript
function updateElements(chart, state, options, mode) {
  const chartAnims = chart.options.animation;
  const animOpts = chartAnims && merge({}, [chartAnims, options.animation]);
  const animations = resolveAnimations(chart, animOpts, mode);

  const elements = chartStates.get(chart).elements;
  elements.forEach(el => {
    const properties = el.resolveElementProperties(chart, el.options);
    properties.options = el.options;
    animations.update(el, properties);
  });	
}
```

What do you think?

stockiNail:
I forgot... of course, if OK, we can close this PR going to new 2 PRs,  the first to split the `updateElements` (as above) and another one to implement the callbacks.

stockiNail:
See PR #326which is refactoring the elements creation and use in the plugin.
Another PR (after approval and merged of #326) will be submitted to implement the value callbacks.

PR #321, opened 2021-01-27
Manages values as date for time scales

### LINKED ISSUE #319: Invalid scale value with time scales
If you are using the time scale and you have an annotation related to that scale, the annotation is incorrectly shown. 

Here is the result of the [sample of the project](https://github.com/chartjs/chartjs-plugin-annotation/blob/master/samples/line-time-scale.html) where the time scale is used:

errorTimeScale

The chart has got 3 annotation, 2 lines and 1 box. 1 line is correctly shown because based on the linear scale. The other 2 are completely wrong.

After some tests, I think the issue is here:

https://github.com/chartjs/chartjs-plugin-annotation/blob/ab7dd60770b11e484f76a2da55cfc3f15dd76bc5/src/helpers.js#L8

`isFinite` function is returning `false` because the value is an object (Date) and not a number.

stockiNail:
Fixes #319

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -4,7 +4,7 @@ const PI = Math.PI;
 const HALF_PI = PI / 2;
 
 export function scaleValue(scale, value, fallback) {
-  value = typeof value === 'string' ? scale.parse(value) : value;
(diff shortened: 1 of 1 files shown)


stockiNail:
The updates added in #323

PR #323, opened 2021-01-29
Adjust scale range method using options instead of elements

stockiNail:
The `adjustScaleRange` method, which is calculating and adjusting the min and max values of a scale based on values of annotations, is evaluating the annotations values taking the options from elements in the state.
Elements are created/updated after the `adjustScaleRange` method therefore the method should not use elements. In this PR the method is using the options. 

Setting the min and max values for a category scale doesn't make sense because the value of annotation should be in the configured labels of the chart.

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -1,6 +1,5 @@
 import {Animations, Chart} from 'chart.js';
 import {clipArea, unclipArea, isFinite, merge, valueOrDefault, callback as callCallback, isObject} from 'chart.js/helpers';
-import {valueAsNumber} from './helpers';
 import {handleEvent, updateListeners} from './events';
 import BoxAnnotation from './types/box';
 import LineAnnotation from './types/line';
@@ -214,7 +213,7 @@ function getScaleLimits(scale, annotations) {
   scaleAnnotations.forEach(annotation => {
     ['value', 'endValue', axis + 'Min', axis + 'Max', 'xValue', 'yValue'].forEach(prop => {
       if (prop in annotation) {
-        const value = valueAsNumber(scale, annotation[prop]);
+        const value = scale.parse(annotation[prop]);
         min = Math.min(min, value);
         max = Math.max(max, value);
       }
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -38,7 +38,7 @@ export default {
   },
 
   afterDataLimits(chart, args, options) {
-    if (scale.type !== 'category') {
+    if (args.scale.type !== 'category') {
       adjustScaleRange(chart, args.scale, options);
     }
   },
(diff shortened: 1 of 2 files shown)


src/helpers.js
```diff
@@ -4,10 +4,14 @@ const PI = Math.PI;
 const HALF_PI = PI / 2;
 
 export function scaleValue(scale, value, fallback) {
-  value = typeof value === 'string' ? scale.parse(value) : value;
+  value = valueAsNumber(scale, value);
   return isFinite(value) ? scale.getPixelForValue(value) : fallback;
 }
 
+export function valueAsNumber(scale, value) {
+  return typeof value === 'string' ? scale.parse(value) : value instanceof Date ? value.getTime() : value;
```
kurkle, line 12: Why not  use the `scale.parse` for date?

src/annotation.js
```diff
@@ -37,8 +38,9 @@ export default {
   },
 
   afterDataLimits(chart, args, options) {
-    const state = chartStates.get(chart);
-    adjustScaleRange(args.scale, state, options);
+    if (args.scale.type !== 'category') {
```
kurkle, line 41: Category scale can be adjusted outside its labels, so I'm not sure why we exclude it here?

kurkle (commented):
I think I'd call the helpers `scaleValue` (new one) and `scalePixel` (old one)

src/helpers.js
stockiNail: I didn't know it... I'll change it.

src/annotation.js
stockiNail: Yes  but I can't understand how from the plugin.
Just an example:

```javascript
....
data: {
   .....
},
options: {
   .....
   plugin: {
     my : {
        ....
        xMax: 'December',
        xMin: 'Feb'
     }
   }
}
```

How you can calculate the min and max? And out to add them? In my opinion, when you have a category scale, you should change min and max because they are not progressive values and the high risk is to have something that you don't want to have on the chart. But Maybe I'm wrong.

src/helpers.js
stockiNail: > 
> 
> Why not use the `scale.parse` for date?

I had a look to `parse` method. Why don't use it for all types (also number)? 
It sounds working

kurkle: because of the fallback logic. but maybe that can be done differently.

stockiNail: > 
> 
> because of the fallback logic. but maybe that can be done differently.

Let me try if I can.

kurkle (approved):

src/annotation.js
stockiNail: @kurkle thank you for approving but am I wrong on above thought about category scale? Maybe I'm missing anything

kurkle: You should be able to use numbers relative to label indices. For example, `-0.5` should work.

PR #326, opened 2021-01-31
New "createElements" method invoked in "beforeUpdate" hook 

stockiNail:
New `createElements` method invoked in `beforeUpdate` hook in order to enable the use of element wherever in the plugin (see PR  #316).

The previous `updateElements` (which created the elements and update them) has been split and now just updates the elements properties called from `afterUpdate` hook.

It reverts part of the PR #323 in order to use again the elements instead of options. This is improving the performance because teh display callback (if present) will be called only once.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ To configure the annotations plugin, you can simply add new config options to yo
 {
     type: 'line',
 
-    // If true, display the annotation, default is true
+    // If true, display the annotation, default is true.
     // A callback can also be used:
     //   display(context) {
     //     // context: {chart, element}
@@ -81,19 +81,19 @@ To configure the annotations plugin, you can simply add new config options to yo
     //   },
     display: true,
 
-    // optional drawTime to control layering, overrides global drawTime setting
+    // optional drawTime to control layering, overrides global drawTime setting.
     drawTime: 'afterDatasetsDraw',
 
-    // ID of the scale to bind onto
+    // ID of the scale to bind onto.
     scaleID: 'y',
 
-    // Data value to draw the line at
+    // Data value to draw the line at.
     value: 25,
 
-    // Optional value at which the line draw should end
+    // Optional value at which the line draw should end.
     endValue: 26,
 
-    // Alternative configuration of values, using two scales
+    // Alternative configuration of values, using two scales.
     xScaleID: 'x',
     yScaleID: 'y',
     xMin: 'January',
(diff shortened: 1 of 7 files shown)


stockiNail:
I have created this PR based on #325 and I know that could create some annoying task to review the code.
Nevertheless this PR makes sense with the other PR implementation.

kurkle:
Commenting the different things happening here:

- Updating the README.md default values
  - I think the update to README.md makes sense. I also think we should have a generic section for things that apply to all elements (like the display), instead of repeating it.
- Creating elements in beforeUpdate
  - Did you try just calling the old `updateElements` from `beforeUpdate`? I think it should work (and should not break any ongoing animations on annotation elements when the chart.update is called)
- Resolving the element defaults
  - Should be ok

kurkle:
> I have created this PR based on #325 and I know that could create some annoying task to review the code.
> Nevertheless this PR makes sense with the other PR implementation.

I see. Please avoid that. These things don't even depend much on each other.

stockiNail:
> Updating the README.md default values
> 
>     * I think the update to README.md makes sense. I also think we should have a generic section for things that apply to all elements (like the display), instead of repeating it.

Fully agree. What do you think to change it in order to go to Chart,js style, with the tables of options instead of the code?
I'm gonna start working on that asap.

> Creating elements in beforeUpdate
> 
>     * Did you try just calling the old `updateElements` from `beforeUpdate`? I think it should work (and should not break any ongoing animations on annotation elements when the chart.update is called)

Yes but it doesn't work because you need `chartArea` object that is available starting form `afterUpdate` hook.

> Resolving the element defaults
> 
>     * Should be ok

That's fine!

stockiNail:
@kurkle rebased. I know that you have already told me to avoid to "extend" another PR.. Really apologize... Let me share with you that I have many things that could be applied here (for instance the callbacks, images in labels content, single instance of context in order that a use can add own options) that I don't write them, I'll forget and I want to have my fork aligned to  the development of main repo.
sorry again.

kurkle:
I'm not a fan of this approach.

How would you feel about storing the merged (and resolved) options into state in `beforeUpdate`, and using those everywhere?
Could also avoid modifying the `options.annotation` when its an object.

stockiNail:
> How would you feel about storing the merged (and resolved) options into state in `beforeUpdate`, and using those everywhere?
> Could also avoid modifying the `options.annotation` when its an object.

Yeah, I agree. That's what I thought at the beginning but I also thought that, when we decided to use the element as part of scriptable context, we have evaluated that the callbacks could consume also other options, owned by element.

You are proposing to use options (merged, resolved and stored into state) instead of elements in the callbacks and wherever is possible instead of element, isn't it?

stockiNail:
> You are proposing to use options (merged, resolved and stored into state) instead of elements in the callbacks and wherever is possible instead of element, isn't it?

@kurkle If yes, I'm gonna start working tomorrow (or the day after tomorrow).

kurkle:
> > You are proposing to use options (merged, resolved and stored into state) instead of elements in the callbacks and wherever is possible instead of element, isn't it?
> 
> @kurkle If yes, I'm gonna start working tomorrow (or the day after tomorrow).

I think I'd add the options as another argument for the callbacks, or stored in the context, not sure yet. I'd still include the element in the context too, when its available. So `callHandler(handler, [{chart, element, options}]);` or `callHandler(handler, [{chart, element}, options]);`, not quite sure yet.

stockiNail:
@kurkle I'd vote for `callHandler(handler, [{chart, element, options}]);`. 
In the meantime if you agree I could try to create new PR (this one sounds being created in the wrong direction) with a possible implementation (shared above) and then you could check if you like it.
Or if you want to do it, I'll wait for that.

kurkle:
> @kurkle I'd vote for `callHandler(handler, [{chart, element, options}]);`.
> In the meantime if you agree I could try to create new PR (this one sounds being created in the wrong direction) with a possible implementation (shared above) and then you could check if you like it.
> Or if you want to do it, I'll wait for that.

I'm quite busy currently, so go ahead if you have the time.

stockiNail:
> I'm quite busy currently, so go ahead if you have the time.

Me too but I promise you I'm gonna do my best.

stockiNail:
Partially addressed by #327. Scriptable options will come by another PR.

PR #329, opened 2021-02-03
Enables borderDash and borderDashOffset options to all annotations

stockiNail:
This PR is enabling `borderDash` and `borderDashOffset` to _box_, _ellipse_ and _point_ annotations.

The README is updated maintaining the current structure. A separate PR will be submitted to change README in order to be more consistent and readable.


stockiNail:
I wanted to create a test case but it's not clear to me how to obtain the picture to compare with the chart (as it did for category-index).

kurkle:
> I wanted to create a test case but it's not clear to me how to obtain the picture to compare with the chart (as it did for category-index).

Not sure what part you are missing, so listing all here:
1. If creating a new folder under fixtures, need to add it in one of the specs (`specs/box.spec.js`):
```js
describe('Box annotation', function() {
  describe('auto', jasmine.fixtures('box'));
});
```
`'box'` is the folder under `fixtures`

2. Add the config for the fixture
3. `npm run dev` and hit the `debug` button to see the image
4. Save that image

src/types/ellipse.js
```diff
@@ -18,6 +18,11 @@ export default class EllipseAnnotation extends BoxAnnotation {
     ctx.strokeStyle = options.borderColor;
     ctx.fillStyle = options.backgroundColor;
 
+    if (ctx.setLineDash) {
```
kurkle, line 21: These if's can be removed, it is for some old IE that is not supported anymore.

https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/setLineDash#browser_compatibility

stockiNail:
>     2. `npm run dev` and hit the `debug` button to see the image

That's what I didn't know! I'm gonna prepare the test case.

src/types/ellipse.js
stockiNail: Removed everywhere. It was previously used in the line annotation and I have removed from there as well

stockiNail:
Test cases have been added

kurkle (approved):

Issue #318, opened 2021-01-27
Annotation Label Content Can't have HTML tags


sasiel:
The annotation label content supports only text, can it support html tags as well?

stockiNail:
@sasiel the annotation are drawn on the canvas element which, afaik, doesn't support any HTML tag, therefore the labels can be only text.

sasiel:
I see, what the complexity of adding an html tag support? I can help with this PR if needed

stockiNail:
@sasiel it doesn't depend on the complexity. Afaik, but maybe I'm wrong, the canvas element doesn't support HTML tags.

sasiel:
@stockiNail Got it, thank you!

stockiNail:
@sasiel maybe my answer hasn't evaluated the possibility to use SVG (and foreignObject) in order to create an image adding HTML tags into SVG. I have already this way in another project (https://github.com/pepstock-org/Charba/wiki/Utilities#html-annotation-builder).

Another thing is that the plugin defines the label text by `content` option and not `text` one. This is probably ecuase who developed the plugin at the beginning thought do not use only text but also other kind of contents. 

I have coded a prototype, allowing to the plugin to accept an image (this could be helpful also per other purposes).

This is a piece of html page where I'm creating the SVG:

```javascript
const image = new Image();
	image.src = 'data:image/svg+xml;charset=utf-8,<svg xmlns="http://www.w3.org/2000/svg" width="150" height="60">'+
	'<foreignObject width="100%" height="100%">'+
	'<div xmlns="http://www.w3.org/1999/xhtml">Here is a <strong style="color: white;">paragraph</strong> that requires <em>word wrap</em></div>'+
	'</foreignObject></svg>';
```

Then I have defined a line annotation passing the image into label content:

```javascript
plugins: {
    annotation: {
        annotations: {
            myHorizontalLine: {
                type: 'line',
                scaleID: 'y',
                value: 25,
                borderColor: 'black',
                borderWidth: 5,
                label: {
                    position: 'center',
                    backgroundColor: 'red',
                    content: image, // <-- SVG image
                    enabled: true
                }
            }
        }
    }
}
```

Here is the result:

![svg](https://user-images.githubusercontent.com/11741250/106047330-c2faef80-60e3-11eb-8194-8371488fa5f6.png)

If it makes sense for you, I can try to submit a PR to enable it.

PR #330, opened 2021-02-03
Enables the label content to be an Image.

stockiNail:
This PR is enabling to set the label `content` options as an image.

This could be also helpful when a user would like to apply an text in the label as HTML, see #318.

It's also adding new options, `resizeRatio` (maybe the name is not the best) which can allow to the user to resize the image in the label, by a ratio (percentage).


kurkle:
I'd think the Github mark is copyrighted, not sure if we should use that.

Not sure about the `resizeRatio`. Size in pixels could be more usable generally.
What do you think about having `width` and `height` that can be pixels or percent?

stockiNail:
> I'd think the Github mark is copyrighted, not sure if we should use that.

I think it could be used if I had added the link to Github project, but to avoid any doubt, it has been removed.

stockiNail:
I've tried to add the test case how you explained me but I can not understand why it doesn't work.

```
[test-karma] Chrome 88.0.4324.96 (Linux x86_64) Line annotation auto /base/test/fixtures/line/contentImage.js FAILED
[test-karma] 	Error: [object HTMLDivElement]
[test-karma] 	    at <Jasmine>
[test-karma] 	    at test/node_modules/chartjs-test-utils/dist/chartjs-test-utils.js:955:23 <- test/index.js:770:25
[test-karma] 	    at Image.image.onload (test/node_modules/chartjs-test-utils/dist/chartjs-test-utils.js:527:5 <- test/index.js:364:7)
[test-karma] Chrome 88.0.4324.96 (Linux x86_64): Executed 6 of 9 (1 FAILED) (0 secs / 0.541 secs)
```

kurkle:
> I've tried to add the test case how you explained me but I can not understand why it doesn't work.
> 
> ```
> [test-karma] Chrome 88.0.4324.96 (Linux x86_64) Line annotation auto /base/test/fixtures/line/contentImage.js FAILED
> [test-karma] 	Error: [object HTMLDivElement]
> [test-karma] 	    at <Jasmine>
> [test-karma] 	    at test/node_modules/chartjs-test-utils/dist/chartjs-test-utils.js:955:23 <- test/index.js:770:25
> [test-karma] 	    at Image.image.onload (test/node_modules/chartjs-test-utils/dist/chartjs-test-utils.js:527:5 <- test/index.js:364:7)
> [test-karma] Chrome 88.0.4324.96 (Linux x86_64): Executed 6 of 9 (1 FAILED) (0 secs / 0.541 secs)
> ```

run the 'dev' + debug again to see a detailed error

kurkle:
> If ok, I'd like to use a regex to get the value when set a percentage. Any constraint to use regex?

A constant regex would be fine. I think it could be something simple too:
```js
> const toPercent = (s) => typeof s === 'string' && s.endsWith('%') && parseFloat(s) / 100;
> toPercent('50.1 %')
> 0.501
```

stockiNail:
> run the 'dev' + debug again to see a detailed error

Strange.. no error with run debug

Capture

kurkle:
What kind of test? Maybe the image is not loaded before trying to draw it?

stockiNail:
> A constant regex would be fine. I think it could be something simple too:

I don't know if you see the code.

I have created the regex (constant for all line.js) and a function to calculate image size:

```javascript
const percentageRegex = RegExp('(\\d+(\\.\\d+)?|\\.\\d+)?%');
.....
function getImageSize(size, value) {
  if (typeof value === 'number') {
    return value;
  } else if (typeof value === 'string') {
    const array = percentageRegex.exec(value);
    if (array) {
      return +array[1] / 100 * size;
    }
  }
  return size;
}
...

Let me know if you prefer to have something string base instead of regex.

**Comment by stockiNail** [comment]
> 
> 
> What kind of test? Maybe the image is not loaded before trying to draw it?

**/test/fixtures/line/contentImage.js**

I have the image creation before module export:

```javascript
const htmlText = new Image();
htmlText.src = 'data:image/svg+xml;charset=utf-8,<svg xmlns="http://www.w3.org/2000/svg" width="150" height="80">' +
'<foreignObject width="100%" height="100%">' +
'<div xmlns="http://www.w3.org/1999/xhtml" style="color: white;">Here is a <strong style="color: black;">HTML</strong> text, used as image, which can be styled as <em style="font-size: 16px;">YOU</em> like!</div>' +
'</foreignObject></svg>';
module.exports = {
```

This is the only difference comparing with the others (I did the test case for the other PR and it works).

stockiNail:
> 
> 
> > If ok, I'd like to use a regex to get the value when set a percentage. Any constraint to use regex?
> 
> A constant regex would be fine. I think it could be something simple too:
> 
> ```js
> > const toPercent = (s) => typeof s === 'string' && s.endsWith('%') && parseFloat(s) / 100;
> > toPercent('50.1 %')
> > 0.501
> ```

I'm testing it, removing REGEX going to string toPercent.

kurkle:
I think you'd still need to wait for the loading of the image.
add `await img.decode();` after setting the `src`

ref: https://stackoverflow.com/a/64747517/10359775

stockiNail:
> add `await img.decode();` after setting the `src`

I'm trying but `await` needs async function.

src/types/line.js
```diff
@@ -198,15 +198,7 @@ function drawLabel(ctx, line, chartArea) {
...
-  if (typeof value === 'number') {
-    return value;
-  } else if (typeof value === 'string') {
-    const array = percentageRegex.exec(value);
-    if (array) {
-      return +array[1] / 100 * size;
-    }
-  }
-  return size;
+  return typeof value === 'number' ? value : typeof value === 'string' ? toPercent(value) * size : size;
```
kurkle, line 201: I think your original function body was better :)

```js
if (typeof value === 'number') {
  return value;
}
if (typeof value === 'string' && value.endsWith('%')) {
  return parseFloat(value) / 100 * size;
}
return size;
```

or (less readable, but slightly smaller footprint due to single `return`)

```js
function getImageSize(size, value) {
	return typeof value === 'number' ? value
		: (typeof value === 'string' && value.endsWith('%')) ? parseFloat(value) / 100 * size
			: size;
}
```

stockiNail: I have changed for that reason, to reduce the size but I can change it quickly

stockiNail: done

kurkle (approved):

PR #332, opened 2021-02-04
Add fixture for image label

kurkle:


kurkle:
That did not work.

stockiNail:
I was curious to are how you solved because I have tried also today without any success:(

PR #333, opened 2021-02-09
Fix configuration instructions in README.md

apphancer:
The `color` option no longer seems to work since the latest versions. I believe the configuration instructions in the README should have this changed to `fontColor`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ To configure the annotations plugin, you can simply add new config options to yo
         },
 
         // Color of label text, default is '#fff'.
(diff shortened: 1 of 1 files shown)


kurkle:
What versions are you using? (Chart.js and chartjs-plugin-datalabels)

apphancer:
chartjs-plugin-annotation 0.5.7
Chart.js 2.9.4

kurkle:
From the readme:

> An annotation plugin for Chart.js >= 3.0.0

PR #335, opened 2021-02-11
Use annotation defaults when resolving options

santam85:


kurkle:
Use case is `plugins: [AnnotationPlugin]`, which does not register the elements to defaults and the defaults don't end up in the options.
Would it be possible to register the plugin instead in ng2-charts?

kurkle:
Actually this is probably just fine, because the use case should be supported.

kurkle (approved):

PR #342, opened 2021-03-07
Do not draw box border when width is 0

### LINKED ISSUE #127: Support borderWidth: 0 for box annotations
Hey folks, I am using box annotations to 'mark' areas of a chart (sort of showing down/inactive time periods against other data) and I noticed that specifying `borderWidth` as `0` does not have the intended effect - you always end up with at least a 1px border. At first I tried setting it to transparent or white, but it doesn't look very good.

After a bit of googling it was clear that the issue is `strokeRect` doesn't support 0 `lineWidth` (kind of makes sense? here's a modified codepen from MDN that [demonstrates](https://codepen.io/anon/pen/NzovJw)). So what I think should actually happen is that `strokeRect` should not be called at all if you set `borderWidth: 0` in the settings -- specifically right here: https://github.com/chartjs/chartjs-plugin-annotation/blob/master/src/types/box.js#L139

This is a pretty easy change, obviously, and I have made it on a fork already for my own use. Is this something you would accept as a PR?

etimberg:
Resolves #127


kurkle:
Some linting issues, I've created #346 to prevent these in the future.

kurkle (approved):

PR #356, opened 2021-03-15
Interactive online samples using VuePress

kurkle:
Motivation:
 - Interactive samples
 - The excellent editor plugin @simonbrunel made for VuePress / chartjs-plugin-datalabels

Only 2 samples created for now, so we can see how this works. Current samples moved to `old_samples` folder. We should remove those once everything has a sample in VuePress, or if we want do keep these included, then maybe rename to `html_samples` or equivalent.

image


kurkle:
image

docs/docusaurus.config.js
```diff
@@ -19,6 +19,9 @@ module.exports = {
     },
     navbar: {
       title: 'Chart.js Annotation Plugin',
+      items: [
+        {href: './samples', label: 'Samples'},
```
etimberg, line 23: Will this automatically be pulled into the deploy? I think you might need something in the CI to copy the samples folder to the static dir

kurkle: The `npm run samples` in root directory builds the samples to the static folder, works locally.
Its added in the documentation action.

etimberg: Ah, got it. Missed that it built directly there

etimberg (approved):

PR #366, opened 2021-03-23
Add custom scriptable context

### LINKED ISSUE #84: Draw line by index of yaxis
Hey guys,

I have one page that shows "usage" in timescale chartjs, with update requests ... I dont know the values because they are dates and dynamic. I need to draw a line at index 0 and the last index (maybe something like -1).

How can I make the parameter start at beginning of index? Something like startIndex and stopIndex rather than value? #73

### LINKED ISSUE #348: Scriptable context for scriptable options
Currently all options are scriptable.

The context that is passed as argument is a context of type `chart`.

You don't have any reference to the options of plugin or to the annotation element.

Is that the architectural decision?

In my opinion, at least the options should be passed in the context and maybe the context could be typed differently from `chart`, maybe marked as `annotation`.

Some thing like the following:

## Option Context

The option context is used to give contextual information when resolving options and currently only applies to [scriptable options](#scriptable-options).
The object is preserved, so it can be used to store and pass information between calls.

There are multiple levels of context objects:

* `chart`
  * `annotation`

Each level inherits its parent(s) and any contextual information stored in the parent is available through the child.

The context object contains the following properties:

### chart

* `chart`: the associated chart
* `type`: `'chart'`

### annotation

In addition to [chart](#chart)

* `element`: plugin element which contains also the options
* `type`: `'annotation'`

kurkle:
Resolves #348 
Resolves #84 

(maybe more)

breaking change, because the option resolution is done in `afterDataLimits` and `afterUpdate` vs the previous `beforeUpdate`


docs/guide/options.md
```diff
@@ -10,7 +10,8 @@ Fonts use the same format as [chart.js](https://www.chartjs.org/docs/master/gene
 
 ## Scriptable Options
 
-As with most options in chart.js, the annotation plugin options are scriptable. This means that a function can be passed which returns the value as needed. In the example below, the annotation is hidden when the screen is less than 1000px wide
+As with most options in chart.js, the annotation plugin options are scriptable. This means that a function can be passed which returns the value as needed. In the example below, the annotation is hidden when the screen is less than 1000px wide.
+The function receives 2 arguments, first is the [option context](#option-context) represnting contextual information. An options resolver is passed as second argumet, which can be used to access other option in the same context.
```
etimberg, line 14: [nit] typo in "representing"

etimberg (dismissed):

etimberg (approved):

PR #376, opened 2021-03-26
Type fixes

joshkel:
Allow `annotations` to be an object (as used in the current documentation) as well as an array (as used by older versions of the plugin).

Allow `id` property, as set by the object form of `annotations`.

Suppress ESLint warning; TypeScript declaration merging must exactly match the original, so ESLint's warning that the type is unused is a false positive.

Fix a typo in options.md.

### DIFF
--- a/docs/guide/options.md
+++ b/docs/guide/options.md
@@ -11,7 +11,7 @@ Fonts use the same format as [chart.js](https://www.chartjs.org/docs/master/gene
 ## Scriptable Options
 
 As with most options in chart.js, the annotation plugin options are scriptable. This means that a function can be passed which returns the value as needed. In the example below, the annotation is hidden when the screen is less than 1000px wide.
-The function receives 2 arguments, first is the [option context](#option-context) representing contextual information. An options resolver is passed as second argument, which can be used to access other option in the same context.
+The function receives 2 arguments, first is the [option context](#option-context) representing contextual information. An options object is passed as second argument, containing the options that were originally passed in to configure the annotation element.
 
 ```js chart-editor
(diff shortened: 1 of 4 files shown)


docs/guide/options.md
```diff
@@ -11,7 +11,7 @@ Fonts use the same format as [chart.js](https://www.chartjs.org/docs/master/gene
 ## Scriptable Options
 
 As with most options in chart.js, the annotation plugin options are scriptable. This means that a function can be passed which returns the value as needed. In the example below, the annotation is hidden when the screen is less than 1000px wide.
-The function receives 2 arguments, first is the [option context](#option-context) representing contextual information. An options resolver is passed as second argumet, which can be used to access other option in the same context.
+The function receives 2 arguments, first is the [option context](#option-context) representing contextual information. An options object is passed as second argument, containing the options that were originally passed in to configure the annotation element.
```
kurkle, line 14: Its a proxy object though, that resolves from the options through the annotation type defaults to the annotation defaults. But what ever wording makes it more usable to someone who did not implement it, should be used :)

If you access `options.whatEver` from scriptable `options.display` and the `whatEver` is also scriptable, its called with the same context the `display` was called. Another fancy thing is that you can actually resolve literal `whatEver` option, even though its not built-in.

kurkle (commented):
A comment on the docs side, types look good to me.

docs/guide/options.md
joshkel: Updated. Thank you,

kurkle (approved):

PR #378, opened 2021-03-27
Skip annotations outside chartArea

### LINKED ISSUE #374: Labels still displayed for annotations outside axis limits
My specific case is a series of four charts, one of which has a shorter horizontal (x-) axis than the others. On the chart with the shortened x-axis, there are two line annotations which are less than the minimum value for the axis. The lines aren’t displayed, but the labels are displayed atop each other near the minimum x-axis value. For an example, check [this page](https://covid.jonblatho.com/) and click “Testing” above the chart.

One can, of course, work around this by manually setting `enabled: false` for labels for which the corresponding annotations are located outside the range of values for the axis, but this seems like fairly boilerplate code that might be better handled at the plugin level.

I’m not sure if there is a particular interest in maintaining the current behavior of `enabled: true`, where a label is *forced* to be shown even if its corresponding annotation isn’t visible. If there is such an interest, I would suggest adding a new option such as `enabled: 'auto'`, which would display a label only if the corresponding annotation is visible. (Maybe make it the default, also? To me, at least, this seems like the more intuitive behavior.)

kurkle:
Fix #374


kurkle:
| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/chartjs-plugin-annotation.esm.js` | 7.25 kB | +97 B (+1%) |
| `dist/chartjs-plugin-annotation.js` | 7.37 kB | +97 B (+1%) |
| `dist/chartjs-plugin-annotation.min.js` | 5.01 kB | +60 B (+1%) |

etimberg (approved):

PR #401, opened 2021-04-15
Fix handling of line annotations that fall outside of the plot area

### LINKED ISSUE #400: Line annotations out of plot area are draw incorrectly
As a result of #378, if a line annotation's yMax is greater than the scale's max, the annotation extends to the bottom of the chart instead of the top.

For a vertical line, I can work around this by swapping yMin and yMax. I don't know of a good workaround for diagonal lines. A reversed scale (`reverse: true`) adds another wrinkle; ideally, the annotation configuration shouldn't have to care whether the scale is reversed when deciding which value is min and which value is max.

See https://codesandbox.io/s/chartjs-annotation-off-top-yr2eo?file=/src/index.ts for a demonstration.

joshkel:
Fixes #400.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -3,17 +3,30 @@ import {isFinite} from 'chart.js/helpers';
 const PI = Math.PI;
 const HALF_PI = PI / 2;
 
-function inArea(scale, pixel) {
+function constrainToArea(scale, pixel, fallback) {
   const area = scale.chart.chartArea;
-  return scale.isHorizontal()
+  const inArea = scale.isHorizontal()
     ? pixel >= area.left && pixel <= area.right
     : pixel >= area.top && pixel <= area.bottom;
+  return inArea ? pixel : fallback;
+}
+
+function getPixelValue(scale, value, fallback) {
+  value = typeof value === 'number' ? value : scale.parse(value);
+  return isFinite(value) ? scale.getPixelForValue(value) : fallback;
 }
 
 export function scaleValue(scale, value, fallback) {
-  value = typeof value === 'number' ? value : scale.parse(value);
-  value = isFinite(value) ? scale.getPixelForValue(value) : fallback;
-  return inArea(scale, value) ? value : fallback;
+  const pixel = getPixelValue(scale, value, fallback);
+  return constrainToArea(scale, pixel, fallback);
+}
+
+export function scaleRange(scale, valueFrom, valueTo, fallbackFrom, fallbackTo) {
+  const pixelFrom = getPixelValue(scale, valueFrom, fallbackFrom);
+  const pixelTo = getPixelValue(scale, valueTo, fallbackTo);
+  return pixelFrom < pixelTo
+    ? [constrainToArea(scale, pixelFrom, fallbackFrom), constrainToArea(scale, pixelTo, fallbackTo)]
+    : [constrainToArea(scale, pixelFrom, fallbackTo), constrainToArea(scale, pixelTo, fallbackFrom)];
 }
 
 /**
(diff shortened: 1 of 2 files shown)


kurkle:
I think the new helper could be used in other types to simplify the logic there (and probably fix similar problems).

Can you add fixture tests for this?

joshkel:
Sounds good. I'll put together an updated PR with these changes this weekend.

kurkle:
Can leave the other types for later too, but a test would be good to have in this PR.

tlserver:
I found a problem introduced by #378 but still not fixed in this PR yet.

For example:
```
const config = {
    type: 'line',
    data: {
        datasets: [{
            data: [10, 20, 30, 0, 55],
        }],
        labels: ['A', 'B', 'C', 'D', 'E']
    },
    options: {
        scales: {
            x: {
                min: 'C'
            },
        },
        plugins: {
            legend: false,
            annotation: {
                annotations: {
                    annotation1: {
                        type: 'line',
                        xScale: 'x',
                        yScale: 'y',
                        borderWidth: 3,
                        borderColor: 'black',
                        xMin: 1,
                        xMax: 3,
                        yMin: 0,
                        yMax: 60,
                    }
                }
            }
        },
    }
};
```
Expected:
image
Actual:
image

I think #378 solve #374 in a wrong way? (maybe? I am not sure)

If I understand correctly, #374 is a problem related to label only. However, #378 also change the annotation coordinate calculation method. It may makes lot of breaking change. This PR is making correction based on #378. Therefore, I think this PR may still introduce some other unexpected behaviors.

joshkel:
Thanks for the added info, @tlserver . I hadn't had time to follow up on this PR yet, but I'll look into it this weekend and try and address the new issue, too.

tlserver:
I am trying to fix #374 in other way. But I found some difficulty. So I want to share my current work and discuss with different developer.

This is only a draft.
https://github.com/tlserver/chartjs-plugin-annotation/commit/1eb10bb66c77480ee433860bc7479ade23e5298a

The basic idea to solve #374 is to skip the label drawing if the whole line annotation is out of range.
Let explain with some pictures. The green box represents the chart area and the red line represents a line annotation.
Case 1:
image
both end-points are in chart area
Case 2:
image
only 1 end-point is in chart area

In this two case , the label should not be skipped. But what if both end-points are out of area?

Case 3:
image
This should be skipped. How about this?
Case 4:
image
Both end-point are out of box but it still should be drawn.

So a simple condition for skip drawing is that both end-points are out of area and in same direction. i.e.
```
      skip: x < left && x2 < left ||
        x > right && x2 > right ||
        y < top && y2 < top ||
        y > bottom && y2 > bottom
```

Is this condition work for all cases? The answer is no.

Case 5:
image
This does not pass the above skip condition but it should not be drawn.

It seem that _"one point is on top and other point is on the right of the area"_ is also a case need to be included in the condition. Should we just simply extend the condition and make it cover all the case? The answer is also no.

Case 6:
image
This is another case that meet _"one point is on top and other point is on the right of the area"_ but in this case the label should be drawn.
We cannot distinguish case 5 and 6 if we only consider which direction the end point go out of area.

Maybe, we need to use the following approach (in pseudocode):
---
let p1, p2 // end-point 1, end-point2 (in a format of `{ x: ??, y: ??}` )
let area // area boundary in format of {left, right, top, bottom}

let slope = calculateSlope(p1, p2) // emm , vertical line have a undefined slope, may be use angle is better?
p1 = inArea(p1) ? p1 : cloestPoint(area, p1, slope)
// function `cloestPoint(area, point, slope)` return a point `p` that fulfill all of following requirements:
// 1. `p` is on the area boundary. 
// 2. the slope of a line that connect `p` and `point` is equal to `slope`
// 3. if more then one points fulfill requirement 1 and 2, return the one which is closed to `point`
p2 = inArea(p2) ? p2 : cloestPoint(area, p1, slope)

if (p1 == p2) {
  skipDrawing....
} else {
  drawLine(p1, p2)
  drawLabel(p1, p2) // draw a label at the middle of p1 and p2
}
---
I know that this is quite complicated but I cannot find other better solution right now.

PR #408, opened 2021-04-19
Include sourcemap(s) in package

### LINKED ISSUE #407: sourceMappingURL error in dist/chartjs-plugin-annotation.min.js
Very minor issue here, but when using this plugin on my site, I receive an error because `dist/chartjs-plugin-annotation.min.js` references a source map file that does not exist:

    Failed to load resource: the server responded with a status of 404 (Not Found)
    http://localhost:1313/js/chartjs-plugin-annotation.min.js.map

For clarity, the `dist/` directory from `node_modules/chartjs-plugin-annotation/` is mapped into `js/` on my site using a static site generator.

kurkle:
Resolves #407

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   },
   "files": [
     "dist/*.js",
+    "dist/*.js.map",
     "types/*.d.ts"
   ],
   "scripts": {


kurkle:
Would need the src folder for the map to be of any use.

PR #410, opened 2021-04-22
add adjustScaleRange options

tlserver:
I am trying to add a new option to disable adjusting scale range in `afterDataLimits` hooks.

Consider this example:
```
const config = {
    type: 'line',
    data: {
        datasets: [{
            label: 'My First Dataset',
            data: [65, 59, 80, 81, 56, 55, 40],
            fill: false,
            borderColor: 'rgb(75, 192, 192)',
            tension: 0.1
        }]
    },
    options: {
        plugins: {
            annotation: {
                annotations: {
                    box1: {
                        // Indicates the type of annotation
                        type: 'box',
                        xMin: 1,
                        xMax: 2,
                        yMin: 50,
                        yMax: 700,
                        // adjustScaleRange: false, // <-- adding this option
                        backgroundColor: 'rgba(255, 99, 132, 0.25)'
                    }
                }
            }
        }
    }
};
```
image
The maximum value of y axis is adjusted from 85 to 700.
After uncommenting the `adjustScaleRange: false`, the maximum value of y axis remain unchanged.
image

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -53,7 +53,6 @@ module.exports = [
       name,
       file: main.replace('.js', '.min.js'),
       format: 'umd',
-      sourcemap: true,
       indent: false,
       globals
     },


stockiNail:
@tlserver maybe I'm wrong, but what you need seems to be achievable by setting `yMax` as scriptable options:

```javascript
yMax(ctx){
  let max = Math.max.apply(Math, ctx.chart.data.datasets[0].data);
  return max > 700 ? 700 : undefined;
},
```

See https://codepen.io/stockinail/pen/ExNdPxd?editors=1111

kurkle:
Don't really need to calculate the max by yourself: `yMax: (ctx) => Math.min(ctx.chart.scales.y.max, 700),`.

An explicit option could still be handy.

tlserver:
@stockiNail WoW, scriptable yMax is so powerful. However, it work well with box annotation but not ellipse annotation or line annotation.
Consider this example:
image
This cannot be done by scriptable yMax.

stockiNail:
> However, it work well with box annotation but not ellipse annotation or line annotation.

I have changed the code pen, using the ellipse, and it works. Have a look again: https://codepen.io/stockinail/pen/ExNdPxd

The ellipse annotation extends the box one, therefore they are acting in the same way.

kurkle:
@stockiNail its a different (and I'd guess incorrect) result though.

kurkle:
Defining the axis max also prevents range extension, but I'm guessing that is not feasible either, in your use case.

kurkle:
| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/chartjs-plugin-annotation.esm.js` | 7.38 kB | +9 B (0%) |
| `dist/chartjs-plugin-annotation.js` | 7.5 kB | +11 B (0%) |
| `dist/chartjs-plugin-annotation.min.js` | 5.06 kB | +14 B (0%) |

stockiNail:
> 
> 
> @stockiNail its a different (and I'd guess incorrect) result though.

AH ok! I was thinking the picture wasn't correct.

etimberg (dismissed):

tlserver:
I find a bug after applying my commit 63a2476.

```
const config = {
    type: 'line',
    data: {
        datasets: [{
            label: 'My First Dataset',
            data: [65, 59, 80, 81, 56, 55, 40],
            fill: false,
            borderColor: 'rgb(75, 192, 192)',
            tension: 0.1
        }]
    },
    options: {
        plugins: {
            annotation: {
                annotations: {
                    box1: {
                        // Indicates the type of annotation
                        type: 'box',
                        xMin: 1,
                        xMax: 2,
                        yMin: 500,
                        yMax: 700,
                        adjustScaleRange: false, // <-- adding this option
                        backgroundColor: 'rgba(255, 99, 132, 0.25)'
                    }
                }
            }
        }
    }
};
```
In this example, I changed yMin from 50 to 500 so the box should be fully out of range and become invisible.
However, it currently draw like this:
image
https://codepen.io/tlserver/pen/BapvxVY

I look around the code of library and found that if I revert the commit #378, it works well. I have also saw #400 and #401. Applying #401 and 63a2476 together does not solve this problem.

Therefore, this commit should not be merged into master until #374 is really solved. I am trying to revert #378 and solve #374 in other way now.

kurkle:
I still think this would a good option to have.

tlserver:
@kurkle yes, I think so. It seem that the PR is closed by accident because I have force-pushed something? Sorry about that. Should I create a new PR instead?

PR #414, opened 2021-04-23
Fix drawing of lines with points out of chartArea

### LINKED ISSUE #400: Line annotations out of plot area are draw incorrectly

### LINKED ISSUE #413: Wrong coordinates on annotations extending chartArea with scale limits applied
I found a problem introduced by #378 but still not fixed in this PR yet.

For example:
```
const config = {
    type: 'line',
    data: {
        datasets: [{
            data: [10, 20, 30, 0, 55],
        }],
        labels: ['A', 'B', 'C', 'D', 'E']
    },
    options: {
        scales: {
            x: {
                min: 'C'
            },
        },
        plugins: {
            legend: false,
            annotation: {
                annotations: {
                    annotation1: {
                        type: 'line',
                        xScale: 'x',
                        yScale: 'y',
                        borderWidth: 3,
                        borderColor: 'black',
                        xMin: 'B',
                        xMax: 'D',
                        yMin: 0,
                        yMax: 60,
                    }
                }
            }
        },
    }
};
```
Expected:
image
Actual:
image

I think #378 solve #374 in a wrong way? (maybe? I am not sure)

If I understand correctly, #374 is a problem related to label only. However, #378 also change the annotation coordinate calculation method. It may makes lot of breaking change. This PR is making correction based on #378. Therefore, I think this PR may still introduce some other unexpected behaviors.

_Originally posted by @tlserver in https://github.com/chartjs/chartjs-plugin-annotation/issues/401#issuecomment-825646289_

https://codepen.io/tlserver/pen/BapvxVY

kurkle:
Closes #400
Closes #413
Closes #401


kurkle:
| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/chartjs-plugin-annotation.esm.js` | 7.5 kB | +131 B (+2%) |
| `dist/chartjs-plugin-annotation.js` | 7.62 kB | +137 B (+2%) |
| `dist/chartjs-plugin-annotation.min.js` | 5.15 kB | +105 B (+2%) |

etimberg (approved):

Issue #402, opened 2021-04-15
Events fire outside of finite lines


joshkel:
For a finite line (a line that uses xMin/xMax/yMin/yMax to cover only part of the chart area), the current mouse events can fire if the mouse cursor is where the line would be if it extended the entire length of the chart. For an example, https://codesandbox.io/s/chartjs-annotation-finite-mgupy?file=/src/index.ts and position the mouse at chart coordinates (6, 6).

PR #428, opened 2021-05-28
Fixes for line mouse events

joshkel:
I found a couple of issues with the mouse event handling:

First, the isOnLabel logic didn't check whether the label was currently shown, so if a label had ever been draw in the past, its rectangle would continue to count for determining whether the mouse was on the line annotation.

Second, the logic I added for restricting mouse events to where the line actually is (#403) didn't work for horizontal or vertical lines.  For those, the denominator in interpolateY / interpolateX was 0, so the resulting value was infinity, which was clamped to 1; effectively, only the endpoint of the line could trigger a mouse enter event.

### DIFF
--- a/src/types/line.js
+++ b/src/types/line.js
@@ -4,10 +4,7 @@ import {scaleValue, roundedRect, rotated} from '../helpers';
 
 const PI = Math.PI;
 const clamp = (x, from, to) => Math.min(to, Math.max(from, x));
-const pointInLine = (p1, p2, t) => {
-  t = clamp(t, 0, 1);
-  return {x: p1.x + t * (p2.x - p1.x), y: p1.y + t * (p2.y - p1.y)};
-};
+const pointInLine = (p1, p2, t) => ({x: p1.x + t * (p2.x - p1.x), y: p1.y + t * (p2.y - p1.y)});
 const interpolateX = (y, p1, p2) => pointInLine(p1, p2, Math.abs((y - p1.y) / (p2.y - p1.y))).x;
 const interpolateY = (x, p1, p2) => pointInLine(p1, p2, Math.abs((x - p1.x) / (p2.x - p1.x))).y;
 const toPercent = (s) => typeof s === 'string' && s.endsWith('%') && parseFloat(s) / 100;
--- a/src/types/line.js
+++ b/src/types/line.js
@@ -12,38 +12,6 @@ const interpolateX = (y, p1, p2) => pointInLine(p1, p2, Math.abs((y - p1.y) / (p
 const interpolateY = (x, p1, p2) => pointInLine(p1, p2, Math.abs((x - p1.x) / (p2.x - p1.x))).y;
 const toPercent = (s) => typeof s === 'string' && s.endsWith('%') && parseFloat(s) / 100;
 
-// https://stackoverflow.com/a/6853926/25507
-function distanceBetweenPointAndLineSegment(x, y, x1, y1, x2, y2) {
-  const A = x - x1;
-  const B = y - y1;
-  const C = x2 - x1;
-  const D = y2 - y1;
-
-  const dot = A * C + B * D;
-  const lenSq = C * C + D * D;
-  let param = -1;
-  if (lenSq !== 0) { // in case of 0 length line
-    param = dot / lenSq;
-  }
-
-  var xx, yy;
-
-  if (param < 0) {
-    xx = x1;
-    yy = y1;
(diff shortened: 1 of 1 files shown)


joshkel:
I don't think that will work. If I'm following things correctly, for a horizontal line:

```
    const interpolateX = (y, p1, p2) => pointInLine(p1, p2, Math.abs((y - p1.y) / (p2.y - p1.y))).x;

    const dx = interpolateX(y, p1, p2);
    return (!isFinite(dx) || Math.abs(x - dx) < epsilon);
```

"Find the X on the p1, p2 line that would correspond to the given Y, then see if it's within limits." However, since the two Y values are identical, the denominator is 0, and it's impossible to find a single X value.

After #403, the X value is clamped to x2, so only a single point matches.

Prior to #403, and with your suggestion, the Infinity result is passed through, and so any point matches. But that's not what we want, either - the goal in #402 / #403 was to only match on the line segment itself, rather than matching anywhere on the plot that had the appropriate Y value (for a horizontal line) or X value (for a vertical line).

If I understand correctly, conceptually, we're trying to find the distance from the mouse's point to the line segment, and I suspect that any adequate fixes to the current math will end up being equivalent to the SO code that I borrowed.

src/types/line.js
```diff
@@ -41,15 +73,7 @@ function limitLineToArea(p1, p2, area) {
 export default class LineAnnotation extends Element {
   intersects(x, y, epsilon) {
```
kurkle, line 74:
```suggestion
  intersects(x, y, epsilon = 0.001) {
    // Adapted from https://stackoverflow.com/a/6853926/25507
    const sqr = v => v * v;
    const {x: x1, y: y1, x2, y2} = this;
    const dx = x2 - x1;
    const dy = y2 - y1;
    const lenSq = sqr(dx) + sqr(dy);
    const t = lenSq === 0 ? -1 : ((x - x1) * dx + (y - y1) * dy) / lenSq;
    let xx, yy;
    if (t < 0) {
      xx = x1;
      yy = y1;
    } else if (t > 1) {
      xx = x2;
      yy = y2;
    } else {
      xx = x1 + t * dx;
      yy = y1 + t * dy;
    }
    return (sqr(x - xx) + sqr(y - yy)) < epsilon;
  }
```

src/types/line.js
```diff
@@ -58,6 +82,10 @@ export default class LineAnnotation extends Element {
   }
 
   isOnLabel(mouseX, mouseY) {
+    if (!this.labelIsVisible()) {
+      return false;
+    }
+
     const {labelRect} = this;
 
     if (!labelRect) {
```
kurkle, line 91:
```suggestion
    const {labelRect} = this;
    if (!labelRect || !this.labelIsVisible()) {
```

kurkle (commented):
As the distance function is only used in one place, lets have it directly there. I wrote a condensed suggestion (without the square root, for optimization). Also the `pointInLine` does not need the clamp anymore, right?

joshkel:
I removed the no-longer-needed `clamp` call, and I believe that all of the feedback is addressed now. Thanks for the review and suggestions.

src/types/line.js
```diff
@@ -39,17 +68,29 @@ function limitLineToArea(p1, p2, area) {
...
-    const me = this;
-    const p1 = {x: me.x, y: me.y};
-    const p2 = {x: me.x2, y: me.y2};
-    const dy = interpolateY(x, p1, p2);
-    const dx = interpolateX(y, p1, p2);
-    return (
-      (!isFinite(dy) || Math.abs(y - dy) < epsilon) &&
-			(!isFinite(dx) || Math.abs(x - dx) < epsilon)
-    );
+    return distanceBetweenPointAndLineSegment(x, y, this.x, this.y, this.x2, this.y2) < epsilon;
```
kurkle, line 93: I could not get the suggestion to match all the lines it should have affected. So these lines should be removed and the `distanceBetweenPointAndLineSegment` function as its inlined above

joshkel: I should have paid more attention; I apologize. Fixed.

kurkle:
| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/chartjs-plugin-annotation.esm.js` | 7.61 kB | +105 B (+1%) |
| `dist/chartjs-plugin-annotation.js` | 7.74 kB | +111 B (+1%) |
| `dist/chartjs-plugin-annotation.min.js` | 5.21 kB | +38 B (+1%) |

PR #448, opened 2021-07-29
Add missing animations prop to options interface.

multipolygon:
Hi, I can see `options.animations` being passed to `new Animations(chart, animOpts)` object here:
https://github.com/chartjs/chartjs-plugin-annotation/blob/v1.0.2/src/annotation.js#L140
(and I am actually using that option)

Animation:
https://github.com/chartjs/Chart.js/blob/a8337a3/types/animation.d.ts#L29

AnyObject:
https://github.com/chartjs/Chart.js/blob/a8337a32c976393e4648de633dcc49cdec2760b3/types/basic.d.ts#L2

### DIFF
--- a/types/options.d.ts
+++ b/types/options.d.ts
@@ -65,5 +65,6 @@ interface PointAnnotationOptions extends CoreAnnotationOptions {
 export interface AnnotationPluginOptions extends AnnotationEvents {
 	annotations: AnnotationOptions[] | Record<string, AnnotationOptions>,
 	dblClickSpeed?: Scriptable<number, PartialEventContext>,
-	drawTime?: Scriptable<DrawTime, PartialEventContext>,
+  drawTime?: Scriptable<DrawTime, PartialEventContext>,
+  animations: Record<string, unknown>,
 }


kurkle:
[AnimationSpec](https://github.com/chartjs/Chart.js/blob/a8337a32c976393e4648de633dcc49cdec2760b3/types/index.esm.d.ts#L1556) might be better type for it. 

Could also pass the `TType`, to make the scriptable context properly typed:
https://github.com/chartjs/chartjs-plugin-annotation/blob/2d485b799e04159b5b20b6d6519976a57394e7c5/types/index.d.ts#L6-L8

kurkle (approved):

types/options.d.ts
```diff
@@ -65,5 +65,6 @@ interface PointAnnotationOptions extends CoreAnnotationOptions {
 export interface AnnotationPluginOptions extends AnnotationEvents {
 	annotations: AnnotationOptions[] | Record<string, AnnotationOptions>,
 	dblClickSpeed?: Scriptable<number, PartialEventContext>,
-	drawTime?: Scriptable<DrawTime, PartialEventContext>,
+  drawTime?: Scriptable<DrawTime, PartialEventContext>,
+  animations: Record<string, unknown>,
```
jbelien, line 69: Since this commit, I have the following error in my project:
> error TS2741: Property 'animations' is missing in type '{ annotations: { [key: string]: AnnotationOptions<keyof AnnotationTypeRegistry>; }; }' but required in type 'AnnotationPluginOptions'.

Shouldn't the `animations` property be optional ?
```diff
- animations: Record<string, unknown>,
+ animations?: Record<string, unknown>,
```

I don't find any documentation on this property in https://www.chartjs.org/chartjs-plugin-annotation/latest/ ; am I missing something ?

Thanks!

Issue #441, opened 2021-07-13
Feature Request: Ability to annotate a single bar


RvBurik:
The possibility to draw a line on specific bars. This feature was already implemented in `v2` ([here](https://github.com/chartjs/chartjs-plugin-annotation/commit/be31a56f3ab1682d93ac84531bfd6528aa19302f)), by setting the `onlyForDataIndex` field. This worked on both horizontal and vertical bars. Unfortunately, this feature is currently not available in the newer version of this plugin. It would be great to see it back in `v3`!

Version in screenshot: 
- ChartJS: 2.9.3
- ChartJS-plugin-annotation: 0.5.7

![image](https://user-images.githubusercontent.com/18439206/125470268-1db75896-6e60-4e50-87fe-14282857b492.png)
![image](https://user-images.githubusercontent.com/18439206/125470468-8f82c084-02be-4283-b176-9db7c5011fba.png)

kurkle:
Already possible, though you'll need to manually adjust for categoryPercentage, barPercentage and number of bars per category

https://codepen.io/kurkle/pen/XWgWYab

PR #450, opened 2021-08-03
Issue-441 - Added functionality to position annotations on barcharts

Aukevanoost:
Pull request regarding a fix for [Issue-441](https://github.com/chartjs/chartjs-plugin-annotation/issues/441)

Added 2 functionalities to the line-annotation functionality: 

1) On barcharts the annotation line started halfway the bar, this fix will compensate to start at the begin of the bar (for horizontal and vertical bars) 
2) Added an xIndex and yIndex to replace min and max on barcharts and linecharts. With as final goal to make it easier for users to add line-notations to only one dataset/tick.


src/types/line.js
```diff
@@ -142,9 +144,46 @@ export default class LineAnnotation extends Element {
...
+
+        y -= this._compensateAxisForBarchart(chart, yScale, y, false);
+        y2 -= this._compensateAxisForBarchart(chart, yScale, y2, false);
+      }
     }
     return limitLineToArea({x, y}, {x: x2, y: y2}, chart.chartArea);
   }
+
+  _compensateAxisForBarchart(chart, scale, axis, shouldBeHorizontal) {
+    const isHorizontal = (chart.config.options.indexAxis !== 'y');
```
LeeLenaleee, line 161: Should be a `===` comparrison instead of `!==` because now its giving the opposite result

kurkle:
I think there are some issues:
- the [x|y][Min|Max] values already support index values on a category scale, so those can already be used to draw the lines
- scale might not be the length of chart area edge [sample](https://www.chartjs.org/docs/latest/samples/scales/stacked.html)
- the `indexAxis` and `type` options can be configured per dataset
- scale.ticks.length is not reliable, ticks can be autoSkipped etc.
- adjusting the line position on bar charts is a breaking change

Aukevanoost:
@kurkle First of all, thank you for your feedback. I've fixed some of your remarks: 

[x] the [x|y][Min|Max] values already support index values on a category scale, so those can already be used to draw the lines
[x] scale might not be the length of chart area edge sample
[x] the indexAxis and type options can be configured per dataset
[  ] scale.ticks.length is not reliable, ticks can be autoSkipped etc.
[x] adjusting the line position on bar charts is a breaking change

Unfortunately I was not able to fix 'scale.ticks.length' (yet), I was hoping you could provide a suggestion to use for the calculation of the annotation coordinate location. 

Cheers in advance

Aukevanoost:
As mentioned on #441 It is more robust to use Index instead of a chart-width calculation, therefore the last issue has been resolved using index to calculate the location of the annotation.

docs/guide/options.md
```diff
@@ -89,7 +89,47 @@ The context object contains the following properties:
 * `chart`: the associated chart
 * `type`: `'chart'`
 
-The [chart](#chart) option context is provided by Chart.js. It is passed to scriptable options when resolving annotation `id`, `type` and `drawTime` or adjusting scale ranges in `afterDataLimits` hook. The options resolved at that time are `scaleID`, `xScaleID`, `yScaleID`, `value`, `endValue`, `xMin`, `xMax`, `yMin`, `yMax`, `xValue` and `yValue`.
+The [chart](#chart) option context is provided by Chart.js. It is passed to scriptable options when resolving annotation `id`, `type` and `drawTime` or adjusting scale ranges in `afterDataLimits` hook. The options resolved at that time are `scaleID`, , `value`, `endValue`, `xMin`, `xMax`, `yMin`, `yMax`, `xValue` and `yValue`.
```
kurkle, line 92: Why did you remove `xScaleID` and `yScaleID`?

kurkle (changes requested):
Sorry for the stalls in reviews.

This needs a rebase.

Current implementation would not work in some situations, for example: mixed stacked charts.
I think it would be safest to get the x, width / y, height from the actual bar element (metaset.data[index]), where the values would always be calculated considering everything.

The bar element has those props: https://www.chartjs.org/docs/latest/api/interfaces/BarProps.html

As we are trying to get the final values (after animation), [getProps](https://www.chartjs.org/docs/latest/api/interfaces/BarElement.html#getprops) should be used with the `final` set to true.

Something like:
```js
function getBarMeta(chart, annotationFor) {
  const meta = defined(annotationFor) && chart.getDatasetMeta(annotationFor);
  return meta.type === 'bar' ? meta : null;
}

resolveElementProperties(chart, options) {
...
const meta = getBarMeta(chart, options.annotationFor);
if (meta) {
  const bar = meta.data[options.index];
  if (bar.horizontal) { 
    const {x: centerX, width} = bar.getProps(['x', 'width'], true);
    x = centerX - width / 2;
    x2 = centerX + width / 2;
  } else {
    const {y: centerY, height} = bar.getProps(['y', 'height'], true);
    x = centerY - height / 2;
    x2 = centerY + height / 2;
  }
}
if (scale) {
...
```

Also I think the new code should only be executed if `annotationFor` is set. And I think it would be ok to only accept the index (and not label or any other value) for it.
It can be scriptable anyway:
`annotationFor: (ctx) => ctx.chart.data.dataset.findIndex(ds => ds.label === 'foobar')`

It would not hurt to make it work for box and ellipse annotations too.. so putting that in a helper and replacing the x, y, x2, y2 initialization with that would do the trick:

```js
let {top: y, left: x, bottom: y2, right: x2} = getDefaultCoords(chart, options);
```
and the getDefaultCoords would be something like:
```js
function getDefaultCoords(chart, options) {
  let {top: y, left: x, bottom: y2, right: x2} = chart.chartArea;
  const meta = getBarMeta...
  ...
  return {x, y, x2, y2};
}
```

kurkle:
Closing as stale. Will reopen if you get back to this.

PR #451, opened 2021-08-16
Pass event instance to listeners

stockiNail:
Currently the listeners on the annotations are receiving only the context.

I think it could be helpful to get also the event instance, separately, as additional argument.

A use case where to use the event could be to check if any modifier key is pressed in order to perform own logic only when a key is pressed.

### DIFF
--- a/docs/guide/interaction.md
+++ b/docs/guide/interaction.md
@@ -15,7 +15,7 @@ The following options are available for all annotation types. These options can 
 
 | Name | Type | [Scriptable](options#scriptable-options) | Notes
 | ---- | ---- | :----: | ----
-| `enter` | `(context) => void` | No | Called when the mouse enters the annotation.
-| `leave` | `(context) => void` | No | Called when the mouse leaves the annotation.
-| `click` | `(context) => void` | No | Called when a single click occurs on the annotation.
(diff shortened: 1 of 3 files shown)


kurkle:
Why not set the event in the context? also I'd pass the ChartEvent, because it contains the canvas coordinates and provides the native event inside.

stockiNail:
> Why not set the event in the context? 

I thought that in the near future the context in the plugin will be "persistent" as in chartjs and in datalabels plugin in order to store own data and to get them in further calls.
In this case it couldn't make sense to add it to the context because you could have a property related to an obsolete object (last managed event). Of course, I can change it and add it to the context. Let me know.

> also I'd pass the ChartEvent, because it contains the canvas coordinates and provides the native event inside.

Yesterday I was used the ChartEvent but today I recognized that the types were not correct because I set to Event.
Of course, I'll change in next minutes.

kurkle:
> I thought that in the near future the context in the plugin will be "persistent" as in chartjs and in datalabels plugin in order to store own data and to get them in further calls.
> In this case it couldn't make sense to add it to the context because you could have a property related to an obsolete object (last managed event). Of course, I can change it and add it to the context. Let me know.

Makes sense

kurkle (approved):

PR #457, opened 2021-08-22
Per corner cornerRadius for box and line label

### LINKED ISSUE #437: Feature Request: Expand Box annotation cornerRadius functionality to be able to provide value per corner
Right now I really love the corner radius functionality, but it would be fantastic to be able to be able to specify a value for each corner similar to how the CSS property `border-radius` behaves (see screenshot).  I would propose the ability to make cornerRadius of type `number | { topLeft: number, topRight: number, bottomLeft: number, bottomRight: number }` to allow for this functionality. I would be happy to attempt to contribute this feature myself, but would love a go ahead thumbs up from one of the maintainers before beginning work. Thanks for all your great work!

Screen Shot 2021-07-01 at 3 50 41 PM

kurkle:
Closes #437

This feature requires Chart.js v3.1 where `addRoundedRectPath` was introduced (missing from type definitions though).

The fixtures including borderDash needed to be updated as `addRoundedRectPath` generates the path counterclockwise. Also some differences between firefox and chrome on how those are rendered, that's why the added tolerance.


test/fixtures/box/cornerRadius.js
```diff
@@ -26,7 +26,7 @@ module.exports = {
               yMax: 10,
               backgroundColor: 'rgba(33, 101, 171, 0.5)',
               borderColor: 'rgb(33, 101, 171)',
-              cornerRadius: 8,
+              cornerRadius: {topLeft: 16, topRight: 8, bottomRight: 4},
```
etimberg, line 29: What happened to the purple annotation in this test? the new image appears to be missing it

kurkle: No idea, need to take a look.

kurkle: It was caused by unclamped infinite value

etimberg (approved):

PR #473, opened 2021-11-09
Adds "label" object to box annotation

### LINKED ISSUE #42: Box annotation labels
AFAICT, labels are not supported on box annotations, it would be nice if they were added.

stockiNail:
Fixes  #42

Creates new `label` element to enable the label in the box  annotation.

Adds the following options:

 * **enabled**: boolean which defines if the label must be shown.
 * **drawTime**: string to define when the label will be drawn.
 * **align**: 'left' | 'center' | 'right', defines the horizontal alignment of the label, inside the element. 
 * **color**: defines the color to apply to the label.
 * **font**: defines the font to apply to the label.
 * **position**: 'top' | 'middle' | 'bottom', defines the vertical alignment of the label, inside the element. 
 * **xPadding**: number which defines the amount of pixels to apply to the label at X position.
 * **yPadding**: number which defines the amount of pixels to apply to the label at Y position.
 * **content**: string or array of strings (multiline) to define the content of the label

- [X] test cases
- [x] documentation


stockiNail:
`textAlign` options and `content` as image will be implemented later, need more time

test/fixtures/box/labelMultiline.js
```diff
@@ -0,0 +1,105 @@
...
+              yMax: 20,
+              backgroundColor: 'rgba(178, 255, 102, 0.5)',
+              borderColor: 'rgb(178, 255, 102)',
+              borderDash: [6, 6],
+              borderWidth: 5,
+              label: {
+                enabled: true,
+                content: ['This is a label', 'but this is multiline'],
+                font: {
+                  size: 18
```
kurkle, line 71: The font used by `spriteText` is 16 pixels and is not scaled, so maybe using that as size would work better?

types/boxLabel.d.ts
```diff
@@ -0,0 +1,46 @@
+import { Color, FontSpec } from 'chart.js';
+import { PartialEventContext } from './events';
+import { DrawTime, Scriptable } from './options';
+
+export interface BoxLabelOptions {
+	drawTime?: Scriptable<DrawTime, PartialEventContext>,
+	font?: FontSpec
+	color?: Scriptable<Color, PartialEventContext>,
+	/**
```
kurkle, line 9: nit: inconsistent style, most comment blocks have an empty line before, and tabs instead of spaces (linter is misconfigured as it does not detect these)

types/options.d.ts
```diff
@@ -1,75 +1,78 @@
...
-}
+import { Color } from 'chart.js';
+import { AnnotationEvents, PartialEventContext } from './events';
+import { LabelOptions } from './label';
+import { BoxLabelOptions } from './boxLabel';
+
+export type DrawTime = 'afterDraw' | 'afterDatasetsDraw' | 'beforeDraw' | 'beforeDatasetsDraw';
+
+export interface AnnotationTypeRegistry {
+	line: LineAnnotationOptions
```
kurkle, line 9: mixed tabs and spaces in this file

types/boxLabel.d.ts
stockiNail: Do you have the config to apply to my lint locally? I always run it to check these kinds of issues

test/fixtures/box/labelMultiline.js
stockiNail: Doesn't change too mucch therefore I think it's better to remove font options.

types/boxLabel.d.ts
kurkle: Fixed in master, rebase to get the new config

kurkle:
| `dist/chartjs-plugin-annotation.esm.js` | 7.97 kB | +386 B (+5%) | 🔍 |
| `dist/chartjs-plugin-annotation.js` | 8.09 kB | +387 B (+5%) | 🔍 |
| `dist/chartjs-plugin-annotation.min.js` | 5.67 kB | +282 B (+5%) | 🔍 |

stockiNail:
Closed in favor of #502

PR #477, opened 2021-11-11
Add new LABEL annotation type

stockiNail:
Fixes   #26

Creates new `LabelAnnotation` element to enable to render a text on the chart at a coordinate, calculated by the chart data values.

Sets the following options:

| Name | Description |
| ---- | ---- |
| `align` | Specifies the text horizontal position of the label with respect to the selected point.
| `adjustScaleRange` | Should the scale range be adjusted if this annotation is out of range.
| `backgroundColor` | Fill color.
| `borderColor` | Stroke color.
| `borderWidth` | Stroke width.
| `borderDash` | Length and spacing of dashes.
| `borderDashOffset` | Offset for line dashes.
| `borderRadius` | Radius of the corners of the annotation.
| `color` | Text color.
| `content` | The content to show in the text annotation.
| `display` | Whether or not this annotation is visible.
| `drawTime` | Determines where in the chart lifecycle the drawing occurs.
| `font` | Text font.
| `position` | Specifies the text vertical position of the label with respect to the selected point.
| `textAlign` | Text alignment of label content when there's more than one line.
| `xAdjust` | Adjustment along x-axis (left-right) of label relative to computed position.
| `xPadding` | Padding of label to add left/right.
| `xScaleID` | ID of the X scale to bind onto.
| `xValue` | X coordinate of the point in units along the x axis.
| `yAdjust` | Adjustment along y-axis (top-bottom) of label relative to computed position.
| `yPadding` | Padding of label to add top/bottom.
| `yScaleID` | ID of the Y scale to bind onto.
| `yValue` | Y coordinate of the point in units along the y axis.

Being new annotation type, you have the events on that as well.

- [X] test cases
- [x] documentation

PR #481, opened 2021-11-15
Add LABEL annotation and label node to BOX annotation 

stockiNail:
Refactoring to reduce the duplicate code of PR #473 and #477


package-lock.json
```diff
@@ -1,18580 +1,8 @@
 {
   "name": "chartjs-plugin-annotation",
   "version": "1.0.2",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
kurkle, line 4: very old npm?

docs/samples/types/box.md
```diff
@@ -1,118 +1,123 @@
...
-    }
-  }
-];
-
-module.exports = {
-  actions: actions,
-  config: config,
-};
-```
+# Box
```
kurkle, line 1: Line ending change (could be valid, did not check)

test/fixtures/box/label.js
```diff
@@ -0,0 +1,102 @@
+module.exports = {
```
kurkle, line 1: Looking at the fixture, I think the label should be clipped inside the border (not draw halfway over it)

test/fixtures/label/borderDash.js
```diff
@@ -0,0 +1,64 @@
...
+            text2: {
+              type: 'label',
+              xScaleID: 'x',
+              yScaleID: 'y',
+              xValue: 'February',
+              yValue: 11,
+              backgroundColor: 'rgba(101, 33, 171, 0.5)',
+              borderColor: 'rgb(101, 33, 171)',
+              borderDash: [6, 6],
+              borderWidth: 1,
```
kurkle, line 40: Should make the border wider for this test, so any regression is more likely catched

test/fixtures/label/borderRadius.js
```diff
@@ -0,0 +1,63 @@
...
+            },
+            text2: {
+              type: 'label',
+              xScaleID: 'x',
+              yScaleID: 'y',
+              xValue: 'February',
+              yValue: 11,
+              backgroundColor: 'rgb(245,245,245)',
+              borderDash: [6, 6],
+              borderWidth: 1,
```
kurkle, line 38: Same here, wider border (and not dash) would make this test more likely to work

test/fixtures/label/contentMultiline.js
```diff
@@ -0,0 +1,59 @@
...
+          annotations: {
+            text1: {
+              type: 'label',
+              xScaleID: 'x',
+              yScaleID: 'y',
+              xValue: 'January',
+              yValue: 20,
+              backgroundColor: 'transparent',
+              borderWidth: 0,
+              content: ['This is my text, row 1', 'This is my text, row 2', 'This is my text, row 3'],
```
kurkle, line 27: Should have different length lines for this test

test/fixtures/label/position.js
```diff
@@ -0,0 +1,59 @@
...
+          annotations: {
+            text1: {
+              type: 'label',
+              xScaleID: 'x',
+              yScaleID: 'y',
+              xValue: 'January',
+              yValue: 20,
+              backgroundColor: 'transparent',
+              borderWidth: 0,
+              content: 'This is my text'
```
kurkle, line 27: Could use the position value as label, for example: `'poistion: undefined'`
Could also place all on same value and maybe make this a line chart with a line on that value, so there would be a reference for the eye.

I'm not 100% `position` is the correct property name for this functionality?

types/options.d.ts
```diff
@@ -1,75 +1,81 @@
...
-  xValue?: Scriptable<ScaleValue, PartialEventContext>;
-  yValue?: Scriptable<ScaleValue, PartialEventContext>;
-}
-
-export interface AnnotationPluginOptions extends AnnotationEvents {
-  annotations: AnnotationOptions[] | Record<string, AnnotationOptions>,
-  dblClickSpeed?: Scriptable<number, PartialEventContext>,
-  drawTime?: Scriptable<DrawTime, PartialEventContext>,
-  animations: Record<string, unknown>,
+import { Color } from 'chart.js';
```
kurkle, line 1: looks like another line end change, is it valid? I guess the linter should validate these.

src/types/labelItem.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+    return widthCache.get(mapKey);
+  }
+}
+
+function getImageSize(size, value) {
+  if (typeof value === 'number') {
+    return value;
+  } else if (typeof value === 'string') {
+    return toPercent(value) * size;
```
kurkle, line 70: A string not ending with '%' would end up being undefined?

kurkle (commented):
Couple of random comments, will review the actual pattern later.

package-lock.json
stockiNail: my fault. I'm gonna replace with the master version

src/types/labelItem.js
stockiNail: No, it isn't. When a string is not ending with % (or is not a string) the function returns false (therefore 0).
Anyway this function is used in the line annotation for long time.

stockiNail:
The line annotation is not touched at the moment. Nevertheless it should changed in order to leverage on Label object.
One of codeClimate issues is related to duplication of `getImage` function which could be solved when also LineAnnotation will use Label.

kurkle:
| `dist/chartjs-plugin-annotation.esm.js` | 8.77 kB | +1.17 kB (+15%) | ⚠️ |
| `dist/chartjs-plugin-annotation.js` | 8.9 kB | +1.18 kB (+15%) | ⚠️ |
| `dist/chartjs-plugin-annotation.min.js` | 6.3 kB | +873 B (+16%) | ⚠️ |

PR #482, opened 2021-11-16
Added note about cornerRadius

sdapkus:
Hi!

I've spend some time figuring out borderRadius option. Docs are misleading at a moment, because `master` branch has change, that can accept `borderRadius` or `cornerRadius` but last tagged version (1.0.2) only accepts `cornerRadius` and docs does not have anything about `cornerRadius`.

Not sure how it is best to write this to docs, feel free to update my PR

### DIFF
--- a/docs/guide/types/box.md
+++ b/docs/guide/types/box.md
@@ -65,7 +65,7 @@ The following options are available for box annotations.
 | [`borderDash`](#styling) | `number[]`| Yes | `[]`
 | [`borderDashOffset`](#styling) | `number`| Yes | `0`
 | [`backgroundColor`](#styling) | [`Color`](../options#color) | Yes | `options.color`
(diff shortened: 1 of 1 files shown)


kurkle:
Thanks and sorry for the wasted time.

I think we should have versioned documentation.

I think it would be better to just push 1.1.0 out now and work out the documentation versioning before pushing more changes to master.

@etimberg thoughts?

LeeLenaleee:
Should the same be done for the zoom plugin also, since that one also is missing versioned and a new version just got released there too so it would be best time to do it there too.

kurkle:
Closing as we now have versioned docs this info can be found there:
https://www.chartjs.org/chartjs-plugin-annotation/1.0.2/guide/types/box.html#configuration

PR #503, opened 2021-11-23
Add POLYGON annotation

### LINKED ISSUE #431: Feature Request: Polygon Annotation
To mark some areas with different kind of shapes (like triangle etc.), it would be nice to have some polygon annotations.

stockiNail:
Fixes #431

Creates new `PolygonAnnotation` element to enable to render whatever polygin on the chart at a coordinate, calculated by the chart data values.

poligon

### TODO

- [x] must be updated in order to reuse common code of #502 
- [x] test cases
- [x] types definition
- [x] documentation


old_samples/polygon.html
```diff
@@ -0,0 +1,189 @@
+<!doctype html>
```
kurkle, line 1: I'd rather not add to old_samples anymore, did you have some reason?

stockiNail: I usually use these samples as first testing. I can remove it.

kurkle (approved):

PR #506, opened 2021-11-26
Enable label padding as object

stockiNail:
This PR enables the feature to set the padding option for labels on the annotations by an object.

See CHARTJS doc: https://www.chartjs.org/docs/latest/general/padding.html

`xPadding` and `yPadding` options have been marked as **deprecated**.

### TODO

- [x] test cases
- [x] types definition
- [x] documentation

PR #507, opened 2021-11-28
Add callout and point nodes to the label annotation

stockiNail:
This PR adds `callout` and `point` nodes in the LABEL annotation in order to see the relation between the selected point and the label.
Often applying a label annotation, you can not see to which point is referring to and a simple callout can help or drawing the point.

callout

The callout and point are nodes of label annotation options:

```javascript
{
  type: 'label',
  ...
  callout: {
    // callout options
  },
  point: {
    // point options
  },
  ...
}
```

### TODO

- [x] test cases
- [x] documentation
- [x] types definition
- [x] waiting for #506 merging


kurkle:
Wondering if point element from core should be used to draw the point.

stockiNail:
> Wondering if point element from core should be used to draw the point.

This PR is still a working in progress. The answer is yes but, having decided to avoid the inheritance among elements, I have in my todo to create a `drawPoint` function in `helpers`. Probably also `inRange` logic could be common with a specific function in the helpers. More in the next days.

Furthermore (not in this PR), I would like to review the current exported `helpers` function in order do not use element methods (i.e. `getProps`) but to enable the usage of that functions by objects... Maybe could be helpful when (but also `if`) the custom annotation elements will be implemented (I have a prototype also for that but, first, the topic needs a discussion).

stockiNail:
> Wondering if point element from core should be used to draw the point.

DONE!

stockiNail:
@kurkle I'd like to wait for PR #506 approval before giving it ready for review. What do you think?

stockiNail:
I was thinking that point could deserve a own node. Going to this direction, this enables the point drawing, also without callout, improving the user experience, seeing the point anyway.

```javascript
{
  type: 'label',
  ...
  callout: {
    // callout options
  },
  point: {
    // point options
  },
  ...
}
``` 

Sample how the point can be drawn without callout:

Screenshot 2021-11-30 105442

What do you think? Make sense?

stockiNail:
I see the Node.js CI / build (windows-latest)  failed for timeout...

kurkle:
This needs a rebase now.
Some samples would be good btw.

stockiNail:
> Some samples would be good btw.

I use the samples in that folder as "first" functional test. When I'm coding, it's really comfortable to test by a web page that you can run when you want (with console) instead to use the test cases.

stockiNail:
> This needs a rebase now.

Done!

kurkle (approved):

Issue #468, opened 2021-10-19
Plugin breaks after specifying xAxis scale


exo-pla-net:
If you specify config.options.scales.xAxis = {}, it will break without an error. No annotations will appear.

If you specify config.options.scales.x = {}, it works fine.

If you specify config.options.scales.yAxis = {}, it works fine.

There's something weird going on with config.options.scales.xAxis.

kurkle:
Nope?

![image](https://user-images.githubusercontent.com/27971921/137841778-9c1944a8-e47e-4dbe-8154-160cecab3679.png)

kurkle:
See `xScaleID`: 
https://www.chartjs.org/chartjs-plugin-annotation/guide/types/box.html#configuration

PR #518, opened 2021-12-02
Throw error when non-existing scale id is used

kurkle:
Closes: #468

Updated the Chrome flags so it would not timeout when in background.


etimberg (approved):

stockiNail:
@kurkle I think this should be documented. Sorry if I recognized it just now.

stockiNail:
@kurkle maybe I'm wrong but it should be marked as breaking change. I think before the chart was shown anyway.

kurkle:
@stockiNail the issue was marked as bug, so this can be considered to be a bug fix.
Although this could be breaking in some use cases, the annotations did not always show before either (see the issue).

Should be mentioned in the release notes when 1.2 is released, but I'm not sure this needs to be documented. I don't think it was never a valid use case to specify a non-existing scale id for an annotation.

stockiNail:
@kurkle thank you! Agree!

stockiNail:
Sorry @kurkle...  I don't want to boring you, but having a look to the doc (for instance box annotation, https://www.chartjs.org/chartjs-plugin-annotation/latest/guide/types/box.html#configuration), I think the sentence 

_If one of the axes does not match an axis in the chart, the box will take the entire chart dimension._

should be removed. If ok for you, I can do it.

kurkle:
Oh, did not see that. Looks like that actually is a supported use case, so this is a breaking change. I'll deal with it tomorrow.

PR #519, opened 2021-12-02
Move annotation types handling to types/index.js

kurkle:
As the number of types grow, they deserve their own index.

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -1,30 +1,10 @@
-import {Animations, Chart, defaults} from 'chart.js';
+import {Animations, Chart} from 'chart.js';
 import {clipArea, unclipArea, isFinite, valueOrDefault, isObject, isArray} from 'chart.js/helpers';
 import {handleEvent, hooks, updateListeners} from './events';
-import BoxAnnotation from './types/box';
-import LineAnnotation from './types/line';
-import EllipseAnnotation from './types/ellipse';
-import LabelAnnotation from './types/label';
-import PointAnnotation from './types/point';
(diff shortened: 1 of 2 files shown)


stockiNail:
> @stockiNail any thoughts about this change?

yes even if I didn't have the courage to write it ...

It seems your going to expose the annotations in order to enable the custom ones. Or am I completely wrong?

kurkle:
This is a step that would make it easier to allow custom annotation types. I'm still not sure it that is a good solution or if a generic CustomAnnotation would be enough. But that aside, I think this makes the source cleaner. There are also ways (exports in package.json) to indicate these exports are only for internal use.

stockiNail:
I don't have enough skills on javascript packaging to lead that... Sorry.

Nevertheless, with my poor experience, it could be enough to expose the exported helpers functions and the methods of the classes. Nothing more. Maybe I'm wrong but it doesn't seem too dangerous.
But you know better than me other techniques how to isolated some functions. I'm more than happy to follow you and to learn.

About cleanup of the code, I fully agree. I have never asked but is there any guidelines how the code should look like? For instance, the classes on the top and all "private" functions on the bottom?

stockiNail:
Anyway the definition of the interface to extend annotation element is very important, I think.

etimberg (approved):

stockiNail:
@kurkle I'm going to start for PR to change and continue developing the prototype that I have started last week. Just a matter of time. For sure I'll come back to you.
Before I'd like to close the my open issues and therefore we could have more time to continue thinking how to implement it.

PR #520, opened 2021-12-02
Add test cases to coverage all events cases

stockiNail:
This PR adds test cases on events for all annotation types.

The display option test spec has been review in order to perform the test against all annotation types.

### TODO

- [x] common test case for enter event for all annotation types
- [x] common test case for click event for all annotation types
- [x] common test case for dblclick event for all annotation types
- [x] common test case for leave event for all annotation types


kurkle:
Looks like a good idea 👍

stockiNail:
> Looks like a good idea 👍

Thank you!! I don't know if you have seen but from version 1.1.0 to the current master, the coverage has been increased from 9% to 86%!! Very good (I guess)!

kurkle:
> Thank you!! I don't know if you have seen but from version 1.1.0 to the current master, the coverage has been increased from 79% to 86%!! Very good (I guess)!

Yes, very good!

stockiNail:
It's not clear to me how coveralls is checking the tests.

Example: Ellipse

image

It seems that `inRange` is never tested but that's not true because is invoked for events and there is a special test case to test it directly....

It's weird.

kurkle:
> It's not clear to me how coveralls is checking the tests.
> 
> Example: Ellipse
> 
> image
> 
> It seems that `inRange` is never tested but that's not true because is invoked for events and there is a special test case to test it directly....
> 
> It's weird.

I'll have to look at this closer, to see why that happens.

stockiNail:
> I'll have to look at this closer, to see why that happens.

II think you shouldn't take time for that. I was just curious about this weird behavior. Anyway the result is quite good, because now the master reached 87%, this PR will add another 4.9% therefore I expect to exceed 91/92% (+12/13% from version 1.1).

test/specs/ellipse.spec.js
```diff
@@ -1,3 +1,16 @@
 describe('Ellipse annotation', function() {
   describe('auto', jasmine.fixtures('ellipse'));
+
+  window.testEvents({
+    type: 'ellispe',
```
kurkle, line 5: Took a closer look anyway and here is the problem with ellipse (its not ellispe) :)

kurkle: This leads me to think that we should at least output a warning when the type is not found.

stockiNail: azz!!!! THANK YOU!

stockiNail: fixed

stockiNail: > This leads me to think that we should at least output a warning when the type is not found.

let me add it

kurkle (approved):

test/specs/ellipse.spec.js
kurkle: Lets do that separately

Issue #511, opened 2021-11-30
Clipped area on chart cuts the label annotation


stockiNail:
Before drawing every annotation, the plugin clips an area equals to the chart one.

With LABEL annotation, positioning it with respect of the point, could be cut because should be drawn outside of the chart area, due to its calculated dimension.

![Screenshot 2021-11-30 185937](https://user-images.githubusercontent.com/11741250/144102876-9db3de69-1e81-4482-aaea-cd7bcf0a8fec.png)

For this reason, the dimension of the label annotation should take care about the available space around, in the chart area.

kurkle:
I think both are valid use cases, so it should be behind an option.

For example when zooming, one could want the label to be clipped at the chart area border.

stockiNail:
> For example when zooming, one could want the label to be clipped at the chart area border.

Ok, therefore we could add a boolean option (for instance `clip`) and then if set to `true`, this behavior is accepted, if `false`, the label is adjusted to the chart area.
Something like that?

PR #523, opened 2021-12-02
Add clip option

kurkle:
Closes: #511

Update the page title from Interactions to Configuration as it seems to be the only place that lists the top level options available.


docs/.vuepress/config.js
```diff
@@ -32,13 +32,15 @@ module.exports = {
               {
                 text: 'Development (master)',
                 link: '/chartjs-plugin-annotation/master/',
+                target: '_self',
```
etimberg, line 35: What does this do?

etimberg (approved):

docs/.vuepress/config.js
kurkle: Opens in same tab. This does not change for old versions of docs though.

PR #526, opened 2021-12-03
Log a warning instead of throwing when non-existing scale id is used

kurkle:
This reverts commit f2e2481b23a52938fe6992aa618a2488d8e38f92.

As pointed out by @stockiNail there is documentation about behavior with missing scale, thus making the change breaking.
Can not have that in 1.2.0.


stockiNail:
@kurkle if I may, maybe we could also cleanup the code, where there are some `if` statement to check if the scale instance is consistent. These checks shouldn't be needed anymore.

For instance:

https://github.com/chartjs/chartjs-plugin-annotation/blob/92a89907e154240999563f8d63f48756d7e68b39/src/types/line.js#L145-L153

LeeLenaleee:
What you can do is instead of throwing an error log a warning to the console, something like this:

`No scale found with ${Id} for annotation ${annotation}, while this could be intended it is likely you misconfigured something`

So you still have some feedback for the users if things don't work. Since I have learned that reading documentation is the last thing a lot of users do.

LeeLenaleee (approved):

etimberg (approved):

PR #521, opened 2021-12-02
Add pointStyle and rotation to point options (label and point annotations)

stockiNail:
Fix #510

This PR enables the features to set the point style and its rotation to the point and label annotation.

star

See CHARTJS doc: https://www.chartjs.org/docs/latest/configuration/elements.html#point-styles

### TODO

- [x] test cases
- [x] types definition
- [x] documentation

PR #527, opened 2021-12-03
Add point positioning by box options for point, polygon and label annotations

stockiNail:
This PR is adding the feature to locate the point to start drawing a point, a polygon or a label, by box options (xMin, xMax, yMin, yMax).

The radius, is not a number or equals to zero, is calculated by the minimum value between width and height of the box.

Renames the `getCircleCenterPoint` helpers function to  `getElementCenterPoint`.

Changes parameters to the `getRectCenterPoint` helpers function. Now it accepts a rect (`{x: number, y: number, width: number, height: number}`) and not an element and the flag to use the final position anymore.

### TODO

- [x] box location for point annotation
- [x] box location for polygon annotation
- [x] box location for label annotation
- [x] test cases
- [x] types
- [x] documentation
- [x] waiting for #521 approval and merge


src/helpers.js
```diff
@@ -256,7 +256,7 @@ export function inPointRange(point, center, radius) {
   return (Math.pow(point.x - center.x, 2) + Math.pow(point.y - center.y, 2)) <= Math.pow(radius, 2);
 }
 
-export function getCircleCenterPoint(element, useFinalPosition) {
+export function getCenterPointElement(element, useFinalPosition) {
```
kurkle, line 259: I'd call this `getElementCenterPoint`

types/options.d.ts
```diff
@@ -88,7 +88,7 @@ export interface EllipseAnnotationOptions extends CoreAnnotationOptions, Annotat
   rotation?: Scriptable<number, PartialEventContext>
 }
 
-export interface PointAnnotationOptions extends CoreAnnotationOptions, AnnotationPointCoordinates {
+export interface PointAnnotationOptions extends CoreAnnotationOptions, AnnotationCoordinates, AnnotationPointCoordinates {
```
kurkle, line 91: Should the coordinate types be merged?

src/helpers.js
stockiNail: Fine!

types/options.d.ts
stockiNail: Do you mean in a unique interface? No, because box and ellipse are using only x/yMin and x/yMax and not x/yValue, bu I could extend PointCoordinates from Coordinates. Ok for you?

kurkle: Was just a thought, ok to leave as is.

stockiNail: have a look! I have change a bit with extends, therefore you don't have some many extends in the annotation options

LeeLenaleee (approved):

docs/guide/types/point.md
```diff
@@ -54,30 +54,42 @@ The following options are available for point annotations.
 | [`borderColor`](#styling) | [`Color`](../options#color) | Yes | `options.color`
 | [`borderDash`](#styling) | `number[]`| Yes | `[]`
 | [`borderDashOffset`](#styling) | `number`| Yes | `0`
+| [`backgroundColor`](#styling) | [`Color`](../options#color) | Yes | `options.color`
```
kurkle, line 57: duplicate backgrondColor

types/options.d.ts
```diff
@@ -43,7 +43,7 @@ interface AnnotationCoordinates {
   yMin?: Scriptable<ScaleValue, PartialEventContext>,
 }
 
-interface AnnotationPointCoordinates {
+interface AnnotationPointCoordinates extends AnnotationCoordinates{
```
kurkle, line 46: Missing space between `AnnotationCoordinates` and `{`
Should take another look at the linter, it should catch this kind of stuff.

stockiNail: added blank before {

docs/guide/types/point.md
stockiNail: duplicated definition has been removed

stockiNail:
`test/fixtures/label/clip-false.js FAILED` coming from another PR. Let me have a look

stockiNail:
npm testlog, run locally, exit code 0: https://gist.github.com/stockiNail/2245e56814b1c01233ccce02c63a756f
No clear why here it failed

src/types/label.js
```diff
@@ -13,10 +13,8 @@ export default class LabelAnnotation extends Element {
       const adjust = gap || 0;
       const {x, y, width, height} = this.isBoxVisible() ? this.getProps(['x', 'y', 'width', 'height']) : this.labelRect;
 
-      return mouseX >= x + adjust &&
```
kurkle, line 16: I assume you changed these for code climate? I think the old formatting was easier to read.

stockiNail:
rerun jobs and everything went well!

src/helpers.js
```diff
@@ -270,3 +270,26 @@ export function getChartCircle(chart, options) {
...
+  if (!isBoundToPoint(options)) {
+    const box = getChartRect(chart, options);
+    const point = getRectCenterPoint(box);
+    let radius = options.radius;
+    if (!radius || isNaN(radius)) {
+      radius = Math.min(box.width, box.height) / 2;
+      options.radius = radius;
+    }
+    return {
+      x: point.x,
```
kurkle, line 288: This makes me think if we should make all of the annotations be "boxes" in v2.
So each would have x,y,width,height and maybe centerX/centerY properties. 
Not sure what the benefirts would be other than consistency of the x/y between elements. Now its either top left corner or center depending of the type.

test/fixtures/polygon/boxLocation.js
```diff
@@ -0,0 +1,78 @@
...
+          min: -10,
+          max: 10
+        }
+      },
+      plugins: {
+        legend: false,
+        annotation: {
+          drawTime: 'afterDraw',
+          annotations: {
+            triangle1: {
```
kurkle, line 23: I'm sure there is a good explanation, but why aren't the triangles centered vertically to the boxes?

src/types/label.js
stockiNail: Yes, I prefer the previous one as well. How can I ignore the CC issue?

kurkle: Cc issues are just fyi, we don't require it to pass. There is also possibility to approve the issues in cc (I've sometimes done that)

src/helpers.js
stockiNail: When I was writing this PR, I see, for instance, box and line with x2 and y2 that could be useless, having x, y, width and height.
At the end of the day, all annotations are boxes, the line as well. I think it could make sense to normalize it. It could also help for the custom one. And anyway, every annotation could add own properties if needed, of top of them.

kurkle: Agreed. Care to write an issue about that for V2?

test/fixtures/polygon/boxLocation.js
stockiNail: let me do another test before answering just to be sure about what I'll write

kurkle: After giving it thought, it seems obvious the center of the points is vertically centered, not the outer bounds of the triangle that I was looking at.

stockiNail: I tested which is confirming my answer. The triangle is built using the center of the box and the circle built inside the box with radius (in this case), in this case the min between height and width of the box (because radius is NaN).

Here the result drawing the circle too:

boxes

I have changed the test case in order to highlight also the circle.

kurkle: Good, the added circle clarifies it nicely without having to thing about it :)

src/helpers.js
stockiNail: @kurkle maybe this could be the content for new PR. What do you think?

kurkle: I think it would be a breaking change and we should do it for annotation plugin v2. So at this point I'd just write and issue about it so we remmeber to do it.

src/types/label.js
stockiNail: reverted to original format

kurkle (approved):

stockiNail:
@kurkle being new as collaborator, having the approval, you gave it, can I merge by myself?
Sorry for stupid question. :(

kurkle:
> @kurkle being new as collaborator, having the approval, you gave it, can I merge by myself? Sorry for stupid question. :(

That is not a stupid question. There are number of things to consider before merging.
 - even if approved, should this be included in next release (=yes in this case)
 - is anything pending, comments, reviews etc (=no in this case)
 - consider the order of merging, to minimize the need to rebase (no other pull requests about to be merged currently, so not needed in this case)
 - its good to wait for any running actions to finish before merging (increases changes of success)

In this case I did not merge it directly, in case you had something to add still. So yes, you can merge it 👍

kurkle:
Btw, do you have slack? it would be easier to exchange some random thoughts in a chat.

stockiNail:
> Btw, do you have slack? it would be easier to exchange some random thoughts in a chat.

Yes I have it but... I can not use it for this stuff.. It's a long story. Anyway Let me take time to check if and how I can use it.

stockiNail:
> > @kurkle being new as collaborator, having the approval, you gave it, can I merge by myself? Sorry for stupid question. :(
> 
> That is not a stupid question. There are number of things to consider before merging.
> 
>     * even if approved, should this be included in next release (=yes in this case)
> 
>     * is anything pending, comments, reviews etc (=no in this case)
> 
>     * consider the order of merging, to minimize the need to rebase (no other pull requests about to be merged currently, so not needed in this case)
> 
>     * its good to wait for any running actions to finish before merging (increases changes of success)
> 
> 
> In this case I did not merge it directly, in case you had something to add still. So yes, you can merge it 👍

Thank you very much! They will be more rules for next time!

PR #528, opened 2021-12-04
Use the context of element for events dispatching

stockiNail:
This PR changes the context instance to pass to event hook when the event is dispatched.

Instead of creating new object every time, the `element.$context` is used.

This enables to the users to store custom properties and to retrieve them in other event hooks.

### TODO

- [x] test case, waiting for #520 approval and merge


src/events.js
```diff
@@ -95,22 +95,22 @@ function handleClickEvents(chart, state, event, options) {
...
-      dispatchEvent(chart, click, element, event);
+      dispatchEvent(click, element, event);
     }
   }
 }
 
-function dispatchEvent(chart, handler, element, event) {
-  callHandler(handler, [{chart, element}, event]);
+function dispatchEvent(handler, element, event) {
+  callHandler(handler, [element.$context, event]);
```
kurkle, line 113: Does the element **always** contain `$context`?
I'd think it does not when there is nothing scriptable, but that maybe not the case?

kurkle: Answering to myself, its always there. And it set there (in `updateElements`):

https://github.com/chartjs/chartjs-plugin-annotation/blob/405433949306d949e4460e8e65487b4baa650346/src/annotation.js#L142

kurkle (approved):

test/events.js
```diff
@@ -1,4 +1,9 @@
+const annotationTypes = ['box', 'ellipse', 'label', 'line', 'point', 'polygon'];
+
 export function testEvents(options, innerElement) {
+  if (!annotationTypes.includes(options.type)) {
+    console.warn(`No annotation found with type '${options.type}'.`);
```
kurkle, line 5: I was thinking the warning should be output when creating the annotation elements. I'll create a pr for that.

kurkle: Created #530

stockiNail: Yes, I did before that PR. I'm going to remove it

stockiNail: And also merge because a rebase is needed

stockiNail: fixed and rebased

kurkle (approved):

PR #529, opened 2021-12-04
Run event tests on plugin and annotation levels

kurkle:


test/events.js
```diff
@@ -1,156 +1,161 @@
-export function testEvents(options, innerElement) {
```
stockiNail, line 1: I'm wrong or it loos like completely change? Maybe lien separator?

stockiNail:
Thank you @kurkle 
You have added the check on the default event callbacks.

stockiNail (approved):

test/events.js
kurkle: It was ndentation level change

PR #534, opened 2021-12-06
Enable HTMLCanvasElement as label content

stockiNail:
This PR enables setting a [HTMLCanvasElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement) as label content, in box, line and label annotation options.

### TODO

- [x] test cases
- [ ] types
- [ ] documentation


kurkle (approved):

src/helpers.js
```diff
@@ -109,6 +109,10 @@ export function rotated(point, center, angle) {
   };
 }
 
+function isImage(content) {
```
LeeLenaleee, line 112: Wouldn't `isImageOrCanvas` be a more appropiate name?

stockiNail: yes, let me update it.

LeeLenaleee (approved):

src/helpers.js
stockiNail: done!

kurkle (approved):

PR #536, opened 2021-12-06
Add note about clip option usage and events catching

### LINKED ISSUE #531: Point of attention: clip: false and events catching on annotations
The PR #523 added `clip` options in order to enable the annotation drawing even if part of them are outside of chart area.

And that's good but this is disabling the events catching on the part drawn outside of the chart.
The events are managed using the `beforeEvent` hook of the plugin which is invoked for events generated inside the chart area.
In this way, if the user clicks or moves on a visible annotation out of chart area, events are not dispatched to events hooks.

If the above thought is not wrong (please correct me), there are some possible solutions that I see:

1. stay as is but maybe we should document it in order that the user can add own events listeners to the canvas, ignoring the events hooks of the plugin. 
2. remove the usage of `beforeEvent` hook, going to the canvas event listeners.
3. ask to change this behavior to Chart.js but I don't think it makes sense.

stockiNail:
Fixes #531

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -16,7 +16,7 @@ The following options are available at the top level. They apply to all annotati
 
 Setting `clip` to `false`, you can enable the possibility to draw part of the annotation outside of the chart area.
 
-Nevertheless, this disables the catching  of events on that part, outside of the chart area.
+Nevertheless events are only catched over the chartArea.
 
 :::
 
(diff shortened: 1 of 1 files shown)


docs/guide/configuration.md
```diff
@@ -12,6 +12,14 @@ The following options are available at the top level. They apply to all annotati
 | `drawTime` | `string` | Yes | `'afterDatasetsDraw'` | See [drawTime](options#draw-time)
 | `dblClickSpeed` | `number` | Yes | `350` | Time to detect a double click in ms.
 
+:::warning
+
+Setting `clip` to `false`, you can enable the possibility to draw part of the annotation outside of the chart area.
+
+Nevertheless, this disables the catching  of events on that part, outside of the chart area.
```
kurkle, line 19: I would word this somewhat differently, because the events are not disabled by the clip option.

Maybe like this:
Nevertheless events are only catched over the chartArea.

stockiNail: done

kurkle (approved):

LeeLenaleee (approved):

PR #540, opened 2021-12-08
Enable animation on text and point of label annotation

### LINKED ISSUE #539: Label annotation is not animated
The label annotation is not animated even if the element is configured to be.

This should be due to that the `labelRect` property of the element, which is containing the label text position, is calculated during the element initialization.

https://github.com/chartjs/chartjs-plugin-annotation/blob/ef7bf770652f21fd151ea3043156b4dbadd59064/src/types/label.js#L57-L62

Probably the `labelRect` should be re-calculated during the draw phase.

EDIT: To be more precise, the content of the label is not animated

stockiNail:
Fixes #539

### DIFF
--- a/src/types/label.js
+++ b/src/types/label.js
@@ -42,6 +42,10 @@ export default class LabelAnnotation extends Element {
     if (this.labelRect) {
       applyCallout(ctx, this);
       applyBox(ctx, this);
+      const {width, height, padding} = this.labelRect;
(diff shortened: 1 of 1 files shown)


kurkle:
This is the wrong direction.

`resolveElementProperties` should return the new values for the element. Those values are animated using animation configuration.

Instead it currently sets new values directly and returns only options to animate.

PR #542, opened 2021-12-08
Enforce unix linebreaks

kurkle:
Make eslint verify proper line endings (as instructed in .editorconfig)
Fix errors.


stockiNail (dismissed):

stockiNail:
@kurkle FYI I have changed my config in order to use LF instead of CRLF

stockiNail:
@kurkle I see many files with CRLF. If you agree, I can submit another PR with the update to all files where CRLF is present.

kurkle:
It looks like git refuses to accept line ending change as a change, you have to modify the content too.

stockiNail (approved):

PR #546, opened 2021-12-09
Fix random test failures

kurkle:
The cause of the random failures was caching of the measured text.
Same text is measured with and without `spriteText`, and the one run first gets cached. 

This changes the cache key by adding '-spriting' at the end when its detected.

Also the offending fixture (without spriteText enabled) that was created with wrong measure is updated.


stockiNail (approved):

stockiNail:
@kurkle awesome

PR #547, opened 2021-12-09
Make a consistent implementations of `inRange` in all annotations

### LINKED ISSUE #535: Inconsistent implementations of `inRange`

stockiNail:
Fix #535


stockiNail:
@kurkle see the polygon implementation. I'm not convinced at 100%

src/types/line.js
```diff
@@ -85,9 +85,9 @@ export default class LineAnnotation extends Element {
       y >= labelRect.y - h2 && y <= labelRect.y + h2;
   }
 
-  inRange(x, y) {
+  inRange(mouseX, mouseY, useFinalPosition) {
     const epsilon = this.options.borderWidth || 1;
-    return this.intersects(x, y, epsilon) || this.isOnLabel(x, y);
+    return this.intersects(mouseX, mouseY, epsilon, useFinalPosition) || this.isOnLabel(mouseX, mouseY);
```
kurkle, line 90: isOnLabel uses current coordinates recardless of `useFinalPosition`

src/types/polygon.js
```diff
@@ -4,8 +4,9 @@ import {setBorderStyle, resolvePointPosition, getElementCenterPoint} from '../he
 
 export default class PolygonAnnotation extends Element {
 
-  inRange(x, y) {
-    return this.vertices && this.vertices.length > 0 && pointIsInPolygon(this.vertices, x, y);
+  inRange(mouseX, mouseY, useFinalPosition) {
+    const {vertices} = this.getProps(['vertices'], useFinalPosition);
```
kurkle, line 8: This wont work. If the vertices are calculated in `draw`, we need to calculate them here, with the final x/y.

I think it could be better to calculate the points of the polygon in resolveElementProperties. Just need to make the animator animate those points. I guess it would be easiest if the points were handled as sub elements and their properties would be updated in similar manner (using updateElement)

We can just leave the Polygon annotation out of this pr, because its more complicated.

stockiNail: Let me propose to apply the double calculattion pf vertices here and then submit another PR to move the calculation in resolveElementOptions. With double calculation the polygon is animated anyway.
What do you think?

stockiNail:
@kurkle I'm working on this PR but I need time... I'm fighting with the updates on line annotations

stockiNail:
@kurkle I think I found a bug, updating the line.

https://github.com/chartjs/chartjs-plugin-annotation/blob/e78e1cfb53ffb939aba849739585d94b1f06806c/src/types/line.js#L323-L324

When the top and bottom are equals, -1 is returned. The same for left and right.

image

In my opinion, the check should be <= and not only =.

image

stockiNail:
I have fixed in this PR.

src/types/line.js
stockiNail: should be fixed

src/types/polygon.js
stockiNail: @kurkle as I wrote, I have implemented the vertices are not stored in the element anymore and in the `draw` (therefore there are animated by x,y center point and not the vertices coordinates, todo) and in `inRange`.

kurkle (approved):
I think it would be really good to have sub elements. Like point for polygon, and label. Almost same as having the labelRect, but it would be a constructed class handling its own coordinates and options. And updated in similar way the main annotation elements are updated. That would reduce complexity of the elements, enable animation with simple x,y,width,height properties etc. There are probably some cons too.

Just a brain dump here, I will try something out when I have couple of hours to spare.

stockiNail:
> I think it would be really good to have sub elements. Like point for polygon, and label. Almost same as having the labelRect, but it would be a constructed class handling its own coordinates and options. And updated in similar way the main annotation elements are updated. That would reduce complexity of the elements, enable animation with simple x,y,width,height properties etc. There are probably some cons too.
> 
> Just a brain dump here, I will try something out when I have couple of hours to spare.

Ok, now it's clear. But I think this new classes (for instance Label) should extend Element and must be registered as well, doesn't it?

For sure some cons are related that some behaviors (about the label mainly) are different between the annotations.

stockiNail:
> Just a brain dump here, I will try something out when I have couple of hours to spare.

I'm curious and I want to try as well, by myself, in a separated branch! I think it makes sense because this morning I was thinking about the huge possibilities to enrich the plugin with the animations.

src/types/line.js
```diff
@@ -65,29 +65,28 @@ export default class LineAnnotation extends Element {
...
-  labelIsVisible(chartArea) {
-    const label = this.options.label;
-
-    const inside = !chartArea || isLineInArea(this, chartArea);
-    return inside && isLabelVisible(label);
+  labelIsVisible(useFinalPosition, chartArea) {
+    if (!this.labelVisible) {
+      return false;
+    }
+    return !chartArea || isLineInArea(this.getProps(['x', 'y', 'x2', 'y2'], useFinalPosition), chartArea);
```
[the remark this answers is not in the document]
stockiNail: Understood. Let me add it to my TODO. I'll have a look.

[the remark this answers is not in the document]
stockiNail: About this code and mainly about `isLineInArea`, I think this part should be completely review due to new `clip` options.
At the moment, the line annotation can not draw anything out of the chart area, even if `clip: false`. Maybe that's not correct

kurkle:
There again I hit the Chart.js resolver bug that is fixed only in current master (will be out in Chart.js v3.7)

PR #550, opened 2021-12-09
Fix `isBoundToPoint` function implementation

### LINKED ISSUE #549: Invalid `isBoundToPoint` function implementation
The implementation of `isBoundToPoint` is wrong.

If I set `xValue: 0` and `yValue: 0`, it returns `false` and that's not correct because I want to have the annotation with center point at value 0, 0.

https://github.com/chartjs/chartjs-plugin-annotation/blob/b274a2112d20caa7532ada7bda2a94af7997987f/src/helpers/helpers.options.js#L54-L56

stockiNail:
Fix #549


src/helpers/helpers.options.js
```diff
@@ -52,5 +52,5 @@ export function toPosition(value) {
 }
 
 export function isBoundToPoint(options) {
-  return options && (options.xValue || options.yValue);
+  return options && (typeof options.xValue !== 'undefined' || typeof options.xValue !== 'undefined');
```
kurkle, line 55: Could use the `defined` helper from Chart.js to make this a bit cleaner.

stockiNail: I didn’t know it 😔

stockiNail: done

kurkle (approved):

PR #554, opened 2021-12-11
Enable label `position` as percentage of the size

### LINKED ISSUE #553: Enable label `position` as percentage of the size
Sometimes the `start`, `center` and `end` position could be the best options to position a label.

Maybe you would like to put it a specific position of the line.

To address the above cases, you could implement to set the position as percentage of the size.

`position` : 'start' | 'center' | 'end' | number | string

The number could be a value between 0 and 1 or a string with % (i.e. '25%') which represents the percentage of the size.

stockiNail:
Fix #553 

### TODO

- [x] line annotation
- [x] box annotation
- [x] label annotation
- [x] test cases
- [x] types
- [x] documentation


src/helpers/helpers.options.js
```diff
@@ -51,6 +52,19 @@ export function toPosition(value) {
   return readValueToProps(value, ['x', 'y'], 'center');
 }
 
+export function getTPosition(position, def = 0.5) {
```
kurkle, line 55: I think you can use getSize instead of this.

stockiNail: getSize is implemented thinking about the size. getTPosition should return a value between 0 and 1.

Trying to use getSize, I should pass always 1 as size otherwise the calculation by string will not be correct. But if the string is not a percentage, it should return 0.5 (center, default) and not 0 (as getSize is doing) and therefore I should add another parameter to getSize with def.

Let try to change getSize and also toPercent enabling a default.

stockiNail: Can you have a look now? Better? I left getTPosition as const invoking getSize

docs/guide/types/box.md
```diff
@@ -118,14 +118,20 @@ All of these options can be [Scriptable](../options#scriptable-options)
 | `font` | [`Font`](../options#font) | `{ weight: 'bold' }` | Label font
 | `height` | `number`\|`string` | `undefined` | Overrides the height of the image or canvas element. Could be set in pixel by a number, or in percentage of current height of image or canvas element by a string. If undefined, uses the height of the image or canvas element. It is used only when the content is an image or canvas element.
 | `padding` | [`Padding`](../options#padding) | `6` | The padding to add around the text label.
-| [`position`](#position) | `string`\|`{x: string, y: string}` | `'center'` | Anchor position of label in the box.
+| [`position`](#position) | `string`\|`{x: string, y: string}`\|`number` | `'center'` | Anchor position of label in the box.
```
kurkle, line 121: I think the type should be `string`\|`{x: string\|number, y: string\|number}`\|`number` (number is allowed for the x/y object too)

types/label.d.ts
```diff
@@ -121,7 +121,7 @@ export interface LabelTypeOptions extends ContainedLabelOptions {
   position?: Scriptable<LabelPosition | LabelPositionObject, PartialEventContext>,
 }
 
-export type LabelPosition = 'start' | 'center' | 'end';
+export type LabelPosition = 'start' | 'center' | 'end' | string | number;
```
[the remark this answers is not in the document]
stockiNail: I'm not  specialist of type script... Anyway this could be applied also of width and height options for label where the content is an image.

docs/guide/types/box.md
stockiNail: yes!!

stockiNail:
I have tested your suggestions but it seems something is not correct:

bugPosition

See y>1 and y<0

src/types/label.js
```diff
@@ -149,7 +149,7 @@ function calculatePosition(start, size, adjust, position) {
   } else if (position === 'end') {
     return start - size + adjust;
   }
-  return start - size / 2 + adjust;
+  return start + getSize(size, position, size / 2) + adjust ;
```
kurkle, line 152: I made a sign error?
```suggestion
  return start - getSize(size, position, size / 2) + adjust ;
```

stockiNail:
Ok, now I undestand better your commits and there is a misunderstanding.

The position as number is not in PIXEL but is a value between 0 and 1 (as percentage) but numeric. I don't know if makes sense to allow the number as position in PX because there is already x/yAdjust options for that.

stockiNail:
If we don't to provide the number as percentage, it's better to remove it and enable only the percentage as string.
What do you think?

test/fixtures/box/labelPosition.js
```diff
@@ -0,0 +1,108 @@
...
+              xMin: -9,
+              yMin: -1,
+              xMax: -1,
+              yMax: -9,
+              backgroundColor: 'white',
+              borderColor: 'red',
+              borderWidth: 2,
+              label: {
+                enabled: true,
+                content: 'p: 0.50-0.75',
```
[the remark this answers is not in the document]
stockiNail: ok!

src/types/line.js
```diff
@@ -307,12 +307,14 @@ function rotatedSize(width, height, rotation) {
...
 function calculateT(line, label, sizes, chartArea) {
-  let t = 0.5;
+  let t;
   const space = spaceAround(line, chartArea);
   if (label.position === 'start') {
     t = calculateTAdjust({w: line.x2 - line.x, h: line.y2 - line.y}, sizes, label, space);
   } else if (label.position === 'end') {
     t = 1 - calculateTAdjust({w: line.x - line.x2, h: line.y - line.y2}, sizes, label, space);
+  } else {
+    t = getTPosition(label.position, 0.5);
```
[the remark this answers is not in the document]
stockiNail: I agree that maybe it's not worth it. Nevertheless we should solve the misunderstanding about position by number meaning. ;)

kurkle:
> Ok, now I understand better your commits and there is a misunderstanding.
>
> The position as number is not in PIXEL but is a value between 0 and 1 (as percentage) but numeric. I don't know if makes sense to allow the number as position in PX because there is already x/yAdjust options for that.

> If we don't to provide the number as percentage, it's better to remove it and enable only the percentage as string. What do you think?

I don't think having 2 ways to do the same thing (define percentage) is good. It creates many code paths for the same result and could be confusing to users.

So, I agree, it would be better to remove and only enable the percentage string.

I think a good set of helpers could be (would also make the `width: 'invalid string'` option for images fall back to the actual width):

```js
const isPercentString = (s) => typeof s === 'string' && s.endsWith('%');
const toPercent = (s) => parseFloat(s) / 100;

export function getRelativePosition(size, positionOption) {
  if (positionOption === 'start') {
    return 0;
  }
  if (positionOption === 'end') {
    return size;
  }
  if (isPercentString(positionOption)) {
    return toPercent(positionOption) * size;
  }
  return size / 2;
}

export function getSize(size, value) {
  if (typeof value === 'number') {
    return value;
  } else if (isPercentString(value)) {
    return toPercent(value) * size;
  }
  return size;
}
```

Again my thoughts only, I can agree to other implementations too.

docs/guide/types/label.md
```diff
@@ -71,7 +71,7 @@ The following options are available for label annotations.
 | [`height`](#general) | `number`\|`string` | Yes | `undefined`
 | [`padding`](#general) | [`Padding`](../options#padding) | Yes | `6`
 | [`point`](#point) | `object` | Yes |
-| [`position`](#position) | `string`\|`{x: string, y: string}` | Yes | `'center'`
+| [`position`](#position) | `string`\|`{x: string, y: string}`\|`number` | Yes | `'center'`
```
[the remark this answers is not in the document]
stockiNail: done

docs/guide/types/box.md
stockiNail: done

stockiNail:
OK, fine for me! Let me change the code and test it!

stockiNail:
the `getRelativePosition` is working well only with position 'start'. I need to refactor it.

kurkle:
> the `getRelativePosition` is working well only with position 'start'. I need to refactor it.

It's assuming you remove paddings and margins etc extra adjustments and pass the actual available size.

stockiNail:
> It's assuming you remove paddings and margins etc extra adjustments and pass the actual available size.

yes, and forgive my lack of knowledge. But for instance, setting 'end', this is returning the size but I can not add to start point but I need to calculate the space anyway. And to do that, I need to re-check the position value (`if` statement). The same for center where I have to take care about the label size.

A label is drawn passing the box where the label must be drawn (see also `textAlign`).

For what I can see now (but open to change my idea) the best is to have something like was implemented before....

But again, feel free to correct me because I can not see another good way to do it.

stockiNail:
In my poor opinion, trying to "centralize" the logic of the position, we could risk to increase complexity.

kurkle:
See https://github.com/pepstock-org/chartjs-plugin-annotation/pull/1

stockiNail:
> See [pepstock-org#1](https://github.com/pepstock-org/chartjs-plugin-annotation/pull/1)

@kurkle really thank you for support! I really appreciate. 

They are code climate similar code... But as we did before, we could accept it.

kurkle:
Yes, we can ignore the codeclimate here.

But there are still things to do.. The docs need to be updated, as it was agreed not to accept the `number` variant of the percantege. Also the getTPoistion function is unused I think.

stockiNail:
yes, I am also rebasing to have the another PR.. need some minutes (let me finish another stuff, unfortunately not related to this).

stockiNail:
> Yes, we can ignore the codeclimate here.

How can I set it to be ignored? :(

kurkle:
> > Yes, we can ignore the codeclimate here.
> 
> How can I set it to be ignored? :(

Don't look at it :)

test/fixtures/box/labelPosition.js
```diff
@@ -0,0 +1,111 @@
...
+              xMin: 1,
+              yMin: -1,
+              xMax: 9,
+              yMax: -9,
+              backgroundColor: 'white',
+              borderColor: 'red',
+              borderWidth: 2,
+              label: {
+                enabled: true,
+                content: 'p: 100%,25%',
```
kurkle, line 96: My mistake here
```suggestion
                content: 'p: 100%,0%',
```

kurkle (commented):
Only one mistake left, made by me. I'll commit the fix directly.

The vertical label percentage fixture is quite a lot better that the horizontal one. Horizontal is missing the edge values (0%, 100%), that would be good to have in a test.
Anyway, I think it is good enough for its purpose.

kurkle:
> Only one mistake left, made by me. I'll commit the fix directly.

Hmm, I can't. Maybe the "allow cnages by maintainers" flag was not checked in the PR.

Also it needs a update to the fixture, so can't do that directly here too :)

stockiNail:
> > Only one mistake left, made by me. I'll commit the fix directly.
> 
> Hmm, I can't. Maybe the "allow cnages by maintainers" flag was not checked in the PR.
> 
> Also it needs a update to the fixture, so can't do that directly here too :)

I can do it.

PR #555, opened 2021-12-11
Enable sub elements

kurkle:
sub-elems

* requires Chart.js v3.7


stockiNail:
@kurkle what can I say? Cool!
I’m having a look and sounds really good!
Let me know if I can help

kurkle:
Not sure if this should be in v2 or if it would be ok to require a certain minor version of chart.js in an minor update.

kurkle:
| `dist/chartjs-plugin-annotation.esm.js` | 12.3 kB | +583 B (+5%) | 🔍 |
| `dist/chartjs-plugin-annotation.js` | 12.4 kB | +577 B (+5%) | 🔍 |
| `dist/chartjs-plugin-annotation.min.js` | 8.18 kB | +227 B (+3%) |  |

kurkle:
Makes sense to clip the sub-elements:

image

to

image

kurkle:
> Not sure if this should be in v2 or if it would be ok to require a certain minor version of chart.js in an minor update.

@etimberg this is the question I requested your review for, but please do a full review too :)

src/types/box.js
```diff
@@ -18,12 +18,11 @@ export default class BoxAnnotation extends Element {
...
   drawLabel(ctx) {
-    const {x, y, width, height, options} = this;
-    const labelOpts = options.label;
-    labelOpts.borderWidth = options.borderWidth;
+    const {x, y, width, height, options: {label: labelOpts}} = this;
+    const halfBorder = labelOpts.borderWidth / 2;
     ctx.save();
     ctx.beginPath();
-    ctx.rect(x + labelOpts.borderWidth / 2, y + labelOpts.borderWidth / 2, width - labelOpts.borderWidth, height - labelOpts.borderWidth);
+    ctx.rect(x + halfBorder, y + halfBorder, width - halfBorder, height - halfBorder);
```
etimberg, line 25: Was the change to use `width - halfBorder` intended?

src/helpers/helpers.core.js
```diff
@@ -25,3 +25,19 @@ export function getElementCenterPoint(element, useFinalPosition) {
   const {x, y} = element.getProps(['x', 'y'], useFinalPosition);
   return {x, y};
 }
+
+const isOlderPart = (act, req) => req > act || (act.length > req.length && act.substr(0, req.length) === req);
```
etimberg, line 29: What about storing the `parseInt` versions from line 36 into variables and then passing numbers to this function. Would be simpler logic I think

etimberg (commented):
Depending on >= 3.7 is fine by me. Is there any way to support other recent versions like 3..5 and 3.6?

src/types/box.js
kurkle: good catch

The version requirement is due to https://github.com/chartjs/Chart.js/pull/9969 and is only required when annotations are passed as an array. There are already similar issues in current version and arrays with chart.js < 3.7.

So for proper functionality 3.7 should be used with any version of this plugin, when using annotations as array.

kurkle: Fixed, the clip did not account for padding either, which I think it should (applied).

src/helpers/helpers.core.js
kurkle: I think it would not work if requiring 4.0.0 and the version was 4.0.0-alpha

etimberg: Good point

etimberg (approved):

etimberg (dismissed):

src/types/box.js
```diff
@@ -58,6 +60,7 @@ BoxAnnotation.defaults = {
   cornerRadius: undefined, // TODO: v2 remove support for cornerRadius
   display: true,
   label: {
+    borderWidth: undefined,
```
stockiNail, line 63: I think this new options should be documented. FYI I didn't add borderWidth because I thought that the label in box couldn't have any border (and any background color).
And, maybe I'm wrong, it's never set to the context, therefore is not used to draw the box of the label.

kurkle: It's actually a leftover abd should be removed. I used this instead of setting the labelOpts.borderWidth = opts.borderWidth, but found out later it was actually the box border width that is used.

stockiNail: :+1: do you want to draw a rectangle around the label anyway?

kurkle: No, I don't have a use case for that right now, but no objection either. I'd ideally like the label to be a sub element, but the label annotation is too fancy for that. Maybe if the callout was a sub element too.. Needs more thought anyway.

stockiNail: I was reviewing this PR and and the usage of sub elements for inner labels (box and line). But you're right, also callout could be.

src/helpers/helpers.core.js
```diff
@@ -27,3 +27,19 @@ export function getElementCenterPoint(element, useFinalPosition) {
...
+export function requireVersion(pkg, min, ver) {
+  const parts = ver.split('.');
+  let i = 0;
+  for (const req of min.split('.')) {
+    const act = parts[i++];
+    if (parseInt(req, 10) < parseInt(act, 10)) {
+      break;
+    }
+    if (isOlderPart(act, req)) {
+      throw new Error(`${pkg} v${ver} is not supported. v${min} or newer is required.`);
```
stockiNail, line 42: @kurkle maybe I'm wrong but this looks like a breaking change.
We are forcing all users to use CHART.JS 3.7.0. Maybe a warning could be emitted and in the version 2.0.0 of the plugin could be an error.
Furthermore, afa I understood, this strong dependency is related to the use of sub elements, currently used only in polygons and then, if a user is not using polygons, is anyway forced to change.

stockiNail: @kurkle if we want to maintain it, we should also add a note in the doc and change the README.

kurkle: You are probably right, we could do with a warning. The requirement is related to fallback from array options. So applies to use cases that proved annotations as array too. Those "work", but defaults come from line annotation and not from the actual type.

The workaround this PR removed from `afterRegister` needs to be restored and verify it contains sub-objects of every annotation types defaults.

PR #558, opened 2021-12-12
Remove `point` node from label annotation

### LINKED ISSUE #556: Remove `point` node from label annotation
Currently label annotation is taking care also of the point drawing, the point selected with coordinates calculated by x/yVaue.

To simplify the label annotation, reducing duplication of similar code and options, the point node could be removed.
If a user wants to draw the selected point, could add another annotation, a point one, with the same x/yVaue.

This is also enabling a common approach if a user wants to draw a polygon instead of a point.

As is now:

```javascript
const labelAndPoint = {
  type: 'label',
  backgroundColor: 'rgba(245,245,245, 0.5)',
  content: (ctx) => 'Maximum value is ' + maxValue(ctx).toFixed(2),
  point: {
    enabled: true,
    backgroundColor: 'transparent',
    borderColor: (ctx) => ctx.chart.data.datasets[0].borderColor,
    pointStyle: 'rectRounded',
    radius: 10
  },
  xValue: 0,
  yValue: 0
};
```

As to be:

```javascript
const label = {
  type: 'label',
  backgroundColor: 'rgba(245,245,245, 0.5)',
  content: (ctx) => 'Maximum value is ' + maxValue(ctx).toFixed(2),
  xValue: 0,
  yValue: 0
};
const point = {
  type: 'point',
  backgroundColor: 'transparent',
  borderColor: (ctx) => ctx.chart.data.datasets[0].borderColor,
  pointStyle: 'rectRounded',
  radius: 10,
  xValue: 0,
  yValue: 0
};
```

This shouldn't be a breaking change because label annotation is not releases yet in any version.

@kurkle what do you think?

stockiNail:
Fix #556


kurkle (approved):
| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/chartjs-plugin-annotation.esm.js` | 11.4 kB | -106 B (-1%) |
| `dist/chartjs-plugin-annotation.js` | 11.5 kB | -108 B (-1%) |
| `dist/chartjs-plugin-annotation.min.js` | 7.78 kB | -86 B (-1%) |

I noticed you used the point extensively in #553, are you sure? :)

stockiNail:
> I noticed you used the point extensively in #553, are you sure? :)

Well, I thought that the point implementation in the label was done because I'm personally still convinced that the point could be important to be highlighted but, playing with samples, I discovered can a user could want to highlight the point with different shapes or with another annotation, therefore I think to have a point as possible options has got more cons than pro (mainly duplication code and config).

I'm learning step by step... Sorry.

stockiNail:
I wait to merge for the decision about the other PR about position options, about percentage about number, in order to rebase the other PR going to the shared decision.

PR #560, opened 2021-12-14
Fix `animations` option type and add missing documentation

### LINKED ISSUE #559: animations is not optional in types and is missing from documentation
Since this commit, I have the following error in my project (when I try to upgrade this library):
> error TS2741: Property 'animations' is missing in type '{ annotations: { [key: string]: AnnotationOptions<keyof AnnotationTypeRegistry>; }; }' but required in type 'AnnotationPluginOptions'.

Shouldn't the `animations` property be optional ?
```diff
- animations: Record<string, unknown>,
+ animations?: Record<string, unknown>,
```

I don't find any documentation on this property in https://www.chartjs.org/chartjs-plugin-annotation/latest/ ; am I missing something ?

Thanks!

_Originally posted by @jbelien in https://github.com/chartjs/chartjs-plugin-annotation/pull/448#discussion_r768806334_

stockiNail:
Fixes #559

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -11,6 +11,7 @@ The following options are available at the top level. They apply to all annotati
 | `clip` | `boolean` | No | `true` | Are the annotations clipped to the chartArea.
 | `drawTime` | `string` | Yes | `'afterDatasetsDraw'` | See [drawTime](options#draw-time)
 | `dblClickSpeed` | `number` | Yes | `350` | Time to detect a double click in ms.
+| [`animations`](#animations) | `object` | No | [see here](#default-animations) | To configure which element properties are animated and how
 
 :::warning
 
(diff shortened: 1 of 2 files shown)


kurkle (approved):

kurkle:
Should pay some more attention to the commit messages :)

Screenshot_20211214-190333

stockiNail:
@kurkle I didn't understand...

kurkle:
It's the first commit message that gets suggested when merging. You can edit it at that point too.

stockiNail:
> You can edit it at that point too.

Ok.. apologize.. but is it in the form opened when I'm merging? Because I can't see where what you reported is located.

kurkle:
https://github.com/chartjs/chartjs-plugin-annotation/commits/master

Commit messages create the git history. Pull request description ends up in the release notes and is more important. But good commit messages are important too.

More information for example here: https://www.google.com/amp/s/cbea.ms/git-commit/amp/

stockiNail:
Now it's clear. My fault is to commit the branch with the "fix [issue]" instead of the real title of the target PR in the project... Sorry.
Be patient

kurkle:
Also don't worry too much about that, I'm not very good with the commit messages either.
Just pointing out its something that we should try to improve.

stockiNail:
> Also don't worry too much about that, I'm not very good with the commit messages either. Just pointing out its something that we should try to improve.

Absolutely I'll do!

PR #567, opened 2021-12-15
Publish latest/next docs in separate job

kurkle:
Previous attempt was flawed, this could actually work.

### DIFF
--- a/.github/workflows/npmpublish.yml
+++ b/.github/workflows/npmpublish.yml
@@ -76,7 +76,7 @@ jobs:
           destination_dir: ${{ needs.setup.outputs.version }}
 
   publish-docs:
-    needs: [test]
+    needs: [test, setup]
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


stockiNail (dismissed):

.github/workflows/npmpublish.yml
```diff
@@ -47,50 +48,65 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 16
+          registry-url: https://registry.npmjs.org/
+
+      - name: Build docs (latest/next)
         if: "!github.event.release.prerelease"
```
LeeLenaleee, line 89: Won't it skip the build now for a next branch since they are prereleases right?

LeeLenaleee (approved):

PR #568, opened 2021-12-16
Remove point nodes from label annotations in the test cases

stockiNail:
Removes point nodes from label annotations in the test cases


kurkle (approved):

kurkle (approved):
This is ok, but the milestone should be removed before merging, because we need to release 1.2.1 to fix the types issue.

This is fine to merge directly though, I just wanted to add this comment before you do :)

stockiNail:
> This is ok, but the milestone should be removed before merging, because we need to release 1.2.1 to fix the types issue.
> 
> This is fine to merge directly though, I just wanted to add this comment before you do :)

Perfect!

PR #569, opened 2021-12-16
Add integration tests (ts)

kurkle:
Add integration tests against a set of typescript versions.
3.7 still worked, 3.6 failed on chart.js:

>Testing on typescript-3.6 ...
>node_modules/chart.js/types/index.esm.d.ts:461:5 - error TS2315: Type 'DeepPartial' is not generic.

This is adapted from https://github.com/graphql/graphql-js

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
     - name: Use Node.js
       uses: actions/setup-node@v2
       with:
+        cache: npm
         node-version: 16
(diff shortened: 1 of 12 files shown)


kurkle:
The timeout could still be a bit small with heavier load and windows. But lets see.

[test-integration]     ✔ chartjs-plugin-annotation should compile with all supported TS versions (76749ms)

etimberg (approved):

PR #571, opened 2021-12-17
Store drawable elements to the state

stockiNail:
Calculating the drawable elements once per chart update and storing them in the state, this PR is improving the performance of annotations drawing.

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -23,7 +23,7 @@ export default {
     chartStates.set(chart, {
       annotations: [],
       elements: [],
-      drawableElements: [],
+      visibleElements: [],
       listeners: {},
       listened: false,
       moveListened: false
@@ -58,6 +58,7 @@ export default {
     const state = chartStates.get(chart);
     updateListeners(chart, state, options);
     updateElements(chart, state, options, args.mode);
+    state.visibleElements = state.elements.filter(el => !el.skip && el.options.display);
   },
 
   beforeDatasetsDraw(chart, _args, options) {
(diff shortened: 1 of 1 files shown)


src/annotation.js
```diff
@@ -23,6 +23,7 @@ export default {
     chartStates.set(chart, {
       annotations: [],
       elements: [],
+      drawableElements: [],
```
kurkle, line 26: I would have called it `visibleElements`, but I think `drawableElements` is actually better name.

kurkle (commented):
The CC issues could be addressed by creating the array in `afterUpdate`:
```js
state.drawableElements = state.elements.filter(el => !el.skip && el.options.display);
```

That would be better for maintenance / amount of code, but could be worse for performance. In this case I'd take the more performant option.

Performance is quite tricky to measure reliably too.

kurkle:
@stockiNail note about the versioning; 1.2.1 will be a patch release, only fixing bug(s) in 1.2.0. So no other improvements to that release.
I'm sure you know this, but as this is the 2nd pr today you place to 1.2.1 milestone that does not really belong there, I'll explain just in case :)

So the version numbers are major.minor.patch
- Matjor releases can include breaking changes + anything else
- Minor releases non-breaking enhancements + anything else
- Patch releases are bug fixes to that minor version.

stockiNail:
@kurkle thank you ! Another thing to learn. I'm working differently in the other project, but no problem! And apologize!

src/annotation.js
stockiNail: done!

kurkle:
> @kurkle thank you ! Another thing to learn. I'm working differently in the other project, but no problem! And apologize!

You are probably not distributing through npm: https://docs.npmjs.com/about-semantic-versioning

etimberg:
The main repo got lots of negative feedback when we accidentally broke things in minor/patch releases. We had to learn the hard way 😂

stockiNail:
@kurkle Yes, I'm not distributing by NPM because my project is not JS one.

kurkle (approved):

PR #574, opened 2021-12-17
Move scale management functions to a dedicated file

stockiNail:
This PR creates the scale module where there are some scale functions invoked from `annotation.js`.
These functions are NOT moved in a dedicated `helpers` or  in `helpers.chart` because they are invoked only from 1 point.

This is done to reduce the size of `annotation.js` and improving readability, concentrating all functions of the same topic in a dedicated module.

### DIFF
--- a/src/scale.js
+++ b/src/scale.js
@@ -2,7 +2,19 @@ import {isFinite, valueOrDefault} from 'chart.js/helpers';
 
 export function adjustScaleRange(chart, scale, annotations) {
   const range = getScaleLimits(scale, annotations);
-  let changed = changeScaleLimit(scale, range, 'min', 'suggestedMin') || changeScaleLimit(scale, range, 'max', 'suggestedMax');
+  let changed = false;
+  if (isFinite(range.min) &&
+		typeof scale.options.min === 'undefined' &&
+		typeof scale.options.suggestedMin === 'undefined') {
+    changed = scale.min !== range.min;
+    scale.min = range.min;
+  }
+  if (isFinite(range.max) &&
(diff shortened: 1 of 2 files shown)


kurkle (dismissed):
Those CC issues can and probably should be ignored here, because its hard to follow the history in GH when things move around and change in the same time. 

So IMO moving a function to new location should not change the function itself.

stockiNail:
> Those CC issues can and probably should be ignored here, because its hard to follow the history in GH when things move around and change in the same time.
> 
> So IMO moving a function to new location should not change the function itself.

@kurkle ok, anyway I have already done. Let me push it, if you don't like it, I'll revert this update.

src/scale.js
```diff
@@ -0,0 +1,63 @@
+import {isFinite, valueOrDefault} from 'chart.js/helpers';
+
+export function adjustScaleRange(chart, scale, annotations) {
+  const range = getScaleLimits(scale, annotations);
+  let changed = changeScaleLimit(scale, range, 'min', 'suggestedMin') || changeScaleLimit(scale, range, 'max', 'suggestedMax');
```
kurkle, line 5: The later function is not called when first one returns true. So this does not work properly.

I guess we don't have any tests that would catch this.

kurkle (commented):
Contains a bug, but looks quite good.
I'd still like to have these changes in separate commit for the reasons stated previously.

src/scale.js
stockiNail: move to another PR

kurkle (approved):

PR #575, opened 2021-12-17
Workaround for array syntax bug with Chart.js<v3.7

kurkle:


stockiNail (dismissed):

stockiNail:
@kurkle just for my info. PR 555 requires Chart.js v3.7 will require Chartjs 3.7. If and when the PR will be merged, should this workaround removed because there will be a string dependency with 3.7? If yes, maybe makes sense to add a TODO.

kurkle:
> @kurkle just for my info. PR 555 will require Chartjs 3.7. If and when the PR will be merged, should this workaround be removed because there will be a strong dependency with 3.7? If yes, maybe makes sense to add a TODO.

Yes that could make sense. I'm not sure at which version of the plugin should we require Chart.js v3.7, but the comment does not need to include a plugin version.

stockiNail (approved):

etimberg (approved):

PR #579, opened 2021-12-18
Add `beforeDraw` and `afterDraw` hooks to the annotations

stockiNail:
This PR is adding before/after draw hooks in order to allow the user to apply whatever needed customization on out-of the-box annotations.

```javascript
plugins: {
  annotation: {
    annotations: {
      line: {
        type: 'line',
        ....
        beforeDraw(context) {
          console.log('beforeDraw');
          return true;
        },
        afterDraw(context) {
          console.log('afterDraw');
        },
      },
    }
  }
}
```

| Hook | Arguments | Return type | Default | Description
| :-- | :-- | :-- | :-- | :--
| beforeDraw | context | boolean | `annotation.beforeDraw` | invoked before the element drawing. If returns `false`, the element is not drawn.
| afterDraw | context | void | `annotation.afterDraw` | invoked after the element drawing

Common hooks for all annotations can be set in the annotation plugin options:

```javascript
plugins: {
  annotation: {
    beforeDraw(context) {
      console.log('Common beforeDraw');
      return true;
    },
    afterDraw(context) {
      console.log('Common afterDraw');
    },
    annotations: {
      line: {
        type: 'line',
        ....
      },
    }
  }
}
```

### TODO

- [x] test cases
- [x] types
- [x] documentation
- [x] samples


stockiNail:
Due to the fact that element and label couldn't be drawn in sequence but, in accordance with `drawTime` options, they could be drawn in a difference sequence, some element internal properties, to manage status of drawing,  has been added.

kurkle:
> Due to the fact that element and label couldn't be drawn in sequence but, in accordance with `drawTime` options, they could be drawn in a difference sequence, some element internal properties, to manage status of drawing, has been added.

I asked for only these two hooks first, so I'd see a small PR that is easily reviewed.
I think you are trying to implement all the use cases you have with these 2 hooks now?

stockiNail:
> I asked for only these two hooks first, so I'd see a small PR that is easily reviewed.
> I think you are trying to implement all the use cases you have with these 2 hooks now?

Yes, @kurkle. As shared, I started with before/afterDraw which are including both element and label drawing (as we shared in the other PR).
I'm trying to see which use cases I can implement with these 2. For instance, issues 145 and 146 can be addressed by these 2 hooks.
Anyway with these hooks you can only extend the existing ones, not create new annotations.
To do that, you need more hooks:

1. before/afterUpdate, to be invoked before and after resolveElementOptions, in order to allow the user to set own dimensions.
2. inRange, in order to catch events.
3. centerPoint, maybe not needed if we go to normalized size (issue 532)

docs/guide/configuration.md
```diff
@@ -67,3 +67,14 @@ The following options are available for all annotation types. These options can
 | `leave` | `(context, event) => void` | No | Called when the mouse leaves the annotation.
 | `click` | `(context, event) => void` | No | Called when a single click occurs on the annotation.
 | `dblClick` | `(context, event) => void` | No | Called when a double click occurs on the annotation.
+
+## Hooks
+
+The following options are available for all annotation types. These options can be specified per annotation, or at the top level which apply to all annotations.
+
+These options enable some user customizations on the annotations.
```
LeeLenaleee, line 75: Would replace the word `options` with `hooks`, sounds more logical to me

docs/samples/box/shadow.md
```diff
@@ -0,0 +1,169 @@
...
+  }, {
+    data: Utils.points({count: DATA_COUNT, min: MIN[1], max: MAX[1]})
+  }]
+};
+// </block:setup>
+
+// <block:annotation1:1>
+const annotation1 = {
+  type: 'box',
+  backgroundColor: 'rgba(0,150,0,0.02)',
```
LeeLenaleee, line 23: Spaces after comma

docs/samples/box/shadow.md
```diff
@@ -0,0 +1,169 @@
...
+  yMax: (ctx) => max(ctx, 0, 'y') + 2,
+  yMin: (ctx) => min(ctx, 0, 'y') - 2,
+  beforeDraw: (ctx) => drawShadow(ctx)
+};
+// </block:annotation1>
+
+// <block:annotation2:2>
+const annotation2 = {
+  type: 'box',
+  backgroundColor: 'rgba(150,0,0,0.02)',
```
LeeLenaleee, line 43: Same as above

docs/samples/line/hook.md
```diff
@@ -0,0 +1,123 @@
...
+  ctx.lineDashOffset = options.borderDashOffset;
+  // Draw
+  ctx.beginPath();
+  ctx.moveTo(x - 40, y);
+  ctx.lineTo(x, y);
+  ctx.stroke();
+  ctx.beginPath();
+  ctx.moveTo(x2, y2);
+  ctx.lineTo(x2 + 40, y);
+  ctx.stroke();
```
LeeLenaleee, line 54: Should also work without the double stroke an begin path
```js
  ctx.beginPath();
  ctx.moveTo(x - 40, y);
  ctx.lineTo(x, y);
  ctx.moveTo(x2, y2);
  ctx.lineTo(x2 + 40, y);
  ctx.stroke();
```

docs/samples/box/shadow.md
```diff
@@ -0,0 +1,169 @@
...
+  ctx.shadowBlur = 9;
+  ctx.shadowOffsetX = 5;
+  ctx.shadowOffsetY = -5;
+  ctx.stroke();
+  ctx.restore();
+  return true;
+}
+// </block:utils>
+
+var actions = [
```
LeeLenaleee, line 131: Using const for other vars, so would be better to use it here too

docs/samples/line/hook.md
```diff
@@ -0,0 +1,123 @@
...
+        annotations: {
+          annotation
+        }
+      }
+    }
+  }
+};
+/* </block:config> */
+
+var actions = [
```
LeeLenaleee, line 87: Same as in previous actions block, const instead of var

src/events.js
```diff
@@ -2,14 +2,14 @@ import {distanceBetweenPoints, callback as callHandler} from 'chart.js/helpers';
 
 const clickHooks = ['click', 'dblclick'];
 const moveHooks = ['enter', 'leave'];
-export const hooks = clickHooks.concat(moveHooks);
+export const eventsHooks = clickHooks.concat(moveHooks);
```
LeeLenaleee, line 5: eventHooks sounds more logical since you have a double multiplication now

src/hooks.js
```diff
@@ -0,0 +1,75 @@
...
+  elementsHooks.forEach(hook => {
+    if (typeof options[hook] === 'function') {
+      state.hooked = true;
+      state.hooks[hook] = options[hook];
+    }
+  });
+
+  if (!state.hooked) {
+    visibleElements.forEach(scope => {
+      if (!state.hooked) {
```
LeeLenaleee, line 19: Redundent if statement, same as line 17 so state.hooked is already false

src/hooks.js
```diff
@@ -0,0 +1,75 @@
+import {callback} from 'chart.js/helpers';
+
+export const elementsHooks = ['afterDraw', 'beforeDraw'];
```
LeeLenaleee, line 3: same as eventHooks

LeeLenaleee (commented):
Mostly small stuff that I think the wording should be slightly different for easyer reading

src/hooks.js
LeeLenaleee: Mb misread it

docs/samples/box/shadow.md
stockiNail: @LeeLenaleee do you mean inside of color definitions?

docs/samples/box/shadow.md
stockiNail: @LeeLenaleee do you mean inside of color definitions?

docs/samples/box/shadow.md
stockiNail: @LeeLenaleee agree! I see that var is iused in many other samples. I think they are coming from previous version.
I fix this new ones and I'm going to submit a PR for the others

docs/samples/line/hook.md
stockiNail: @LeeLenaleee agree! I see that var is iused in many other samples. I think they are coming from previous version.
I fix this new ones and I'm going to submit a PR for the others

stockiNail:
@LeeLenaleee thank you very much! I'm leaving your 2 notes (about comma) open because not clear to me where to add spaces.

docs/samples/box/shadow.md
LeeLenaleee: yes so instead of `backgroundColor: 'rgba(0,150,0,0.02)'` you get `backgroundColor: 'rgba(0, 150, 0, 0.02)',`. 

Note the space in front of the `150`, `0` and `0.02`

stockiNail: OK! as for `var`, this should be applied to all samples in order to keep them aligned. I'm going to submit a PR for that as well

stockiNail: done!

docs/samples/box/shadow.md
stockiNail: done|

src/hooks.js
```diff
@@ -0,0 +1,77 @@
...
+  const method = element[options.method];
+  method.call(element, chart.ctx, chart.chartArea);
+  element[options.status] = true;
+  afterDraw(state, element);
+}
+
+function beforeDraw(state, element) {
+  if (element._draw === false) {
+    return false;
+  }
```
kurkle, line 53: why is this needed?

test/hooks.js
```diff
@@ -0,0 +1,63 @@
...
+      }
+    },
+  };
+
+  describe('hooks', function() {
+    const pluginOpts = chartConfig.options.plugins.annotation;
+
+    [pluginOpts, options].forEach(function(targetOptions) {
+
+      it(`should detect before and after draw on ${options.type}`, function(done) {
```
kurkle, line 31: should call

kurkle (commented):
only partial review, will continue tomorrow.

docs/samples/box/shadow.md
```diff
@@ -0,0 +1,169 @@
...
+  ctx.save();
+  ctx.fillStyle = 'transparent';
+  ctx.strokeStyle = options.borderColor;
+  ctx.beginPath();
+  addRoundedRectPath(ctx, {
+    x, y, w: width, h: height,
+    radius: options.borderRadius
+  });
+  ctx.closePath();
+  ctx.fill();
```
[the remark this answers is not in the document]
stockiNail: I used transparent in order to show the shadow internally to the box

docs/samples/box/shadow.md
```diff
@@ -0,0 +1,169 @@
...
+  const dataset = ctx.chart.data.datasets[datasetIndex];
+  return dataset.data.reduce((v, point) => Math.min(point[prop], v), Infinity);
+}
+
+function max(ctx, datasetIndex, prop) {
+  const dataset = ctx.chart.data.datasets[datasetIndex];
+  return dataset.data.reduce((v, point) => Math.max(point[prop], v), -Infinity);
+}
+
+function addRoundedRectPath(ctx, rect) {
```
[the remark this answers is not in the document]
stockiNail: Yes, but I think I should do in the utils and wrap it, doesn't it?

src/hooks.js
stockiNail: The _draw is needed because element and label are drawn in 2 different time frame. Therefore if the hook returns false (for instance before drawing the element), the label shouldn't be drawn, when invoked.

docs/samples/box/shadow.md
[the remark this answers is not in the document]
stockiNail: added to Utils

test/hooks.js
stockiNail: done

docs/samples/line/hook.md
```diff
@@ -0,0 +1,121 @@
...
+// <block:utils:3>
+function drawExtraLine(context) {
+  const ctx = context.chart.ctx;
+  const {x, y, x2, y2, options} = context.element;
+  ctx.save();
+  ctx.lineWidth = options.borderWidth;
+  ctx.strokeStyle = options.borderColor;
+  ctx.setLineDash([6, 6]);
+  ctx.lineDashOffset = options.borderDashOffset;
+  // Draw
```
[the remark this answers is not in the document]
stockiNail: done

docs/samples/box/shadow.md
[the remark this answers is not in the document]
kurkle: How is it different compared to no fill?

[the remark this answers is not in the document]
stockiNail: you're right. It looks like better with background:

withBack

stockiNail:
@kurkle @LeeLenaleee I was thinking about this PR. I think that probably the complexity we are adding here is too much comparing with the benefits.

At the moment, there are 2 issues which can be addressed by this PR but one of them can be also solved implementing shadowing on the annotations and the other one can't be solved as out of scope of the plugin (at least at the moment).

We could freeze this PR and spend more time to think about. What do you think?

LeeLenaleee:
I don't mind setting this on a hold to see if it might be implemented in a different less complex way.
But I do still think it is a verry good addition to the plugin for users to apply a lot of customizability in a relative easy way.

stockiNail:
> I don't mind setting this on a hold to see if it might be implemented in a different less complex way. But I do still think it is a verry good addition to the plugin for users to apply a lot of customizability in a relative easy way.

I agree with you about the capabilities but I wouldn't add complexity if not really needed. In my opinion now it's not so complex but probably the current features are not enough and some other hooks will be needed. Anyway I'm continue thinking to as simple as possible implementation

kurkle:
Yeah, the "draw state" is a thing I'd like to avoid. Its the only thing I dislike here actually. But lets give it some more time.

kurkle:
Looks like we could actually release 1.2.2 instead of 1.3.0 without this/other unmerged enhancements.

stockiNail:
> Looks like we could actually release 1.2.2 instead of 1.3.0 without this/other unmerged enhancements.

It makes sense to me!

stockiNail:
> Looks like we could actually release 1.2.2 instead of 1.3.0 without this/other unmerged enhancements.

Do we have to change the milestone label to PR?

stockiNail:
I know I said to take time to think about but I had item and I remove `element._draw` status. 

Now I'm using 2 flags (`_drawElement` and `_drawLabel`) at element level to be aware if the element and the label has been drawn.

src/hooks.js
stockiNail: the `_draw` property has been removed

stockiNail:
I need to close this PR because I need to re-fork the project.
As soon as I'll have new project, I'll submit new PR with the same content.

PR #582, opened 2021-12-19
Remove undefined event listeners from state if disabled at runtime

### LINKED ISSUE #581: State doesn't update the listeners which has been removed at runtime
When you define some event hooks at top level of plugin options, if during the lifecycle you remove it, the state doesn't change the list of listeners and therefore if the event is fired against that hook, the hook is invoked even if the user disabled it.

This happens when you have more than 1 hooks and, after removing some of them, at least one remains in  the configuration.

See codepen: https://codepen.io/stockinail/pen/dyVRZzY

stockiNail:
Fix #581

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -1,4 +1,4 @@
-import {distanceBetweenPoints, callback as callHandler} from 'chart.js/helpers';
+import {distanceBetweenPoints, defined, callback as callHandler} from 'chart.js/helpers';
 
 const clickHooks = ['click', 'dblclick'];
 const moveHooks = ['enter', 'leave'];
@@ -13,7 +13,7 @@ export function updateListeners(chart, state, options) {
     if (typeof options[hook] === 'function') {
       state.listened = true;
       state.listeners[hook] = options[hook];
-    } else {
+    } else if (defined(state.listeners[hook])) {
       delete state.listeners[hook];
     }
   });
--- a/src/events.js
+++ b/src/events.js
@@ -1,4 +1,4 @@
-import {distanceBetweenPoints, defined, callback as callHandler} from 'chart.js/helpers';
+import {distanceBetweenPoints, callback as callHandler} from 'chart.js/helpers';
 
 const clickHooks = ['click', 'dblclick'];
 const moveHooks = ['enter', 'leave'];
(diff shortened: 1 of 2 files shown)


src/events.js
```diff
@@ -13,6 +13,8 @@ export function updateListeners(chart, state, options) {
     if (typeof options[hook] === 'function') {
       state.listened = true;
       state.listeners[hook] = options[hook];
+    } else if (defined(state.listeners[hook])) {
+      delete state.listeners[hook];
```
kurkle, line 17: [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete) says:
> If the property which you are trying to delete does not exist, delete will not have any effect and will return true. 

```suggestion
    } else {
      delete state.listeners[hook];
```

test/events.js
```diff
@@ -27,6 +27,37 @@ export function testEvents(options, innerElement, getInnerPoint) {
   describe('events', function() {
     const pluginOpts = chartConfig.options.plugins.annotation;
 
+    it(`should detect only leave event on ${descr}`, function(done) {
```
kurkle, line 30: I think something like:

```suggestion
    it(`should not call removed hook on ${descr}`, function(done) {
```

would better describe the test.

kurkle (commented):
Couple of nits. Also not that the `defined` helper is not needed anymore, if you agree with my comments.

src/events.js
stockiNail: Yes @kurkle but I have also seen the delete is a costly operation and therefore I want to avoid if not needed.

stockiNail:
@kurkle I didn't set the milestone because I don't know if you want to release it in 1.3.0 or before with another release

src/events.js
kurkle: Turns out FireFox is 90% slower deleting non-existent property vs checking its type. So this suggestion was bad.

kurkle:
> @kurkle I didn't set the milestone because I don't know if you want to release it in 1.3.0 or before with another release

We have already merged bunch of things that are marked for 1.3.0. So would need to do it in a separate branch if we wanted to release 1.2.2 for example. This is not a new bug, so I think it can be released in 1.3.0.

src/events.js
stockiNail: > Turns out FireFox is 90% slower deleting non-existent property vs checking its type. So this suggestion was bad.

Ok, I revert this update

kurkle (approved):

PR #583, opened 2021-12-19
Fix events catching on rotated ellipse annotations

### LINKED ISSUE #580: Rotated ellipse annotation does not catch the events
When an ellipse annotation is rotated, it doesn't catch the events.

This happens because in `inRange` method, the rotation is ignored.

Codepen: https://codepen.io/stockinail/pen/ExwXjEe

The sample is rotated of 90 degrees and if you click out of ellipse (original size with rotation 0), the events is caught even if is outside of the annotation.

stockiNail:
Fix #580

### DIFF
--- a/src/types/ellipse.js
+++ b/src/types/ellipse.js
@@ -5,7 +5,7 @@ import {getRectCenterPoint, getChartRect} from '../helpers';
 export default class EllipseAnnotation extends Element {
 
   inRange(mouseX, mouseY, useFinalPosition) {
-    return pointInEllipse({x: mouseX, y: mouseY}, this.getProps(['x', 'y', 'width', 'height'], useFinalPosition));
+    return pointInEllipse({x: mouseX, y: mouseY}, this.getProps(['x', 'y', 'width', 'height'], useFinalPosition), this.options.rotation);
   }
 
   getCenterPoint(useFinalPosition) {
@@ -68,7 +68,7 @@ EllipseAnnotation.defaultRoutes = {
   backgroundColor: 'color'
 };
 
-function pointInEllipse(p, ellipse) {
+function pointInEllipse(p, ellipse, rotation) {
   const {width, height} = ellipse;
   const center = ellipse.getCenterPoint(true);
   const xRadius = width / 2;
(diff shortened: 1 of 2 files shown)


stockiNail:
@kurkle as usual, I write step by step about math calculations... For sure, you don't like it ;)

stockiNail:
@kurkle I didn't set the milestone because I don't know if you want to release it in 1.3.0 or before with another release

src/types/ellipse.js
```diff
@@ -77,6 +77,11 @@ function pointInEllipse(p, ellipse) {
...
   }
-
-  return (Math.pow(p.x - center.x, 2) / Math.pow(xRadius, 2)) + (Math.pow(p.y - center.y, 2) / Math.pow(yRadius, 2)) <= 1.0;
+  // https://stackoverflow.com/questions/7946187/point-and-ellipse-rotated-position-test-algorithm
+  const angle = toRadians(rotation || 0);
+  const cosAngle = Math.cos(angle);
+  const sinAngle = Math.sin(angle);
+  const a = Math.pow(cosAngle * (p.x - center.x) + sinAngle * (p.y - center.y), 2);
+  const b = Math.pow(sinAngle * (p.x - center.x) - cosAngle * (p.y - center.y), 2);
+  return (a / Math.pow(xRadius, 2)) + (b / Math.pow(yRadius, 2)) <= 1;
```
kurkle, line 86: Do the width and height need to be adjusted?

stockiNail: No, they don't. At least I don't think so. Why?

kurkle: I was thinking it should be the rotated xRadius/yRadius that we need to compare against, but is it?

I think this one would deserve some dedicated tests.

stockiNail: in my opinion, width and height should be affected by a rotation, because they are remaining the same. But I could be wrong.
In fact, in the stackoverflow, where I took the solution, the width and height are not touched by rotation.

stockiNail: > I think this one would deserve some dedicated tests.

added specific test case on rotated ellipse

kurkle (approved):

PR #586, opened 2021-12-22
new annotation types added to ElementOptionsByType

elitastic:
New annotation types added to ElementOptionsByType to be able to set defaults in typescript with correct typings like:

```
import { defaults } from 'chart.js';

defaults.elements.labelAnnotation.font = {
    color: '#999'
};
```


stockiNail:
@elitastic even if not related to changes of PR, be aware that `color` is not an options for font.

```javascript
defaults.elements.labelAnnotation.font = {
    color: '#999' // <-- not a font options
};
```

kurkle (approved):

elitastic:
Thank you @stockiNail.

PR #595, opened 2021-12-29
Fix dynamically showing/hiding line label

### LINKED ISSUE #589: Dynamically showing/hiding labels in response to mouse events
I set up mouse `enter` and `leave` events to dynamically show labels when the user mouses over the annotation:

```
    enter({ chart, element }) {
      if ('label' in element.options && element.options.label) {
        element.options.label.enabled = true;
        chart.draw();
      }
    },
    leave({ chart, element }) {
      if ('label' in element.options && element.options.label) {
        element.options.label.enabled = false;
        chart.draw();
      }
    },
```

This was based off of the old [horizontal-line.html](https://github.com/chartjs/chartjs-plugin-annotation/blob/6a1037f69b90e22cfbe9f61da5b24eefd344aecb/old_samples/horizontal-line.html#L143) sample.

This worked fine until chartjs-plugin-annotation 1.2.0 - specifically, #547. If I'm understanding correctly, the changes in #547 cause `labelVisible` to be precalculated so that it no longer responds to changes in `enabled`.

I tried fixing this by calling `chart.update()` instead of `chart.draw()`, but things still didn't work. I think the code is using its options proxy and context to retrieve the original annotation options object, instead of the dynamically updated element options, but I'm not sure that I fully understand how contexts and options proxies work.

I can directly access the original chart options, but that feels ugly (if using TypeScript, due to its optional properties, and if annotations are in an array instead of keyed by unique ID):

```
        chart.options.plugins.annotation.annotations[0].label.enabled = true;
```

Is dynamically updating element options supported? What's the proper way to do it?

Thank you.

kurkle:
Fix #589


kurkle:
| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/chartjs-plugin-annotation.esm.js` | 11.6 kB | -148 B (-1%) |
| `dist/chartjs-plugin-annotation.js` | 11.7 kB | -143 B (-1%) |
| `dist/chartjs-plugin-annotation.min.js` | 7.88 kB | -88 B (-1%) |

src/types/label.js
```diff
@@ -1,21 +1,18 @@
...
-    return this.visible && inBoxRange(mouseX, mouseY, this.getProps(['x', 'y', 'width', 'height'], useFinalPosition));
+    return this.options.display && inBoxRange(mouseX, mouseY, this.getProps(['x', 'y', 'width', 'height'], useFinalPosition));
   }
 
   getCenterPoint(useFinalPosition) {
     return getRectCenterPoint(this.getProps(['x', 'y', 'width', 'height'], useFinalPosition));
   }
 
   draw(ctx) {
-    if (!this.visible) {
```
stockiNail, line 16: Not sure how it works if the label has an empty content. I think drawCallout will be invoked.

stockiNail: AFAIK a label is displayable if display/enabled  is not undefined and the content is consistent.

src/annotation.js
```diff
@@ -230,7 +230,7 @@ function draw(chart, caller, clip) {
   }
 
   state.visibleElements.forEach(el => {
-    if ('drawLabel' in el && el.options.label && (el.options.label.drawTime || el.options.drawTime) === caller) {
+    if ('drawLabel' in el && el.options.label && el.options.label.enabled && (el.options.label.drawTime || el.options.drawTime) === caller) {
```
stockiNail, line 233: AFAIK a label is displayable if enabled and the content is consistent.

src/types/label.js
kurkle: So is the "empty content case" for hiding a configuration error or is it another way of hiding label annotaton?
if the latter, the use case can be achieved with scriptable `enabled`:
`enabled: (ctx, opts) => !!opts.content`

kurkle: restored the empty content behavior

kurkle:
| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/chartjs-plugin-annotation.esm.js` | 11.6 kB | -139 B (-1%) |
| `dist/chartjs-plugin-annotation.js` | 11.7 kB | -136 B (-1%) |
| `dist/chartjs-plugin-annotation.min.js` | 7.89 kB | -74 B (-1%) |

src/types/label.js
stockiNail: > So is the "empty content case" for hiding a configuration error or is it another way of hiding label annotation?

I think, since version 0.5.7, this was implemented (in line label) to avoid to perform any "label" calculation. In the label annotation we used the same concept.  I don't know why was implemented in this way but I think for both the reasons.
I recall that at that time I thought that `enabled` option was useless because you could check only the content.

stockiNail (approved):

stockiNail:
@kurkle sorry for the stupid question and my poor knowledge but why in the test case you invoke the `chart.draw()` instead of `chart.update()`? Is there some performance implication?

kurkle:
> @kurkle sorry for the stupid question and my poor knowledge but why in the test case you invoke the `chart.draw()` instead of `chart.update()`? Is there some performance implication?

Yes, performance. Update does a lot of things not needed when just showing a label which had it position pre-calculated already.
draw just clears the canvas and draws all the things again, without doing layout etc.

stockiNail:
THANK YOUU!

stockiNail:
@kurkle sorry if I'm jumping again to this topic.

I was wondering why I didn't implement `draw()` method in my lib which wraps CHART.JS. Then, I went to [CHART.JS API doc](https://www.chartjs.org/docs/latest/developers/api.html) and I don't find any `draw()` method. Very similar it seems `render()`.
because I don't know why (bug or not?), I'm going to open a question in CHART.JS for that and move the possible further discussions overthere.

PR #600, opened 2021-12-29
Enable shadowing on  the annotations

### LINKED ISSUE #145: Question: How to add a shadow around horizontal line
I´m drawing a horizontal line that represents average speed during a day. Now I want to add a horizontal shadow (or similar) around that line to represent standard deviation. 

Does this plugin allow what I need?

stockiNail:
Fix #145

This PR is enabling shadowing on the annotations, only on the shapes (no text of labels).
The shadowing is applied when a shape is fill, in order to apply the the fill to the shadow as well. Instead, the shadow is reset before the `stroke`.

New options:

| Name | Type | Default| Descriptions
| :-- | :-- | :-- | :-- 
| backgroundShadowColor | Color | 'transparent'| The color of shadows. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/shadowColor).
| borderShadowColor | Color | 'transparent'| The color of border shadows. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/shadowColor).
| shadowBlur | number | 0 | The amount of blur applied to shadows. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/shadowBlur).
| shadowOffsetX | number | 0 | The distance that shadows will be offset horizontally. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/shadowOffsetX).
| shadowOffsetY | number | 0 | The distance that shadows will be offset  vertically. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/shadowOffsetY).

### TODO

- [x] test cases
- [x] type definitions
- [x] samples
- [x] documentation


src/types/box.js
```diff
@@ -75,6 +75,10 @@ BoxAnnotation.defaults = {
     yAdjust: 0,
     width: undefined
   },
+  shadowBlur: 0,
+  shadowColor: undefined,
```
kurkle, line 79: Should probably be transparent, because setting undefined to ctx usually results it to keeping the previously set value, which could be anything leaked from some other plugin etc.

We should also look into having some kind of "common defaults", to avoid the repeated stuff.

stockiNail:
### Known issue

Using the `drawPoint` function, exposed by Chartjs, it's not possible to reset the shadow options before `stroke()`. This is causing that the shadow border overlaps the original shape of the annotation.

@kurkle I start implementing shadowing addressing the shadow of the shape more than the shadow of the border. For this reason I'm resetting the shadow setting on the context before to stroke the borders.

What do you think? If the above decision (to make the shadow only of  the shape), probably I need to "copy" the `drawPoint` function from Chartjs and put it in the point annotation, updating in order to respect the shadow of the shape and avoiding the shadow border overlapping. But not sure if there is another way.

src/types/box.js
stockiNail: > Should probably be transparent, because setting undefined to ctx usually results it to keeping the previously set value, which could be anything leaked from some other plugin etc.

I had a doubt if to set to undefined or transparent. The MDN documentation says:

```
Shadows are only drawn if the shadowColor property is set to a non-transparent value. 
One of the shadowBlur, shadowOffsetX, or shadowOffsetY properties must be non-zero, as well. 
```

Now I set to transparent to be aligned with MDN doc. Nevertheless I used undefined because I see that in Datalabels, where it is used, the default was undefined but offsets are not configurable but only blur which is set to 0 for reset.

kurkle:
> ### Known issue
> Using the `drawPoint` function, exposed by Chartjs, it's not possible to reset the shadow options before `stroke()`. This is causing that the shadow border overlaps the original shape of the annotation.
> 
> @kurkle I start implementing shadowing addressing the shadow of the shape more than the shadow of the border. For this reason I'm resetting the shadow setting on the context before to stroke the borders.
> 
> What do you think? If the above decision (to make the shadow only of the shape), probably I need to "copy" the `drawPoint` function from Chartjs and put it in the point annotation, updating in order to respect the shadow of the shape and avoiding the shadow border overlapping. But not sure if there is another way.

We could pass it borderWidth: 0 and do the stroke ourselves:
https://github.com/chartjs/Chart.js/blob/a5b46a859c6398618ec4787222b35bdd09b9f834/src/helpers/helpers.canvas.js#L238-L240

stockiNail:
> We could pass it borderWidth: 0 and do the stroke ourselves:

Let me try. I was convinced that passing borderWidth: 0, some point styles would not be drawn

stockiNail:
@kurkle it works!!! you're awesome

stockiNail:
@kurkle I put again in draft. I'm working to enable also the shadow on the stroke

src/helpers/helpers.canvas.js
```diff
@@ -26,6 +26,27 @@ export function setBorderStyle(ctx, options) {
...
+}
+
+/**
+ * Resets shadow options to the canvas context before stroking,
+ * in order to manage teh shadow at "fill" level
+ * @param {CanvasRenderingContext2D} ctx - chart canvas context
+ */
+export function resetShadow(ctx) {
+  ctx.shadowColor = 'transparent';
+}
```
LeeLenaleee, line 48: Either all shadown propertys should be reset as the descriptions and fucntion name suggest or the description and name should be changed to something like `setShadowColorTransparent` and the description something like `set shadow color to transparent before stroking in order to manage the shadow at "fill" level`.

Also small nitpick `the` on line 43 between `manage` and `shadow` has a typo

types/options.d.ts
```diff
@@ -32,6 +32,10 @@ export interface CoreAnnotationOptions extends AnnotationEvents {
   value?: Scriptable<number|string, PartialEventContext>,
   xScaleID?: Scriptable<string, PartialEventContext>,
   yScaleID?: Scriptable<string, PartialEventContext>,
+  shadowColor: Scriptable<Color, PartialEventContext>,
+  shadowBlur: Scriptable<number, PartialEventContext>,
+  shadowOffsetX: Scriptable<number, PartialEventContext>,
+  shadowOffsetY: Scriptable<number, PartialEventContext>
```
LeeLenaleee, line 38: Missing the `?` behind the property name to make them optional

types/label.d.ts
```diff
@@ -106,6 +106,10 @@ export interface LabelOptions extends ContainedLabelOptions {
    * @deprecated
    */
   yPadding?: Scriptable<number, PartialEventContext>,
+  shadowColor: Scriptable<Color, PartialEventContext>,
+  shadowBlur: Scriptable<number, PartialEventContext>,
+  shadowOffsetX: Scriptable<number, PartialEventContext>,
+  shadowOffsetY: Scriptable<number, PartialEventContext>
```
LeeLenaleee, line 112: Missing the `?` behind the property name to make them optional

src/helpers/helpers.canvas.js
kurkle: I have to disagree slightly. I would not create a function called `setShadowColorToTransparent` in any case. Maybe a better name could be `disableShadow`.

The description however is off, because in this function you can not know if its called between fill and stroke. I'd remove the description and maybe add some comment in the code using this before stroke.

test/fixtures/point/canvas.js
```diff
@@ -1,6 +1,6 @@
 const canvas = window.createCanvas();
 module.exports = {
-  tolerance: 0.0055,
+  tolerance: 0.0095,
```
kurkle, line 3: How and why does it affect this fixture?

src/helpers/helpers.canvas.js
stockiNail: :+1:

test/fixtures/point/canvas.js
stockiNail: image

stockiNail: I don't know why exactly but it seems the canvas provided as pointStyle works well, but there a difference on chartArea border.

stockiNail: I was convinced to be related to chart area but now having a look better sounds related to the shadow, when pointStyle is a canvas

stockiNail: My fault, apologize

types/options.d.ts
stockiNail: :+1: done!

types/label.d.ts
stockiNail: :+1: done!

test/fixtures/point/canvas.js
stockiNail: I forgot that pointStyle could be an image or canvas, where border is not applicable.

src/types/point.js
```diff
@@ -1,6 +1,8 @@
 import {Element} from 'chart.js';
 import {drawPoint} from 'chart.js/helpers';
-import {inPointRange, getElementCenterPoint, resolvePointPosition, setBorderStyle} from '../helpers';
+import {inPointRange, getElementCenterPoint, resolvePointPosition, setBorderStyle, setShadowStyle, removeShadowStyle, isImageOrCanvas} from '../helpers';
+
+const isFillablePointStyle = (ps) => !(typeof ps === 'string') || ps === 'circle' || ps === 'triangle' || ps.startsWith('rect');
```
kurkle, line 5: You probably guessed, I don't like this.

What about separate fill / border (or line) shadow settings, so the responsibility is transferred to the user.
Eg. `fillShadowColor` only to fill and `borderShadowColor` to stroke?

stockiNail: Yes, I thought you don't like it.
Yesterday during developing, I was thinking about a `type` but I think fill/border shadow color could be easier to understand.
Let me change it and we will see how it will look like!

stockiNail: maybe `fillShadowColor` and `strokeShadowColor` ? Sounds to me closer to the context operations.
And  what about `backgroundShadowColor` and `borderShadowColor` ?

kurkle: I think we are using background/border to be closer to css. Probably good to stick with those, but I'm not very confident about this.

stockiNail: used background/border shadow color

docs/guide/types/box.md
```diff
@@ -91,13 +96,18 @@ If one of the axes does not match an axis in the chart, the box will take the en
...
 | `borderDashOffset` | Offset for border line dashes. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineDashOffset).
 | `borderJoinStyle` | Border line joint style. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineJoin).
 | [`borderRadius`](#borderradius) | Radius of box rectangle (in pixels)
+| `borderShadowColor` | The color of the border shadow. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/shadowColor).
 | `borderWidth` | Border line width (in pixels).
+| `shadowBlur` | The amount of blur applied to shadow. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/shadowBlur).
+| `shadowOffsetX` | The distance that shadow will be offset horizontally. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/shadowOffsetX).
+| `shadowOffsetY` | The distance that shadow will be offset  vertically. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/shadowOffsetY).
```
kurkle, line 110: double space between `offset` and `vertically`

not sure if we should mention these apply to both shadow types or if we should just have all the properties for both shadow types separately?

docs/guide/types/ellipse.md
```diff
@@ -65,7 +65,12 @@ The following options are available for ellipse annotations.
 | [`borderWidth`](#styling) | `number`| Yes | `1`
 | [`borderDash`](#styling) | `number[]`| Yes | `[]`
 | [`borderDashOffset`](#styling) | `number`| Yes | `0`
+| [`borderShadowColor`](#styling) | [`Color`](../options#color) | Yes | `'transparent'`
 | [`backgroundColor`](#styling) | [`Color`](../options#color) | Yes | `options.color`
+| [`backgroundShadowColor`](#styling) | [`Color`](../options#color) | Yes | `'transparent'`
```
kurkle, line 70: wondering why background is after border here. would be ok to fix here if you want.

docs/guide/types/ellipse.md
```diff
@@ -92,4 +97,10 @@ If one of the axes does not match an axis in the chart, the ellipse will take th
 | `borderWidth` | Stroke width
 | `borderDash` | Length and spacing of dashes. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/setLineDash).
 | `borderDashOffset` | Offset for line dashes. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineDashOffset).
+| `borderShadowColor` | The color of the border shadow. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/shadowColor).
 | `backgroundColor` | Fill color
+| `backgroundShadowColor` | The color of shadow. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/shadowColor).
```
kurkle, line 102: same here, background should be before border

src/helpers/helpers.canvas.js
```diff
@@ -80,6 +93,8 @@ export function drawBox(ctx, rect, options) {
   ctx.closePath();
   ctx.fill();
   if (stroke) {
+    // sets shadow color for border
```
kurkle, line 96: these comments were probably more meaningful before changing to background/borderShadowColor, now they are quite useless can be removed (all instances)

kurkle (commented):
Should have tests for both background and border shadow in the same shape, but lets have some more reviews first.

docs/guide/types/ellipse.md
stockiNail: All the options in the doc must be reordered because they are not sorted alphabetically. It was already in my TODO list.

docs/guide/types/ellipse.md
stockiNail: All the options in the doc must be reordered because they are not sorted alphabetically. It was already in my TODO list.

docs/guide/types/ellipse.md
stockiNail: a specific PR is needed

docs/guide/types/ellipse.md
stockiNail: a specific PR is needed

src/helpers/helpers.canvas.js
stockiNail: :+1:

docs/guide/types/box.md
stockiNail: > not sure if we should mention these apply to both shadow types or if we should just have all the properties for both shadow types separately?

Could make sense. The topic could be valid also for other common options I guess. If you agree, let me add it in my list and I'll submit a specific PR.

etimberg: I think the space issue affects all the different annotation types

stockiNail: > double space between `offset` and `vertically`

done

src/helpers/helpers.canvas.js
stockiNail: done

stockiNail:
> Should have tests for both background and border shadow in the same shape, but lets have some more reviews first.

Added test case (1 for annotation type) where background and border shadow color are used.

src/types/point.js
```diff
@@ -15,11 +15,19 @@ export default class PointAnnotation extends Element {
...
   draw(ctx) {
     const options = this.options;
+    const borderWidth = options.borderWidth;
     ctx.save();
     ctx.fillStyle = options.backgroundColor;
-    setBorderStyle(ctx, options);
+    setShadowStyle(ctx, options);
+    const stroke = setBorderStyle(ctx, options);
+    options.borderWidth = 0;
     drawPoint(ctx, options, this.x, this.y);
```
etimberg, line 24: Might be worth having a test to ensure that `borderWidth = 0` correctly skips in this helper. We could add a test in this package to know if that ever breaks when upgrading the chart.js version. An alternative is to add a test in the main chart.js repo about this case

etimberg (dismissed):
Overall looks good. I was going to comment about adding tests with a different `shadowBorderColor` but I see those just got added. I left one minor comment about testing the point borderWidth == 0 behaviour to know if chart.js ever breaks that

src/types/point.js
stockiNail: I think it's better to add a test here. Let me implement it.

stockiNail:
> shadowBorderColor

@etimberg you mentioned `shadowBorderColor` but be aware the options is called borderShadowColor. Do you prefer shadowBorderColor and shadowBackgroundColor instead of borderShadowColor and backgroundShadowColor?

etimberg:
> > shadowBorderColor
> 
> @etimberg you mentioned `shadowBorderColor` but be aware the options is called `borderShadowColor`. Do you prefer `shadowBorderColor` and `shadowBackgroundColor` instead of `borderShadowColor` and `backgroundShadowColor`?

I think your choices are fine. I just had the wrong thing in mind when I typed out my reply

src/types/point.js
stockiNail: I have added 2 test cases for point annotation with `borderWidth: 0` (1 with shadow and 1 without). If something will change in Chart.js, breaking the plugin, we should be able to recognize it

kurkle (approved):

etimberg (approved):

LeeLenaleee (approved):

PR #601, opened 2021-12-30
Combine common defaults

kurkle:
Draft to see the size effect. Probably not worth the added complexity.

### DIFF
--- a/src/types/_commonDefaults.js
+++ b/src/types/_commonDefaults.js
@@ -1,0 +1,12 @@
+export const common = {
+  adjustScaleRange: true,
+  borderDash: [],
+  borderDashOffset: 0,
+  display: true,
+  xMax: undefined,
+  xMin: undefined,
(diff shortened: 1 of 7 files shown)


kurkle:
Yeah, really not worth it.

| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/chartjs-plugin-annotation.esm.js` | 11.6 kB | +2 B (0%) |
| `dist/chartjs-plugin-annotation.min.js` | 7.84 kB | -14 B (0%) |

stockiNail:
> We should also look into having some kind of "common defaults", to avoid the repeated stuff.

@kurkle Apologize for my late reply but yesterday I was busy on other stuff. I see your closed PR and your comment. 
Do we stay as is?

kurkle:
Yes, lets stay as is, because there is nothing to gain by combining.

PR #603, opened 2021-12-30
Fixes events triggering on the label of a line annotation

### LINKED ISSUE #602: Events on line label is not triggered
The events on the label of the line is not triggered if the event is fired only in the label (not on the line).

This bug seems to be generated because `inRange` method is invoking in cascade mode `isOnLabel` and `labelIsVisible`, without any `chartArea` instance.

https://github.com/chartjs/chartjs-plugin-annotation/blob/4d6d66df6ceffa8387441f116a5f130b636bab9d/src/types/line.js#L67-L74

stockiNail:
Fix #602

Adds additional test cases

### DIFF
--- a/src/types/line.js
+++ b/src/types/line.js
@@ -67,10 +67,10 @@ export default class LineAnnotation extends Element {
   // TODO: make private in v2
   labelIsVisible(useFinalPosition, chartArea) {
     const labelOpts = this.options.label;
-    if (!chartArea || !labelOpts || !labelOpts.enabled) {
+    if (!labelOpts || !labelOpts.enabled) {
       return false;
     }
-    return isLineInArea(this.getProps(['x', 'y', 'x2', 'y2'], useFinalPosition), chartArea);
+    return !chartArea || isLineInArea(this.getProps(['x', 'y', 'x2', 'y2'], useFinalPosition), chartArea);
   }
 
   // TODO: make private in v2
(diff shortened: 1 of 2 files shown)


stockiNail:
FYI I have discovered it because Coveralls has marked as red the coverage of `helpers.geometric` where the `rotate` function (used in `isOnLabel`) was never tested,

src/types/line.js
```diff
@@ -67,10 +67,10 @@ export default class LineAnnotation extends Element {
   // TODO: make private in v2
   labelIsVisible(useFinalPosition, chartArea) {
     const labelOpts = this.options.label;
-    if (!chartArea || !labelOpts || !labelOpts.enabled) {
+    if (!labelOpts || !labelOpts.enabled) {
       return false;
     }
-    return isLineInArea(this.getProps(['x', 'y', 'x2', 'y2'], useFinalPosition), chartArea);
+    return !chartArea || isLineInArea(this.getProps(['x', 'y', 'x2', 'y2'], useFinalPosition), chartArea);
```
kurkle, line 73: Oh right, the chartArea is optional parameter that makes this function call more precise. Could add some jsdoc to indicate the parameter is optional (and not just undefined in some scenario)

kurkle (approved):

LeeLenaleee (approved):

src/types/line.js
stockiNail: I'll submit another PR, adding JSdoc

PR #607, opened 2022-01-03
Remove `point` node from label annotation sample

stockiNail:
The `point` node from label annotation sample because it is not valid anymore.

### DIFF
--- a/docs/samples/label/point.md
+++ b/docs/samples/label/point.md
@@ -30,13 +30,6 @@ const annotation1 = {
     right: 6,
     bottom: 12
   },
-  point: {
-    enabled: true,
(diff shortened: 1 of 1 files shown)


LeeLenaleee:
Well it is already removed in code and released but shouldn't this have waited for a major version increase since its a breaking change?

LeeLenaleee (approved):

stockiNail:
@LeeLenaleee the code about point node removing has been in the version 1.2.0 when the label has been released ( #558).

Therefore there hasn't been any breaking change because we have done before releasing label annotation feature (#502).

LeeLenaleee:
Ooh MB, thanks for pointing it out 👍

PR #608, opened 2022-01-03
Add arrows decoration to line annotation

stockiNail:
This PR is adding the optional arrowheads to a line annotation.

### TODO

- [x] test cases
- [x] types
- [x] documentation
- [x] samples


stockiNail:
It's still a working progress. There is a bug that I'm having a look at.

stockiNail:
@kurkle before going on, when you have time (no hurry), could you check if there is anything to change or you don't like it?

stockiNail:
I set milestone 1.3.0 but feel free to change it.

docs/guide/types/line.md
```diff
@@ -51,6 +51,8 @@ The following options are available for line annotations. All of these options c
 | Name | Type | [Scriptable](../options#scriptable-options) | Default
 | ---- | ---- | ---- | :----: | ----
 | [`adjustScaleRange`](#general) | `boolean` | Yes | `true`
+| [`arrowHeads`](#arrow-heads) | `{start: object, end:object}` | Yes |
+| [`backgroundShadowColor`](#styling) | [`Color`](../options#color) | Yes | `'transparent'`
```
kurkle, line 55: If this only applies to arrowHeads, then it should be nested in there.

src/types/line.js
```diff
@@ -368,3 +397,50 @@ function adjustLabelCoordinate(coordinate, labelSizes) {
...
+  }
+  const {length, width} = arrowOpts;
+  const adjust = line.options.borderWidth / 2;
+  const p1 = {x: length, y: width + adjust};
+  const p2 = {x: 0, y: adjust};
+  return Math.abs(interpolateX(0, p1, p2));
+}
+
+function drawArrowHead(ctx, {offset, adjust}, arrowOpts, options) {
+  if (!arrowOpts || !arrowOpts.display) {
```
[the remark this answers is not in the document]
stockiNail: Ok!!

src/types/line.js
```diff
@@ -368,3 +397,50 @@ function adjustLabelCoordinate(coordinate, labelSizes) {
...
+  ctx.moveTo(arrowOffsetX, -width);
+  ctx.lineTo(offset + adjust, 0);
+  ctx.lineTo(arrowOffsetX, width);
+  if (fill === true) {
+    ctx.fillStyle = backgroundColor || borderColor;
+    ctx.closePath();
+    ctx.fill();
+    ctx.shadowColor = 'transparent';
+  } else {
+    ctx.shadowColor = options.borderShadowColor;
```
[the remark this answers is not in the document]
stockiNail: For shadows options, we can even if the shadow should be consistent with the line one. About borderWidth, I did in this way for 2 reasons:

1. arrows should be an extension of the line and  I think the thickness of the border should be the same
2. for visual standpoint, if the borderWidth of the line is different of the arrows ones, the line can be draw beyond the arrows

test/fixtures/line/borderWidth0.js
```diff
@@ -0,0 +1,105 @@
...
+        y: {
+          display: true,
+          min: 0,
+          max: 100
+        }
+      },
+      plugins: {
+        annotation: {
+          annotations: {
+            line: {
```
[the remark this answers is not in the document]
stockiNail: done!

src/types/line.js
```diff
@@ -190,8 +203,24 @@ export default class LineAnnotation extends Element {
...
+  borderDashOffset: 0,
+  display: false,
+  fill: false,
+  length: 12,
+  width: 6
+};
+
 LineAnnotation.defaults = {
   adjustScaleRange: true,
+  arrowHeads: {
```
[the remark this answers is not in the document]
stockiNail: > Both of these can be done later, so these are not suggested changes, but mere comments.
> 
> if the `start` and `end` was described with `_fallback: true`, then you could place common options directly in arrowHeads and only put the things different for start/end in those objects.

Understood! Additional PR in the near future (in the meanwhile I put it in my TODO)?

> I think it should also include borderWidth, so it does not have to equal the border width of the actual line.

About borderWidth, I did in this way for 2 reasons:

1. arrows should be an extension of the line and I think the thickness of the border should be the same
2. for visual standpoint, if the borderWidth of the line is different of the arrows ones, the line can be draw beyond the arrows

docs/guide/types/line.md
stockiNail: done

src/types/line.js
```diff
@@ -107,15 +107,28 @@ export default class LineAnnotation extends Element {
...
   draw(ctx) {
     const {x, y, x2, y2, options} = this;
+
     ctx.save();
+    const stroke = setBorderStyle(ctx, options);
+    // no border width, then line is not drawn
+    if (!stroke) {
+      ctx.restore();
+      return;
+    }
```
kurkle, line 117: The `stroke` variable is used only for the if, so this block can be condensed.

couple of different styles to do the same thing (saving couple of bytes in size):

```js
    if (!setBorderStyle(ctx, options)) {
      // no border width, then line is not drawn
      return ctx.restore();
    }
```

OR

```js
    if (setBorderStyle(ctx, options)) {
      // the drawing stuff
    }
    ctx.restore();
```

src/types/line.js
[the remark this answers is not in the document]
stockiNail: I'm testing the fallback: true right now, as you suggested

test/fixtures/line/shadow.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
               yMax: 8,
               borderColor: 'blue',
               borderWidth: 4,
-              backgroundShadowColor: 'black',
+              borderShadowColor: 'black',
```
kurkle, line 21: Why are these needed and does the fixture actually catch an missing shadow in this test?

test/fixtures/line/shadowArrowHeads.js
```diff
@@ -0,0 +1,119 @@
...
+            },
+            {
+              type: 'line',
+              xMin: 1,
+              xMax: 8,
+              yMin: 5,
+              yMax: 5,
+              borderColor: 'purple',
+              borderWidth: 4,
+              backgroundShadowColor: 'black',
```
kurkle, line 50: This is not used, right?

test/fixtures/line/shadowArrowHeads.js
```diff
@@ -0,0 +1,119 @@
...
+            },
+            {
+              type: 'line',
+              xMin: 1,
+              xMax: 8,
+              yMin: 2,
+              yMax: 2,
+              borderColor: 'red',
+              borderWidth: 4,
+              backgroundShadowColor: 'black',
```
kurkle, line 83: This is not used, right?

src/types/line.js
stockiNail: done (options 1)

test/fixtures/line/shadowArrowHeads.js
stockiNail: yes, removed

test/fixtures/line/shadowArrowHeads.js
stockiNail: yes, removed

test/fixtures/line/shadow.js
stockiNail: backgroundColor is not needed because as suggested, has been moved to arrow heads config.
borderColor is needed to apply the shadow to the line.
And yes, the fixture catches any missing shadow.

src/types/line.js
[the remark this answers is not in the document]
stockiNail: I have enabled the `borderWidth` at arrow level. I think we will get some issues about that (hopefully I'm wrong!).

src/types/line.js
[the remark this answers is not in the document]
stockiNail: I have enabled the `borderWidth` at arrow level. I think we will get some issues about that (hopefully I'm wrong!).

src/types/index.js
```diff
@@ -36,4 +36,16 @@ Object.keys(annotationTypes).forEach(key => {
   defaults.describe(`elements.${annotationTypes[key].id}`, {
     _fallback: 'plugins.annotation'
   });
+  if (key === 'line') {
```
kurkle, line 39: Placing these to `LineAnnotation.descriptors` should work.

Eg: (in line.js)

```js
LineAnnotation.descriptors = {
  arrowHeads: {
    _fallback: true
  }
}
```

Are the `.start` and `.end` descriptors actually needed?

src/types/line.js
```diff
@@ -190,9 +201,44 @@ export default class LineAnnotation extends Element {
...
+  shadowOffsetX: undefined,
+  shadowOffsetY: undefined,
+  width: undefined
+};
+
 LineAnnotation.defaults = {
   adjustScaleRange: true,
-  backgroundShadowColor: 'transparent',
+  arrowHeads: {
+    backgroundColor: undefined,
```
kurkle, line 225: Probably works fine without the `undefined` values at this level. But common defaults would be preferred at this level (and when it should be distinct by default from line defaults).

For example do we prefer arrows to have borderDash by default when `defaults.elements.lineAnnotation.borderDash = [3, 3]`?

kurkle (commented):
The fixtures, could you combine some of those. The "borderWidht0" ones can be added as single arrows in the previous fixture. The "removed at runtime" fixtures can be combined into single one. (removing end, start, end+start from 3 different annotations in the same chart)

src/types/index.js
stockiNail: Removing them, it doesn't do any fallback, at least in my test case.

stockiNail: but I moved in

```javascript
LineAnnotation.descriptors = {
  arrowHeads: {
    start: {
      _fallback: true
    },
    end: {
      _fallback: true
    },
    _fallback: true
  }
};
```

src/types/line.js
stockiNail: In my opinion yes, because is the line the main object and should define the undefined options for the arrows heads.
Let me know if you disagree.

kurkle: agreed, I think you should then remove everything else but `start` and `end` from `arrowHeads`. Options can be given at that level, but we don't have any defaults there.

stockiNail:
> Remove all redundant options like: `xScaleID: 'x'` (things that specify the default), as @kurkle already started doing in #631 to make it a bit more readable

It should be ok now

src/types/index.js
stockiNail: done

kurkle:
| `dist/chartjs-plugin-annotation.esm.js` | 12.3 kB | +465 B (+4%) |  |
| `dist/chartjs-plugin-annotation.js` | 12.4 kB | +469 B (+4%) |  |
| `dist/chartjs-plugin-annotation.min.js` | 8.39 kB | +391 B (+5%) | 🔍 |

src/types/line.js
```diff
@@ -190,9 +201,44 @@ export default class LineAnnotation extends Element {
...
+  shadowBlur: undefined,
+  shadowOffsetX: undefined,
+  shadowOffsetY: undefined,
+  width: undefined
+};
+
 LineAnnotation.defaults = {
   adjustScaleRange: true,
-  backgroundShadowColor: 'transparent',
+  arrowHeads: {
```
kurkle, line 224: I don't think you replied to this "remove everything else" suggestion:

```suggestion
  arrowHeads: {
    end: Object.assign({}, arrowHeadsDefaults),
    length: 12,
    start: Object.assign({}, arrowHeadsDefaults),
    width: 6
  }
```

does it not work?

stockiNail:
it sounds working well. Thank you

kurkle (approved):

PR #614, opened 2022-01-07
Improve event handling adding test cases

stockiNail:
This PR is adding some additional test cases for event handling, in order to test specific condition in the code.

Furthermore it removes from `event.js` : 

1. a useless check on `state.annotations` because that instance can never be `undefined`. 
2. a if statement `if (!state.moveListened) {` because the `state.moveListened` could be always `false` to reach that code.

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -31,12 +31,14 @@ export function updateListeners(chart, state, options) {
           }
         });
       }
-      moveHooks.forEach(hook => {
-        if (typeof scope[hook] === 'function') {
-          state.listened = true;
-          state.moveListened = true;
-        }
-      });
+      if (!state.moveListened) {
+        moveHooks.forEach(hook => {
+          if (typeof scope[hook] === 'function') {
+            state.listened = true;
+            state.moveListened = true;
+          }
+        });
+      }
     });
   }
 }
--- a/src/events.js
+++ b/src/events.js
@@ -31,14 +31,12 @@ export function updateListeners(chart, state, options) {
           }
         });
       }
-      if (!state.moveListened) {
-        moveHooks.forEach(hook => {
-          if (typeof scope[hook] === 'function') {
-            state.listened = true;
-            state.moveListened = true;
-          }
-        });
-      }
+      moveHooks.forEach(hook => {
+        if (typeof scope[hook] === 'function') {
+          state.listened = true;
+          state.moveListened = true;
+        }
+      });
     });
   }
 }
--- a/src/events.js
+++ b/src/events.js
@@ -5,7 +5,6 @@ const moveHooks = ['enter', 'leave'];
 export const hooks = clickHooks.concat(moveHooks);
 
 export function updateListeners(chart, state, options) {
-  const annotations = state.annotations || [];
   state.listened = false;
   state.moveListened = false;
 
@@ -24,7 +23,7 @@ export function updateListeners(chart, state, options) {
   });
 
   if (!state.listened || !state.moveListened) {
-    annotations.forEach(scope => {
+    state.annotations.forEach(scope => {
       if (!state.listened) {
         clickHooks.forEach(hook => {
           if (typeof scope[hook] === 'function') {
--- a/test/events.js
+++ b/test/events.js
@@ -146,6 +146,23 @@ export function testEvents(options, innerElement, getInnerPoint) {
         });
       });
 
+      it(`should not detect any events (because unmanaged event) on ${descr}`, function(done) {
+        const clickSpy = jasmine.createSpy('click');
+
+        targetOptions.click = clickSpy;
+        pluginOpts.annotations = [options];
+
+        const chart = window.acquireChart(chartConfig);
+        const eventPoint = getEventPoint(chart, getInnerPoint);
+
+        window.afterEvent(chart, 'touchstart', function() {
+          expect(clickSpy.calls.count()).toBe(0);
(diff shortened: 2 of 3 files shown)


src/events.js
```diff
@@ -24,22 +23,20 @@ export function updateListeners(chart, state, options) {
...
       }
-      if (!state.moveListened) {
-        moveHooks.forEach(hook => {
-          if (typeof scope[hook] === 'function') {
-            state.listened = true;
-            state.moveListened = true;
-          }
-        });
-      }
+      moveHooks.forEach(hook => {
```
kurkle, line 34: not sure why you'd like to remove this optimization?

stockiNail: @kurkle the  `if (!state.moveListened) {` is useless in that point of code because it will be always `false`. There is the previous check at row 26:

 ` if (!state.listened || !state.moveListened) {`

`state.moveListened` is part of state.listened, therefore it's NOT possible to have `state.moveListened = true` and `state.listened = false`, therefore at that point of code the `state.moveListened` can be only `true`

kurkle: That check is before the loop (forEach) that goes through all the annotations. 
The removed check is inside that loop.

This is a "short-circuit" optimization, if one annotation has a "move hook", no need to check the other annotations.

We could debate if this optimization has any real effect, and that would need some benchmarking done.

stockiNail: > The removed check is inside that loop.

yes, but the check is always true therefore it's quite useless.

> We could debate if this optimization has any real effect, and that would need some benchmarking done.

I don't know, and I agree it could need some benchmarking. Anyway I didn't submit the PR to improve the optimization but to avoid useless check and to improve maintainability.
I have to admit that it was tricky to identify it. The key here is if moveListened is true, listened is true as well. ;)

stockiNail: @kurkle maybe you have already understood, I found this using Coveralls report. Now 95% of the code is tested therefore it was weird that some code is never tested. Could be for a missing test case or for other reasons. In this case, that if statement couldn't be tested because the check was always true.

Maybe (but I'm not sure) there could be 2 states, 1 for click and  for move listeners instead of 1 for all listeners and 1 for move. But in this case (apart to review and change the code) the risk id to duplicate the objects scanning to retrieve the hooks.

kurkle: I think we just don't have a test with something like: 
```js
annotations: {
  first: { ... }, //moveListened = false
  second: { ..., enter: function() }, // moveListened =  true
  third: { ... }, // not checked, because was true already
}
```

stockiNail: I can add it to this PR.

But just for my understanding, doesn't the third annotation have any hook?

kurkle: > But just for my understanding, doesn't the third annotation have any hook?

should not make any difference either way (in this context)

stockiNail: I have added the test case. Hopefully it is as you thought,

kurkle: I was trying to explain a test that would cover both branches of the `if` that is removed.

stockiNail: I don't think this is applicable because you are reaching that statement (`if`) only when `state.moveListened` is `false`. The `if` statement at the beginning of that code block makes the difference

| listened | moveListened | in code block? | note
| -- | -- | -- | --
| `true` | `true` |  `false` | an enter or leave callback has been defined
| `true` | `false` |  `true` | any enter or leave callbacks has been defined, but a click one
| `false` | `true` |  `false` |  **this is a not possible condition because if `moveListened` is true, `listened` is true as well**
| `false` | `false` |  `true` | any event callbacks has been defined

As you can see, you can entry in the code block only with `moveListened = false`

kurkle: Yes, I know it will not enter in the block when moveListened is already true. But in the test case I described is it not, right?
So we enter the forEach LOOP
then we find an annotation that listens for move. we set moveListened to true.
your proposal is to continue checking the move events even if we already know some of the annotations are listening for them, where as the original code stops checking when it knows at least one annotaion listens for move events.

consider this:
```js
let i = 0;
if (i === 0) {
 for (; i < 1E100; i++) {
   if (i !== 0) break; // can this be reached? we made sure i === 0 before.
   doSomeOperation(); // how many times is this done if the above break is removed?
 }
}
```

again we can debate if that optimization is useless and / or handled by some or all javascript engines, but your pr is not about that.

stockiNail: I'm a bit confused and forgive me if I cannot understand. 

> your proposal is to continue checking the move events even if we already know some of the annotations are listening for them

The above sentence is not correct because  I'm proposing to continue checking the move events because **we are sure that no annotation is listening for them**, and therefore we should check in `annotation` node.

> where as the original code stops checking when it knows at least one annotation listens for move events.

The above sentence is not correct as well because the original code was checking something that is always true and therefore it didn't stop any checking.

kurkle: How do you explain this: https://codepen.io/kurkle/pen/mdBaZzY

stockiNail: Ok, I'm very stupid and don't kill me for that.... What I wrote was wrong because I didn't recognize that the if is inside a forEach, or better I have seen but I don't know why I didn't consider it (even if you write me several times).

Really sorry.... I'll revert that if and I'll add a specific test case

kurkle (approved):

stockiNail:
@kurkle I reverted the removing of the `if` statement and apologize again... THANK YOU, as usual!

PR #615, opened 2022-01-07
Improve event handling on point annotation adding test cases

stockiNail:
This PR is adding a specific test case in order to test `inRange` method when `radius` of the point annotation is 0.

Furthermore, it changes the invocation on `isPointRange`, passing radius and `borderWidth` separately. There was a case where the `radius` could be set to 0 and `borderWidth` to another figure and the `isPointRange` could return `true` (instead of `false`).


stockiNail:
Closed din favor to a common update to all annotations to take care of border thickness.

PR #617, opened 2022-01-07
Remove `readValueToProps` function of options helpers

stockiNail:
This PS is removing part of the code, currently not used because the function was taken by Chart.js (changing the default) but it had additional features not used in this plugin.

### DIFF
--- a/src/helpers/helpers.options.js
+++ b/src/helpers/helpers.options.js
@@ -37,20 +37,18 @@ export function calculateTextAlignment(size, options) {
   return x;
 }
 
-function readValueToProps(value, keys, defValue) {
-  const ret = {};
-  const read = isObject(value)
-    ? prop => value[prop]
-    : () => value;
-
-  for (const prop of keys) {
-    ret[prop] = valueOrDefault(read(prop), defValue);
-  }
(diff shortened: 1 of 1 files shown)


src/helpers/helpers.options.js
```diff
@@ -37,14 +37,10 @@ export function calculateTextAlignment(size, options) {
   return x;
 }
 
-function readValueToProps(value, props, defValue) {
+function readValueToProps(value, keys, defValue) {
```
kurkle, line 40: Because this function is only used in `toPosition`, I think it can be as well inlined there.

```js
export function toPosition(value) {
  if (isObject(value)) {
    return {
      x: valueOrDefault(value.x, 'center'),
      y: valueOrDefault(value.y, 'center'),
    };
  }
  value = valueOrDefault(value, 'center');
  return {
    x: value,
    y: value
  };
}
```

stockiNail: I agree. I'll update it

stockiNail: done

kurkle (approved):

PR #619, opened 2022-01-10
Properly consider the border thickness on the annotations' events

### LINKED ISSUE #618: The thickness of border looks like to be ignored on some annotations when events should be triggered
When a box, ellipse, label of a line or polygon annotations are set with a `borderWidth` > 1, the events are fired on the annotations only when the events are inside without considering the thickness of the `borderWidth`.

stockiNail:
Fix #618

This PR enables the use of the `borderWidth` in the `inRange` method for the events.

### TODO

- [x] box
- [x] ellipse
- [x] point
- [x] polygon
- [x] label
- [x] label of the line


kurkle (commented):
The changes look good to me, but is the testing a bit extensive and still not complete?

I think we should also test one pixel in the other direction not triggering the hooks.
So I'd look into creating different simpler test function for the extends, maybe using enter and leave hooks and 2 points, first move to "enter" point, expecting the hook to be called and then move "leave" point and expect "leave" hook to be called. what do you think?

stockiNail:
I think it make sense because we can test not only the positive triggering.
At the end, the proposal is to have 1 function for each test case, maybe with more parameters.
Or 2 function, 1 for positive and 1 for negative. What do you prefer?

stockiNail:
maybe the functions could be:

1. testClickEvents
2. testMoveEvents

with a positive or negative check (by param).

If you agree, I'd wait for #614 approval/merge because there some common tests which can be moved to `events.spec.js`, created in the other PR.

stockiNail:
I'm implementing 3 new functions, 1 for enter, 1 for leave and 1 for click (with dblclick).
The other couple of tests (more generic) will be move to a common event spec when the other PR will be merged

stockiNail:
@kurkle I have committed a temporary implementation in order to, if you have time, you could have a look to event.js and box.spec.js and ellipse.spec.js, if I'm going to the right direction for you.

src/types/point.js
```diff
@@ -6,7 +6,7 @@ export default class PointAnnotation extends Element {
 
   inRange(mouseX, mouseY, useFinalPosition) {
     const {width} = this.getProps(['width'], useFinalPosition);
-    return inPointRange({x: mouseX, y: mouseY}, this.getCenterPoint(useFinalPosition), width / 2 + this.options.borderWidth);
+    return inPointRange({x: mouseX, y: mouseY}, this.getCenterPoint(useFinalPosition), width / 2, this.options.borderWidth);
```
kurkle, line 9: I think this one already worked as expected

src/helpers/helpers.core.js
```diff
@@ -7,18 +7,20 @@ export function clampAll(obj, from, to) {
   return obj;
 }
 
-export function inPointRange(point, center, radius) {
+export function inPointRange(point, center, radius, borderWidth) {
```
kurkle, line 10: I don't think we should move the borderWidth in this helper, making it more complex than it needs to be. We have a point and we test if that is in range from the center using radius. Radius should be the full radius tested.

src/types/point.js
stockiNail: the use case is the radius is 0 and `borderWidth` > 0. the `isPointRange` could return true with the previous implementation, instead, with radius = 0 should return false.

src/helpers/helpers.core.js
stockiNail: the use case is the radius is 0 and borderWidth > 0. the isPointRange could return true with the previous implementation, instead, with radius = 0 should return false.

stockiNail: should be `false` because the annotation is not drawn

stockiNail: https://codepen.io/stockinail/pen/ExwGBZX

Click at x:5, y:5 and a popup will appear. Maybe this is a desired behavior, I don't know, but I think if the annotation is not visible, it shouldn't fire any event.

kurkle: it has display: true doesn't it? I'm not sure why we need multiple ways to hide the annotations while disabling interaction with them too.

Maybe the drawing is actually wrong? you can draw a border around a point that has zero radius.

kurkle: Looks like borderWidth is actually applied to radius? https://codepen.io/kurkle/pen/rNGoEEp

kurkle: As the drawing code is from chart.js lets live with that. So it does not draw when radius is 0.
I think you are correct that its probably unexpected that an invisible point reacts to click. But what else would be the purpose of a point that is not drawn and has `display: true`?

stockiNail: I fully agree with you. Anyway also in CCS, there are always 2 attributes: enabled to enable the componenet and visible, to show or not the component.
I think display = enabled and visible, in the plugin, is something calculated automatically (borderWidth for line, radius for point and polygons, content for label)

stockiNail: > But what else would be the purpose of a point that is not drawn and has `display: true`?

I think display : true declares the willing of the user to draw the annotation but it does not mean it will. 
I don't have any specific use case, anyway, if you use a callback and some specific logic, it could be that will happen, I guess.

stockiNail: Anyway I'm available to revert it and to take time to think about that.

kurkle: Its consistent with the other methods this way.

test/specs/box.spec.js
```diff
@@ -1,7 +1,25 @@
 describe('Box annotation', function() {
   describe('auto', jasmine.fixtures('box'));
 
-  window.testEvents({
+  // event point callbacks
+  const top = function(xScale, yScale, element, xAdjust, yAdjust) {
+    const opts = element.options;
+    return {x: xScale.getPixelForValue(opts.xMin) + xAdjust, y: yScale.getPixelForValue(opts.yMax) + yAdjust};
```
kurkle, line 7: Why not use the element properties instead?

test/specs/ellipse.spec.js
```diff
@@ -1,7 +1,21 @@
 describe('Ellipse annotation', function() {
   describe('auto', jasmine.fixtures('ellipse'));
 
-  const options = {
+  // event point callbacks
+  const top = function(xScale, yScale, element, xAdjust, yAdjust) {
+    return {x: element.x + element.width / 2 + xAdjust, y: element.y + yAdjust};
```
kurkle, line 6: looks like x/yAdjust are applied in wrong place to start with, if we need to consider those everywhere.

test/specs/events.spec.js
```diff
@@ -1,56 +1,224 @@
...
-            },
-            {
-              type: 'line',
-              scaleID: 'y',
-              value: 3
-            }
-          ]
+      }
+    };
+    it('should detect listened and moveListened state set to true', function() {
```
kurkle, line 33: This was labelled "should not throw any exception" before, but it was not testing that.
Does removing this test reduce coverage? Because we should not write tests for internal state.
Tests for internal state end up being maintenance burden, not improving quality at all.

test/specs/events.spec.js
```diff
@@ -1,56 +1,224 @@
...
-  it('should not throw any exception', function() {
+    const pluginOpts = chartConfig.options.plugins.annotation;
 
-    const chart = window.acquireChart(chartConfig);
-    const Annotation = window['chartjs-plugin-annotation'];
-    const state = Annotation._getState(chart);
+    [pluginOpts, options].forEach(function(targetOptions) {
+      it('should not detect any move events (because no callback is set)', function(done) {
+        const enterSpy = jasmine.createSpy('enter');
+        const leaveSpy = jasmine.createSpy('leave');
```
kurkle, line 90: These spies are not attached to anything, and thus will never be called?
You can not test if a callback is called when it does not exist :)

I think there is no good way of testing this, and thus this test can be removed.

test/specs/events.spec.js
```diff
@@ -1,56 +1,224 @@
...
+          window.triggerMouseEvent(chart, 'mousemove', x0y0);
+
+          window.afterEvent(chart, 'mousemove', function() {
+            expect(leaveSpy.calls.count()).toBe(0);
+            done();
+          });
+        });
+      });
+
+      it('should not detect any move events (because only click is set)', function(done) {
```
kurkle, line 108: This test seems similar, a spy that is not attached, does not really test anything (except maybe some of jasmine's code)

test/specs/events.spec.js
```diff
@@ -1,56 +1,224 @@
...
+          window.afterEvent(chart, 'mousemove', function() {
+            expect(leaveSpy.calls.count()).toBe(1);
+            delete pluginOpts.enter;
+            delete pluginOpts.leave;
+            done();
+          });
+        });
+      });
+
+      it('should detect a property in the context, to check persistency', function(done) {
```
kurkle, line 173: This (old) description seems misleading, its not detecting any properties. In my opinion, a better description would be "it should persist properties set in context"

test/specs/events.spec.js
```diff
@@ -1,56 +1,224 @@
...
+
+        const chart = window.acquireChart(chartConfig);
+        const eventPoint = center(chart);
+
+        let dblClick = false;
+        window.afterEvent(chart, 'click', function() {
+          if (!dblClick) {
+            dblClick = true;
+            setTimeout(() => {
+              window.triggerMouseEvent(chart, 'click', eventPoint);
```
kurkle, line 211: I'm not sure why this completes, because `afterEvent` only listens for one event. Anyway, its been like for a long time, so lets leave it as is for now.

kurkle (commented):
I still think the testing is too extensive. Current master:
TOTAL: 1316 SUCCESS  = 96% coverage

This PR:
TOTAL: 2952 SUCCESS = 96% coverage (+0.2% to be precise)

Adding 818 tests for 0.2% more coverage to existing 658 tests that already cover 96%. (each run 4 times, in linux/windows/chrome/firefox)

So lets think about this for a minute:
Each event is essentially using `inRange` to determine if it hits the element.
=> We should be testing inRange to see if it works with borders and things. This will still be 8 or more tests for each element type. 6 types, so 48 tests for each element. So those should not be more than 100 tests total (considering we have labels and stuff that need to be tested separately).

Whe we are sure `inRange` works correctly, then we can just do one test per event per element (+ features, like label).  4 events, 6 elements, so no more than 50 tests for those.

That should give the same coverage, and keep DX (developer experience) adequate (its already running tests too long for my liking).

From the 658 current tests, 383 are already for events.

Anyway, the bordeWidth changes to the code looks fine, I just don't want to merge these tests.

test/specs/events.spec.js
stockiNail: OKI!

test/specs/box.spec.js
stockiNail: I wanted to use the "methods" of chartjs to get the coordinates but I think we can use the element properties

test/specs/events.spec.js
stockiNail: This test has been added to test an event type not managed. We can remove but a part of the code is not tested (when we test the event type).

https://github.com/chartjs/chartjs-plugin-annotation/blob/d9996081e8e7bc896dccdbb0e811fa237a28b7d8/src/events.js#L46-L59

I can remove it but the coverall will provide a warning for missing tests.

test/specs/events.spec.js
stockiNail: OK

test/specs/events.spec.js
stockiNail: uhm... I didn't go in deep of afterEvent but I have the feeling that is catching more than 1 event. Maybe I'm wrong

test/specs/events.spec.js
stockiNail: This test case is testing that only click are set and test what is happening when a move event is coming (all `else` cases). 
I can remove it but I'm going to check which part of the code will remain uncovered by tests

test/specs/ellipse.spec.js
stockiNail: no clear what you meant.

Being the top, I move the point in the middle of the width (for x) and at the same y.

kurkle: Actually I thought xAdjust and yAdjust were the options by that name, but are actually just parameters so my thoguht was wrong.

test/specs/ellipse.spec.js
```diff
@@ -11,16 +25,111 @@ describe('Ellipse annotation', function() {
     xMax: 4,
     yMax: 4,
     borderWidth: 0
-  };
-
-  window.testEvents(options);
+  }, '(without border)');
 
-  options.rotation = 90;
+  window.testEventsOnBorder({
```
kurkle, line 30: so I think this, and the `catch*` and `notCatch*` could be converted to test inRange instead of events.

stockiNail:
> Whe we are sure `inRange` works correctly, then we can just do one test per event per element (+ features, like label). 4 events, 6 elements, so no more than 50 tests for those.

@kurkle I agree to reduce them. I fact I have seen they are too much.

But I'm a little bit confused because this was at the beginning of this PR.

stockiNail:
Just for your info, the minimum amount of events test could not be 50 but 100.

Annotation type: 6
Events type: 4
catch/not catch cases: 2
callbacks (common and for each annotation type): 2
Total: 96

test/specs/ellipse.spec.js
stockiNail: to test inRange and to test an event sounds to me the same thing. Testing an event I can also reuse the same code, otherwise specific functions are needed.

kurkle: I created an example of what I mean: https://github.com/chartjs/chartjs-plugin-annotation/pull/632

stockiNail:
I don't know why GHA is not scheduled...

stockiNail:
@kurkle I would suggest to have another PR for `inRange` test cases, if you agree

test/specs/ellipse.spec.js
```diff
@@ -2,12 +2,19 @@ describe('Ellipse annotation', function() {
...
+  };
+  const eventOut = function(xScale, yScale, element) {
+    const options = element.options;
+    const adjust = options.borderWidth / 2 + 1;
+    return {x: element.x + element.width / 2, y: element.y - adjust};
   };
 
   window.testEvents({
-    type: 'ellipse',
+    type: 'box',
```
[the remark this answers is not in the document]
stockiNail: :( I'm changing it

stockiNail:
> @kurkle I would suggest to have another PR for `inRange` test cases, if you agree

I'm going to add here, in this PR

kurkle:
> > @kurkle I would suggest to have another PR for `inRange` test cases, if you agree
> 
> I'm going to add here, in this PR

I'm not sure why. Did I explain why small changes are good?

stockiNail:
> I'm not sure why. Did I explain why small changes are good?

Yes, you did! OK, no problem. I'm still fixing a test case on polygons

kurkle (approved):
I have some nits about this still, but I'll create a pr for the nits later, because this is going to take forever if I comment on everything.

stockiNail:
> I have some nits about this still, but I'll create a pr for the nits later, because this is going to take forever if I comment on everything.

ok

test/specs/box.spec.js
```diff
@@ -1,14 +1,93 @@
...
     xMin: 2,
     yMin: 2,
     xMax: 4,
     yMax: 4,
-    borderWidth: 0
-  });
+    borderWidth: 10
+  }, eventIn, eventOut);
+
+  describe('events, removing borderWidth by callback, ', function() {
```
kurkle, line 25: @stockiNail what are these use cases, and why should we support these?

stockiNail: Someone can remove the option from `element.options` by the event callback.
This is it has been suggested by #589, where the user can change the element options therefore er are not sure that the borderWidth can have a consistent value. In fact, this is tested and use 0 if borderWidth / 2 is NaN

stockiNail: This is also true for other options which can be changed at runtime without passing thru the `update` and therefore the code can be broken for that.

kurkle: User can delete everything in the element, but we need to consider supported use cases vs stupid users :)

kurkle: for example, I'd not support something like `delete element.inRange`. And deleting borderWidth goes in same category. It'd be ok to set it to a new value, like `0`, but why delete?

stockiNail: > for example, I'd not support something like `delete element.inRange`. And deleting borderWidth goes in same category. It'd be ok to set it to a new value, like `0`, but why delete?

If someone is not familiar with fallback, it could think that remove it the default will be used.

stockiNail: > User can delete everything in the element, but we need to consider supported use cases vs stupid users :)

I agree but the users are users... that's my opinion. Usually I try to do something doesn't get broken, whoever is the user and use case. In the other project, this approach reduces the amount of issues (usually).

stockiNail: @kurkle anyway I think we can remove the test cases and also the check in the code. Let me know.

PR #622, opened 2022-01-10
Add test cases to test one or both missing scales

stockiNail:
This PR is adding test cases.
The goal is to test the annotations behaviors when one or both scales are missing.


test/fixtures/label/missingYScale.js
```diff
@@ -0,0 +1,47 @@
...
+              backgroundColor: 'rgba(33, 101, 171, 0.5)',
+              borderColor: 'rgb(33, 101, 171)',
+              borderWidth: 1,
+              content: ['Missing y scale and', 'then is located to vertical center'],
+              position: {
+                x: 'center',
+                y: 'center'
+              },
+              xAdjust: 0,
+              yAdjust: -100,
```
LeeLenaleee, line 34: Why adjust the Y, makes it more confusing too look at since if you look at the image you would expect the text to be centered

test/fixtures/label/missingBothScales.js
```diff
@@ -0,0 +1,47 @@
...
+              backgroundColor: 'rgba(33, 101, 171, 0.5)',
+              borderColor: 'rgb(33, 101, 171)',
+              borderWidth: 1,
+              content: ['Missing x and y scales and', 'then is located in the center of the chart'],
+              position: {
+                x: 'center',
+                y: 'center'
+              },
+              xAdjust: 0,
+              yAdjust: -100,
```
LeeLenaleee, line 34: Why adjust the Y, makes it more confusing too look at since if you look at the image you would expect the text to be centered

test/fixtures/label/missingXScale.js
```diff
@@ -0,0 +1,47 @@
...
+              backgroundColor: 'rgba(33, 101, 171, 0.5)',
+              borderColor: 'rgb(33, 101, 171)',
+              borderWidth: 1,
+              content: ['Missing x scale and', 'then is located to horizontal center'],
+              position: {
+                x: 'center',
+                y: 'center'
+              },
+              xAdjust: 0,
+              yAdjust: -100,
```
LeeLenaleee, line 34: Why adjust the Y, makes it more confusing too look at since if you look at the image you would expect the text to be centered

test/fixtures/label/missingYScale.js
stockiNail: I added to enable the callout. Anyway I agree with you. I'm removing it.

test/fixtures/label/missingBothScales.js
stockiNail: I added to enable the callout. Anyway I agree with you. I'm removing it.

test/fixtures/label/missingXScale.js
stockiNail: I added to enable the callout. Anyway I agree with you. I'm removing it.

test/fixtures/label/missingYScale.js
stockiNail: done!

LeeLenaleee (approved):

test/fixtures/label/missingBothScales.js
stockiNail: done!

stockiNail:
@LeeLenaleee it seems there is a issue with CC: https://docs.codeclimate.com/docs/fetch-failed. Could be the integration between GH and CC?

LeeLenaleee:
Don't know but since its only test's that have been added in this PR I wouldn't worry about it

PR #626, opened 2022-01-13
Move label drawTime default to line annotation defaults

### LINKED ISSUE #625: Annotations are not drawn if the id is `animations` or `label`
If you set an annotation with id is equals `animations` or `label`, the annotations is not drawn on the chart.

Config example:

```javascript
annotations: {
  label: {
    type: 'point',
    xScaleID: 'x',
    yScaleID: 'y',
    xValue: 5,
    yValue: 5
  }
}
```
This seems to be related to the fallback because `animations` and `label` are 2 properties of annotations defaults and are objects.

Codepen: https://codepen.io/stockinail/pen/JjrmRBR

EDIT: the annotations are not drawn because `drawTime` is undefined.

stockiNail:
Fix #625

Not sure why the label `drawTime` was present in the annotations node. Maybe there are other reasons.

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -111,10 +111,7 @@ export default {
     },
     clip: true,
     dblClickSpeed: 350, // ms
-    drawTime: 'afterDatasetsDraw',
-    label: {
-      drawTime: null
-    }
+    drawTime: 'afterDatasetsDraw'
   },
 
   descriptors: {
--- a/src/types/line.js
+++ b/src/types/line.js
@@ -213,7 +213,7 @@ LineAnnotation.defaults = {
     color: '#fff',
     content: null,
     cornerRadius: undefined, // TODO: v2 remove support for cornerRadius
-    drawTime: undefined,
+    drawTime: null,
     enabled: false,
     font: {
       family: undefined,


kurkle:
The plugin options act as defaults for all annotation types.

https://github.com/chartjs/chartjs-plugin-annotation/blob/2167375ac06a04b3ddf2b26ea0e9234f024ae4f0/src/types/index.js#L27-L39

I think it now resolves `options.plugins.annotation.label` and finds a defined (null) drawTime there. 

https://codepen.io/kurkle/pen/PoJXboE

kurkle:
> Not sure why the label drawTime was present in the annotations node. Maybe there are other reasons.

It was put there in #345

stockiNail:
> I think it now resolves `options.plugins.annotation.label` and finds a defined (null) drawTime there.

I'll have a look. Let me switch this PR in draft.

stockiNail:
I had a look and I have to admit my lack of knowledge about fallback/proxies.

I was expecting that the fallback will ignore the name of the annotation because to the options are at another level.

```javascript
plugins.annotation.annotations.myId = myOptions
```

I'm expecting that the fallback will act on `myOptions` (and and not `myId.myOptions`) to `plugins.annotation`.

kurkle:
> I'm expecting that the fallback will act on myOptions (and and not myId.myOptions) to plugins.annotation.

image

0. options.annotation.annotations.label
1. options.annotation.annotations.label.label (includes defaults, but it should not)
2. options.annotation.label (this does not exist, but it resolves to defaults. it should not)
3. defaults.elements.lineAnnotation
4. lineAnnotation descriptors
5. options.annotation
6. defaults.plugins.annotation
7. annotation plugin descriptor

stockiNail:
> I think the actual issue is in `beforeUpdate`, we are storing a resolver in the state, when we should only store the options.
> 
> Lets keep this open, but I'll try to create a fix this weekend.

**THANK YOU** !!! I'm looking forward for that in order to understand better.

stockiNail:
> 1. options.annotation.annotations.label
> 1. defaults.elements.lineAnnotation.label (this should not be here, its a point annotation)
> 1. defaults.plugins.annotation.label (this is the culprit, _fallback = 'plugins.annotation' and the current key 'label' is looked in there.
> 1. defaults.elements.pointAnnotation
> 1. lineAnnotation descriptors
> 1. options.annotation
> 1. defaults.plugins.annotation
> 1. annotation plugin descriptor

@kurkle I was thinking about this fallback and your closed PR #601. 
But, and maybe it's not a good idea, if we will move all common options (and their defaults) at `annotation plugin descriptor`?
Shouldn't we reach the goal to have common options (and defaults) only in a place?
If I understood well, we should anyway to maintain the keys (set to undefined) in the defaults options of each annotation.

PR #631, opened 2022-01-16
Remove (some) reduntant properties from fixtures

kurkle:
Some chores in fixtures


test/fixtures/line/scriptableOptions.js
```diff
@@ -16,8 +16,10 @@ module.exports = {
       },
       plugins: {
         annotation: {
-          drawTime() {
-            return 'afterDatasetsDraw';
+          common: {
+            drawTime() {
+              return 'afterDatasetsDraw';
+            }
```
LeeLenaleee, line 22: Same as above

test/fixtures/line/labelsScriptableOptions.js
```diff
@@ -16,8 +16,10 @@ module.exports = {
       },
       plugins: {
         annotation: {
-          drawTime() {
-            return 'afterDatasetsDraw';
+          common: {
+            drawTime() {
+              return 'afterDatasetsDraw';
+            }
```
LeeLenaleee, line 22: Isn't the old syntax correct, because when I try that it works and with common it doesnt. Also cant find any reference to common in the source code

LeeLenaleee (commented):
Only 1 thing about scriptable draw time that seems incorrect to me, but can be wrong there

**EDIT:**
Just noticed the open PR to add this, so looks fine to me after #630 has been merged

test/fixtures/line/scriptableOptions.js
kurkle: Yes, took this off from the other PR and forgot to update these.

test/fixtures/line/labelsScriptableOptions.js
kurkle: fixed, will add the changes to the other pr, which is a breaking one.

LeeLenaleee (approved):

PR #632, opened 2022-01-17
Box: change event tests to inRange tests

kurkle:
Essentially test the thing directly, instead of creating a chart and firing events and stuff. Coverage is the same, but some lines get less throughput:

image

### DIFF
--- a/test/specs/box.spec.js
+++ b/test/specs/box.spec.js
@@ -1,6 +1,44 @@
 describe('Box annotation', function() {
   describe('auto', jasmine.fixtures('box'));
 
+  describe('inRange', function() {
+    const annotation = {
+      type: 'box',
+      xScaleID: 'x',
+      yScaleID: 'y',
+      xMin: 2,
+      yMin: 4,
+      xMax: 8,
+      yMax: 6,
+      borderWidth: 0,
+      rotation: 0
+    };
+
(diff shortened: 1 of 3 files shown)


stockiNail:
may I create another PR with inRange test cases for all, getting this as sample?

kurkle:
> may I create another PR with inRange test cases for all, getting this as sample?

Not quite sure if everything should be combined in a big PR.

test/specs/box.spec.js
```diff
@@ -1,14 +1,39 @@
...
+      type: 'box',
+      xMin: 2,
+      yMin: 4,
+      xMax: 8,
+      yMax: 6,
+      borderWidth: 0,
+      rotation: 0
+    };
+
+    const chart = window.scatter10x10({box: annotation});
```
stockiNail, line 15: Is not better to pass only the annotation options and use in the scatter10x10 chart config the annotations as an array?

kurkle: I thought of leaving the choice of array vs object and how many annotations to the caller.

stockiNail (approved):

stockiNail:
@kurkle rebase needed.

stockiNail (approved):

PR #634, opened 2022-01-17
Box: test borderWidths. Ellipse: change events tests to inRange tests

kurkle:


test/specs/box.spec.js
```diff
@@ -12,117 +12,37 @@ describe('Box annotation', function() {
       rotation: 0
     };
 
-    const chart = window.scatter10x10({box: annotation});
+    const chart = window.scatter10x10({test: annotation});
     const element = window['chartjs-plugin-annotation']._getState(chart).elements[0];
```
stockiNail, line 16: just thinking loud, could make sense to have a public function for that (there is a similar function in test/events.js).

https://github.com/chartjs/chartjs-plugin-annotation/blob/52adba72188e6474db7e4829424c56beabdf03e5/test/events.js#L156-L160

stockiNail (approved):

PR #635, opened 2022-01-17
Fix auto callout position calculation in label annotation

### LINKED ISSUE #636: Invalid position calculation of callout when set to auto
When the callout of label annotation is set to top or bottom, the position of callout is wrong.

stockiNail:
Fix #636


src/types/label.js
```diff
@@ -131,21 +131,23 @@ function drawCallout(ctx, element) {
...
-  const {sideStart, sideEnd} = getCalloutSideCoord(element, calloutPosition, separatorStart);
+
   ctx.save();
   ctx.beginPath();
   const stroke = setBorderStyle(ctx, callout);
+  if (!stroke) {
+    return ctx.restore();
+  }
+  const {separatorStart, separatorEnd} = getCalloutSeparatorCoord(element, calloutPosition);
+  const {sideStart, sideEnd} = getCalloutSideCoord(element, calloutPosition, separatorStart);
```
kurkle, line 142: Are these changes somehow related to the bug, or just an optimization?

One less line and maybe one point less complex in CC without the return (`if (stroke) { ...draw... }`). Though its going to nest one level more, so the complexity might be the same.

test/fixtures/label/calloutHorizontalPosition.js
```diff
@@ -58,6 +57,99 @@ module.exports = {
               type: 'point',
               xValue: 'May',
               yValue: 10
+            },
+            text3: {
```
kurkle, line 61: Wondering why in this fixture we don't have `point1`? Can you add that to be consistent

stockiNail: I'm wondering as well because I have changed yesterday and I didn't recognize it...
I'm pushing it.

stockiNail: done

kurkle (approved):

PR #639, opened 2022-01-18
Remove old test cases

stockiNail:
This PR removes old and useless test cases


kurkle (approved):
Looks like label with no content was tested only here, but that one should be added to a fixture.

LeeLenaleee (approved):

PR #640, opened 2022-01-18
Label fixtures: remove point tests, update padding

kurkle:
Originally label annotation had point feature, but not those points are point annotations.
So makes no sense to test point annotation styling in label annotation fixtures.

Updated the padding fixture to only test padding, positioning is tested in other fixtures.


LeeLenaleee (dismissed):

kurkle:
Edit, moved the stringifying thing to window, will use it in some other tests too.

LeeLenaleee (approved):

PR #331, opened 2021-02-03
Chart.js beta.13 compatibility

kurkle:
Closes: #265

[(outdated) demo](https://codepen.io/kurkle/pen/rNWOdNG)

PR #642, opened 2022-01-18
Use resolver proxy as element options

### LINKED ISSUE #621: Background color context element does not have position values
Reproduction: https://codesandbox.io/s/chart-js-gradient-annotations-broken-8pf3z

It seems that on the first draw, `ctx.element` does not have `x`, `x2`, `y` nor `y2`, making it impossible to draw gradients.

These values are only missing on the first draw; You can uncomment line 52 and see that on the first draw `green` is returned, but after clicking "Force update", the actual gradient works.

kurkle:
Fix #621

Todo:
- [x] check why it was originally changed to resolved options
- [ ] resolve caching issues

Edit: it was changed in #370
I'm not able to reproduce the error how its described, but I see problems with caching in the samples.


stockiNail:
I think it was implemented here: https://github.com/chartjs/chartjs-plugin-annotation/pull/331

kurkle:
Ok, the problem is with animations. Animations update the options, and when options is a proxy, the actual options get updated.

I don't think this is the way to go.

PR #643, opened 2022-01-19
Label fixtures: merge position tests in vertical and horizontal ones

stockiNail:
The percentage position test cases are consolidated in vertical and horizontal ones.
The test case with position and points are removed.


kurkle (dismissed):
Nice!

Could have ='100%' and ='invalid' in both fixtures for completeness.

kurkle:
Another note, we have some nice looking annotations in tests, maybe some of those should be converted to samples. In tests, the contrast is more important than looks :)

stockiNail:
> we have some nice looking annotations in tests

Which ones are you referring to in particular?

kurkle:
Maybe some of those we are not needing in tests, like label with call-out +point combination. And actually these new fixtures here would be good samples too.

stockiNail:
> Could have ='100%' and ='invalid' in both fixtures for completeness.

added! ;)

kurkle (approved):

stockiNail:
> Maybe some of those we are not needing in tests, like label with call-out +point combination. And actually these new fixtures here would be good samples too.

I'll have a look following your suggestion and I'll come back with a PR (it also means **PR**oposal in this case).

stockiNail:
Today the testing on Windows sounds to have some issues (always timeout)

```console
[test-*integration]   1) Integration Tests
[test-*integration]        chartjs-plugin-annotation should compile with all supported TS versions:
[test-*integration]      Error: Timeout of 300000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. 
[test-*integration]      (D:\a\chartjs-plugin-annotation\chartjs-plugin-annotation\test\integration\integration-test.js)
```

PR #647, opened 2022-01-20
Label fixtures: merge canvas test cases and remove useless point test

stockiNail:
This PR merges canvas test cases and removes useless point test from label fixtures


kurkle (commented):
I think scales could be hidden in this one, and then should not need to have additional tolerance

stockiNail:
> I think scales could be hidden in this one, and then should not need to have additional tolerance

done!

kurkle (approved):

PR #648, opened 2022-01-20
Label: change event tests to inRange tests

stockiNail:
This PR remove events test cases with `inRange` method tests.

### DIFF
--- a/test/specs/label.spec.js
+++ b/test/specs/label.spec.js
@@ -1,30 +1,47 @@
 describe('Label annotation', function() {
   describe('auto', jasmine.fixtures('label'));
 
-  const eventIn = function(xScale, yScale, element) {
-    const options = element.options;
-    const adjust = options.borderWidth / 2 - 1;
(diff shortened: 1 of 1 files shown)


stockiNail:
@kurkle where do we have to move the events' end-2-end test cases?

In my opinion the events' end-2-end test cases must be done. We should share:

1. where?
2. for all annotations types or only 1?

kurkle:
I think for all types and all events, but only for center point for example.

stockiNail:
> OK, let me change this PR with event tests for label.

I had a look and I think it's better to have a specific PR to fix tests on events.

kurkle (commented):
I guess the callout is not interactable?

kurkle (approved):

stockiNail:
> I guess the callout is not interactable?

No, it isn't. Be aware that also arrow heads in line annotation are not interactable. Maybe for arrow heads it could be implemented.

PR #649, opened 2022-01-20
Point: change event tests to inRange tests

stockiNail:

### DIFF
--- a/test/specs/ellipse.spec.js
+++ b/test/specs/ellipse.spec.js
@@ -21,7 +21,7 @@ describe('Ellipse annotation', function() {
       const xRadius = element.width / 2;
       const yRadius = element.height / 2;
 
-      it('should return true when point is inside element', function() {
+      it(`should return true when point is inside element\n{x: ${center.x}, y: ${center.y}, xRadius: ${xRadius.toFixed(1)}, yRadius: ${yRadius.toFixed(1)}}`, function() {
         for (const borderWidth of [0, 10]) {
           const halfBorder = borderWidth / 2;
           element.options.borderWidth = borderWidth;
@@ -34,12 +34,12 @@ describe('Ellipse annotation', function() {
               y: center.y + Math.sin(rad) * (yRadius + halfBorder)
             }, center, rotation / 180 * Math.PI);
 
-            expect(element.inRange(x, y)).toEqual(true);
+            expect(element.inRange(x, y)).withContext(`rotation: ${rotation}, angle: ${angle}, borderWidth: ${borderWidth}, {x: ${x.toFixed(1)}, y: ${y.toFixed(1)}}`).toEqual(true);
           }
         }
       });
 
-      it('should return false when point is outside element', function() {
+      it(`should return false when point is outside element\n{x: ${center.x}, y: ${center.y}, xRadius: ${xRadius.toFixed(1)}, yRadius: ${yRadius.toFixed(1)}}`, function() {
         for (const borderWidth of [0, 10]) {
           const halfBorder = borderWidth / 2;
           element.options.borderWidth = borderWidth;
@@ -52,7 +52,7 @@ describe('Ellipse annotation', function() {
               y: center.y + Math.sin(rad) * (yRadius + halfBorder + 1)
             }, center, rotation / 180 * Math.PI);
 
-            expect(element.inRange(x, y)).toEqual(false);
+            expect(element.inRange(x, y)).withContext(`rotation: ${rotation}, angle: ${angle}, borderWidth: ${borderWidth}, {x: ${x.toFixed(1)}, y: ${y.toFixed(1)}}`).toEqual(false);
           }
         }
       });
--- a/test/specs/ellipse.spec.js
+++ b/test/specs/ellipse.spec.js
@@ -21,7 +21,7 @@ describe('Ellipse annotation', function() {
       const xRadius = element.width / 2;
       const yRadius = element.height / 2;
 
-      it(`should return true when point is inside element\n{x: ${center.x}, y: ${center.y}, xRadius: ${xRadius.toFixed(1)}, yRadius: ${yRadius.toFixed(1)}}`, function() {
+      it('should return true when point is inside element', function() {
         for (const borderWidth of [0, 10]) {
           const halfBorder = borderWidth / 2;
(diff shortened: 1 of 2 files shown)


test/specs/point.spec.js
```diff
@@ -1,92 +1,55 @@
...
+    const chart = window.scatter10x10({test: annotation});
+    const element = window.getAnnotationElements(chart)[0];
 
-      pointOpts.click = function(ctx) {
-        if (ctx.element.options.borderWidth) {
-          delete ctx.element.options.borderWidth;
-          ctx.chart.draw();
-        } else {
-          ctx.element.options.click = clickSpy;
+    it('should return true inside element', function() {
```
kurkle, line 15: This could be improved a little, testing 8 points around the point center, so an rectangle would not pass. Same for the "outside" test.

test/specs/point.spec.js
```diff
@@ -1,92 +1,55 @@
...
-            }
-          }
-        }
-      },
+  describe('inRange', function() {
+    const annotation = {
+      type: 'point',
+      xValue: 5,
+      yValue: 5,
+      radius: 30
```
kurkle, line 9: Could test couple of radiuses too. For test performance, its better to have couple of points in the same chart with different radiuses and test each of those.

kurkle (approved):
I had couple of ides how to make this test more strict, but I guess its quite ok without any additional changes.

test/specs/point.spec.js
stockiNail: It's not clear to me what "so an rectangle would not pass." mean. Here we are testing a point (whatever point style will be applied, the rule of inRange is always related to the circle).

test/specs/point.spec.js
stockiNail: I agree. Do you want I will add the tests on 2 points in this PR?

test/specs/point.spec.js
kurkle: image

A (bad) picture is better than 1000 words, they say. :)
Testing the points marked with blue arrows does not detect if the inRange returned true in those gray areas.

test/specs/point.spec.js
stockiNail: Done!

test/specs/point.spec.js
stockiNail: done

stockiNail: I use the same angles as you did for the ellipses. ;)

test/specs/point.spec.js
```diff
@@ -2,54 +2,77 @@ describe('Point annotation', function() {
...
+    const annotation3 = {
+      type: 'point',
+      xValue: 3,
+      yValue: 3,
+      radius: 0
+    };
 
-    const chart = window.scatter10x10({test: annotation});
-    const element = window.getAnnotationElements(chart)[0];
+    const chart = window.scatter10x10({test1: annotation1, test2: annotation2, test3: annotation3});
```
kurkle, line 24:
```suggestion
    const chart = window.scatter10x10({annotation1, annotation2, annotation3});
```

So the the id from error message is directly linked to annotation config rather than needing read the map (even thought numbers match, it still requires more processing from reader)

kurkle (commented):
Couple of comments still, the tests look fine now

test/specs/point.spec.js
```diff
@@ -2,54 +2,77 @@ describe('Point annotation', function() {
...
-        const halfBorder = borderWidth / 2;
-        element.options.borderWidth = borderWidth;
-        for (const x of [element.x - halfBorder, element.x + halfBorder]) {
-          expect(element.inRange(x, element.y)).toEqual(false);
-        }
-        for (const y of [element.y - halfBorder, element.y + halfBorder]) {
-          expect(element.inRange(element.x, y)).toEqual(false);
+    elemsNoRad.forEach(function(element) {
+      it(`should return false radius is 0 element '${element.options.id}'`, function() {
+        element.width = 0;
```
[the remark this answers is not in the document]
stockiNail: it remained there after I changed my impl. I remove it.

test/specs/point.spec.js
```diff
@@ -2,54 +2,77 @@ describe('Point annotation', function() {
...
+          const halfBorder = borderWidth / 2;
+          element.options.borderWidth = borderWidth;
+          const radius = element.height / 2;
+          for (const angle of [0, 45, 90, 135, 180, 225, 270, 315]) {
+            const rad = angle * (Math.PI / 180);
+            const {x, y} = {
+              x: element.x + Math.cos(rad) * (radius + halfBorder + 1),
+              y: element.y + Math.sin(rad) * (radius + halfBorder + 1)
+            };
+            expect(element.inRange(x, y)).withContext(`in, angle: ${angle}, radius: ${radius}, borderWidth: ${borderWidth}, {x: ${x.toFixed(1)}, y: ${y.toFixed(1)}}`).toEqual(false);
```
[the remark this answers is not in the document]
stockiNail: OK!

test/specs/point.spec.js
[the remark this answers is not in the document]
stockiNail: done

test/specs/point.spec.js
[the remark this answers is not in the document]
stockiNail: done (in ellipse too)

test/specs/point.spec.js
stockiNail: done

test/specs/ellipse.spec.js
```diff
@@ -34,12 +34,12 @@ describe('Ellipse annotation', function() {
               y: center.y + Math.sin(rad) * (yRadius + halfBorder)
             }, center, rotation / 180 * Math.PI);
 
-            expect(element.inRange(x, y)).withContext(`in, rotation: ${rotation}, angle: ${angle}, borderWidth: ${borderWidth}, {x: ${x.toFixed(1)}, y: ${y.toFixed(1)}}`).toEqual(true);
```
kurkle, line 37: No, I did not mean removing the whole context, just the `in` or `out` word. Those numbers in an failure message are really helpful when something fails.

stockiNail: done

kurkle (approved):

PR #650, opened 2022-01-21
Move events test cases to events.spec

stockiNail:
This PR is moving the E2E events tests in the `events.spec.js` file


test/events.js
```diff
@@ -1,92 +1,79 @@
...
         const enterSpy = jasmine.createSpy('enter');
         const leaveSpy = jasmine.createSpy('leave');
 
         targetOptions.enter = enterSpy;
         targetOptions.leave = leaveSpy;
         pluginOpts.annotations = [options];
 
-        const chart = window.acquireChart(context.chartConfig);
-        const eventPoint = retrieveEventPoint(chart, getCenterPoint);
+        const chart = window.acquireChart(chartConfig);
```
kurkle, line 45: could use the `scatter10x10` helper

test/events.js
```diff
@@ -1,92 +1,79 @@
...
-function testEnterEvent(options, toBe, getEventPoint) {
-  const context = getTestCaseContext(toBe);
+export const getCenterPoint = (chart) => window.getAnnotationElements(chart)[0].getCenterPoint();
+export const eventPoint0 = {
+  x: 0,
+  y: 0
+};
 
+export function testEvents(options) {
   describe('events', function() {
```
kurkle, line 8: could have the `on on ${options.type}` in this describe statement so the tests would not need it

kurkle (commented):
could improve the speed of testing by creating a chart with each type of annotation and then testing the events on each annotation. it would also allow to test the untested case when moving mouse from annotation to another (triggering leave in the old and enter in the new annotation).

something like `annotations = [{type: 'point'}, {type: 'box'}, ...]`
and then looping those for each test.

this can be done later too, if the change gets too complicated to do directly

stockiNail:
> this can be done later too, if the change gets too complicated to do directly

I want to take time for this PR therefore I'll do in this one, makes a lot of sense!

test/events.js
stockiNail: > could use the `scatter10x10` helper

`scatter10x10` is returning a chart already acquired but we need to manage the plugin configuration (not only the annotation one) because we are testing the events hooks at `annotation` node level.

test/events.js
stockiNail: done

test/events.js
stockiNail: I forgot that `scatter10x10` is accepting only annotations as argument.

stockiNail:
> could improve the speed of testing by creating a chart with each type of annotation and then testing the events on each annotation. it would also allow to test the untested case when moving mouse from annotation to another (triggering leave in the old and enter in the new annotation).
> 
> something like `annotations = [{type: 'point'}, {type: 'box'}, ...]` and then looping those for each test.
> 
> this can be done later too, if the change gets too complicated to do directly

I have spent some time to have a look how this could be implemented but I see more complexity than benefits. I'm continuing find out a possible solution, not too much complex.

stockiNail:
I have tried to use a unique chart with all annotation and perform the tests in this single chart instance.

But there is an issue because if the chart instance is not inside the `it` statement block, the `chart.canvas` is `null` and the chartjs-test-utils throws an exception when you try to trigger an event.

```log
Unhandled promise rejection: TypeError: Cannot read properties of null (reading 'getBoundingClientRect')
TypeError: Cannot read properties of null (reading 'getBoundingClientRect')
    at triggerMouseEvent (node_modules/chartjs-test-utils/dist/chartjs-test-utils.js:673:19 <- test/index.js:488:21)
```

```javascript
describe(`events on ${annotation.type}`, function() {
  const chart = window.acquireChart(chartConfig);
  ...
  it('should detect enter and leave events', function(done) {
    ...
    window.triggerMouseEvent(chart, 'mousemove', eventPoint); // <- exception!
    ...
```

```javascript
describe(`events on ${annotation.type}`, function() {
  ...
  it('should detect enter and leave events', function(done) {
    const chart = window.acquireChart(chartConfig);
    ...
    window.triggerMouseEvent(chart, 'mousemove', eventPoint); // <- works!
    ...
```

I don't know exactly why it behaves in this way but I think it depends on jasmine.

kurkle:
> But there is an issue because if the chart instance is not inside the it statement block, the chart.canvas is null and the chartjs-test-utils throws an exception when you try to trigger an event. 
>
>I don't know exactly why it behaves in this way but I think it depends on jasmine.

We have that kind of tests already. Those should be failing too?

It probably has something to do with the random order of execution of the `it` statements.

stockiNail:
> We have that kind of tests already. Those should be failing too?

We don't have any other test case where `triggerMouseEvent` function is invoked in the same manner.

kurkle:
> > We have that kind of tests already. Those should be failing too?
> 
> We don't have any other test case where `triggerMouseEvent` function is invoked in the same manner.

Meaning its calling `done()` callback when completed?
Could try that as async function instead.

it('blah' async function() {
  await triggerMouse...
  expect(something)
});

stockiNail:
> it('blah', async function() {
> await triggerMouse...
> expect(something)
> });

ERROR: 'DEPRECATION: An asynchronous before/it/after function was defined with the async keyword but also took a done callback. This is not supported and will stop working in the future. Either remove the done callback (recommended) or remove the async keyword. (in spec: Common events on box should detect enter and leave events)'

And the same error:

Unhandled promise rejection: TypeError: Cannot read properties of null (reading 'getBoundingClientRect')
TypeError: Cannot read properties of null (reading 'getBoundingClientRect')
  at triggerMouseEvent (node_modules/chartjs-test-utils/dist/chartjs-test-utils.js:673:19 <- test/index.js:488:21)


I'm trying with before/afterAll.

stockiNail:
Using `beforeAll`:

1. moving the invocation of the tests from `beforeAll` (instead of from `events.spec.js`), there is a bug because `describe` mist be inside to a `describe` function.
2. creating a chart for events and storing it in the window, the canvas is anyway `null` inside the `it` statement block.

stockiNail:
@kurkle I have tried all the day without success. Any idea? In my opinion we should stay as is and then try to improve in next future.
But I leave the decision to you.

kurkle (approved):

PR #651, opened 2022-01-21
Polygon: change event tests to inRange tests

stockiNail:

it also adds a further check in `inRange` of polygon because if the radius is set to zero, the point (element.x, element.y) is accepted.

The approach for polygon is a little bit different comparing with point because the polygon is calculating the point elements (used in `inRange`) at initialization of the element.

### DIFF
--- a/test/specs/polygon.spec.js
+++ b/test/specs/polygon.spec.js
@@ -1,91 +1,94 @@
 describe('Polygon annotation', function() {
   describe('auto', jasmine.fixtures('polygon'));
 
-  const eventIn = function(xScale, yScale, element) {
-    const options = element.options;
-    const adjust = options.borderWidth / 2 - 1;
-    return {x: element.x, y: element.y - element.height / 2 - adjust};
-  };
-  const eventOut = function(xScale, yScale, element) {
-    const options = element.options;
-    const adjust = options.borderWidth / 2 + 1;
-    return {x: element.x, y: element.y - element.height / 2 - adjust};
-  };
+  describe('inRange', function() {
+    const annotation1 = {
+      type: 'polygon',
+      xValue: 7,
+      yValue: 7,
+      radius: 30
+    };
+    const annotation2 = {
+      type: 'polygon',
+      xValue: 3,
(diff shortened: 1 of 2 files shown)


test/specs/polygon.spec.js
```diff
@@ -1,91 +1,125 @@
...
-          ctx.element.options.click = clickSpy;
+      it(`should return true inside element '${element.options.id}'`, function() {
+        const halfBorder = borderWidth / 2;
+        const vertices = [];
+        let rad = rotation * (Math.PI / 180);
+        for (let i = 0; i < sides; i++, rad += angle) {
+          const sin = Math.sin(rad);
+          const cos = Math.cos(rad);
+          vertices.push({
+            x: center.x + sin * (radius + halfBorder - 1),
```
kurkle, line 92: is the  -1 actually needed?

test/specs/polygon.spec.js
```diff
@@ -1,91 +1,125 @@
...
-        } else {
-          ctx.element.options.click = clickSpy;
+      it(`should return true inside element '${element.options.id}'`, function() {
+        const halfBorder = borderWidth / 2;
+        const vertices = [];
+        let rad = rotation * (Math.PI / 180);
+        for (let i = 0; i < sides; i++, rad += angle) {
+          const sin = Math.sin(rad);
+          const cos = Math.cos(rad);
+          vertices.push({
```
kurkle, line 91: no real need to store the vertices in an array and loop that array right after, can just have the `expect` in this first loop, right?

test/specs/polygon.spec.js
stockiNail: I think yes, because removing them, the test case failed. With -1 you are sure to be inside and it works

test/specs/polygon.spec.js
stockiNail: I stored because I wasn't sure if you want to test also some points in "gray" area. Anyway I remove them

kurkle (approved):

PR #653, opened 2022-01-23
Directly assign new element resolved properties

### LINKED ISSUE #621: Background color context element does not have position values

kurkle:
Fix #621 

This fixes the issue by directly assigning the `resolveElementProperties` result to a newly created annotation (detected by undefined `x` property). So element has its position values set before styling options are resolved.

This also changes what is passed to `resolveElementProperties`, its now the resolved, instead of resolved options. So any options needed by the `resolveElementProperties` are still resolved before the element has position values.

Edit, using a resolver for resolveElementProperties removes the need to have all of those options in element defaults.


kurkle:
Note, the defaults removed from element defaults will not end up in `element.options` anymore. That one could be considered a breaking change.

So in scriptable context the `ctx.element.options.xScaleID` would not be defined anymore (because its not listed in defaults).

So maybe the defaults cleanup needs to be reverted and done in v2 instead.

stockiNail (commented):
AFA I understood, only the options to locate the annotations are removed

src/types/label.js
```diff
@@ -84,24 +84,12 @@ LabelAnnotation.defaults = {
     style: undefined,
     weight: undefined
   },
-  height: undefined,
```
stockiNail, line 87: Why did you remove `height` and `width`?

src/types/label.js
```diff
@@ -84,24 +84,12 @@ LabelAnnotation.defaults = {
...
   padding: 6,
   position: 'center',
   shadowBlur: 0,
   shadowOffsetX: 0,
   shadowOffsetY: 0,
   textAlign: 'center',
-  width: undefined,
-  xAdjust: 0,
```
stockiNail, line 95: The same for x/yAdjust. Why?

src/types/line.js
```diff
@@ -273,17 +273,9 @@ LineAnnotation.defaults = {
     yAdjust: 0,
     yPadding: undefined, // TODO: v2 remove support for yPadding
   },
-  scaleID: undefined,
   shadowBlur: 0,
   shadowOffsetX: 0,
   shadowOffsetY: 0,
-  value: undefined,
```
stockiNail, line 280: I think `endValue` is still in the line defaults. Should be removed as did for `value`.

stockiNail:
@kurkle I have another doubt. What does it happen if a styled options (like `backgroundColor`) is called in `resolveElementProperties`? I think the element is not completed, I guess. Or am I wrong?
If I'm not wrong, maybe we should move all logic where these styled options are called from `resolveElementProperties` to `draw` method. In these case should we take care about possible performance issue?

src/types/label.js
kurkle: Aren't those used in `resolveElementProperties` only?

stockiNail:
> @kurkle I have another doubt. What does it happen if a styled options (like `backgroundColor`) is called in `resolveElementProperties`? I think the element is not completed, I guess. Or am I wrong?
> If I'm not wrong, maybe we should move all logic where these styled options are called from `resolveElementProperties` to `draw` method. In these case should we take care about possible performance issue?

Just an example where backgroundColor is invoked in `resolveElementProperties`:

https://github.com/chartjs/chartjs-plugin-annotation/blob/a311f5dfd871907be4972bacfe00351042dda570/src/types/label.js#L28-L35

src/types/label.js
stockiNail: > Aren't those used in `resolveElementProperties` only?

Well, they are, to calculate the size of the label when an image or canvas is used, but together many others (i.e. font, padding), when the content is a text. Do we have to remove all options which are affecting the annotation dimension?

kurkle:
> If I'm not wrong, maybe we should move all logic where these styled options are called from resolveElementProperties to draw method. In these case should we take care about possible performance issue?

Moving to draw would disable animating the style options (colors) and probably degrease performance.

> Just an example where backgroundColor is invoked in resolveElementProperties

I have no idea what is the purpose of that check. Why do we need to know if the box is visible and not transparent?
Removing the backgroundColor stuff there does not make anything fail.

src/types/label.js
kurkle: We don't have to remove any of them, but we can remove the ones that do not have a value.

src/types/label.js
kurkle: The xAdjust should not be removed. Not sure why any tests did not fail.

kurkle:
I've reverted the removal of defaults, because it would be breaking change.

stockiNail:
If you agree, I think we should have a best practice to remove any use of styling options from the `resolveElementProperties`.
Make sense?

test/specs/box.spec.js
```diff
@@ -45,4 +45,27 @@ describe('Box annotation', function() {
       }
     });
   });
+
+  describe('scriptable options', function() {
```
stockiNail, line 49: Could make sense to have a fixture for box and ellipse where we set a gradient as background color?

kurkle: > > I have no idea what is the purpose of that check. Why do we need to know if the box is visible and not transparent?
> > Removing the backgroundColor stuff there does not make anything fail.
> 
> This specific check was done to avoid to draw the box if border width = 0 and background is transparent. But we can remove it. drawBox function will be called anyway even if it shouldn't draw anything.

Yes, but that is again a breaking change, so lets do that in v2

kurkle: > Could make sense to have a fixture for box and ellipse where we set a gradient as background color?

Added a fixture

stockiNail (approved):

Issue #532, opened 2021-12-05
Normalize all annotation to be based on box model (x, y, width, height)


kurkle:
For consistency and thus easier handling of annotations regardless of type. 

Could be used for rough collision detection for example.

stockiNail:
Proposal, maybe can help:

![elementProps](https://user-images.githubusercontent.com/11741250/145609678-792aac72-2cf0-4bb7-9b05-464a605e8d80.png)

stockiNail:
REMINDER: Breaking change: https://github.com/chartjs/chartjs-plugin-annotation/pull/655#issuecomment-1024531750

stockiNail:
This normalization will also enable (with common "properties"):

1. inner labels components (in line and box annotation) as sub-elements, based on the box
2. animation of labels

PR #655, opened 2022-01-24
Enable stacked scales and with reverse option support

### LINKED ISSUE #652: yMax / yMin does not work anymore without explicit boundaries
Hey,

I used to split the chart in 3 colored zones.

```
annotations: {
              first: {
                ...
                type: "box",
                yMin: 30,
              ..
              },
              second: {
                ...
                type: "box",
                yMax: 30,
                yMin: 1,
              ..
              },
              third: {
                ...
                type: "box",
                yMax: 1,
              ..
              },
....
}
```

But now it does not color the graph correctly anymore, unless I add yMax in the first box and yMin in the third box. Those values are dynamic though and I don't want to add boundaries.

Any advice?

Thanks!

stockiNail:
Fix #652

This PR is changing the defaults of `yMin/Max` options when missing.
It applies the same logic was implemented in the plugin version < 1.

It adds the control of the scale `reverse` options and stacked scales.

Recap about which locate methods is affected:

| Name | Scale Options | Values options | Stacked Scales | Reverse Scales |
| :- | :- | :- | :- | :-
| box | xScaleID, yScaleID | x/yMin, x/yMax | Yes | Yes
| point | xScaleID, yScaleID | x/yValue | Yes| No
| line | scaleID | value, endValue | No | No


src/helpers/helpers.chart.js
```diff
@@ -36,8 +36,8 @@ function getChartDimensionByScale(scale, options) {
     };
   }
   return {
-    start: options.start,
-    end: options.end
+    start: Math.min(options.start, options.end),
```
kurkle, line 39: These are potentially scale values. How does it work with a category scale with arbitrary labels like `['10', '5', '20', '4']`?

stockiNail: you're right, it doesn't work. Let me have a look.

stockiNail: I was, wrong because I have checked and using the a category scale and it works (I'm preparing a test case):

```javascript
scales: {
  x: {
    type: 'category',
    display: true,
  },
  y: {
    display: false,
    min: -10,
    max: 50
  }
},
plugins: {
  annotation: {
    annotations: {
      first: {
        type: 'box',
        xMax: 'February',
        backgroundColor: 'rgba(159, 226, 191, 0.5)',
      },
      second: {
        type: 'box',
        xMax: 'April',
        xMin: 'February',
        backgroundColor: 'rgba(255, 191, 0, 0.5)',
      },
      third: {
        type: 'box',
        xMin: 'April',
        backgroundColor: 'rgba(255, 191, 0, 0.5)',
      }
    }
  }
}
```
issueCoordinates1

stockiNail: Do you have other use cases that I forgot or I'm not aware of?

kurkle: Yeah, actually the code changed here is in the branch where scale is not defined. So the added test does not go there.

kurkle: But I was wrong, those are not potentialy scale values, those are chartArea coordinates.

stockiNail: > But I was wrong, those are not potentialy scale values, those are chartArea coordinates.

Yeah!

stockiNail: > Yeah, actually the code changed here is in the branch where scale is not defined. So the added test does not go there.

Ok... do you mean PR #653 where x/yScale will not be anymore in the options?

kurkle: No, by branch I mean the `if` conditions in that function, specifically:
image

src/helpers/helpers.chart.js
```diff
@@ -80,7 +80,7 @@ export function getChartRect(chart, options) {
   const xDim = getChartDimensionByScale(xScale, {min: options.xMin, max: options.xMax, start: x, end: x2});
   x = xDim.start;
   x2 = xDim.end;
-  const yDim = getChartDimensionByScale(yScale, {min: options.yMin, max: options.yMax, start: y, end: y2});
+  const yDim = getChartDimensionByScale(yScale, {min: options.yMin, max: options.yMax, start: y2, end: y});
```
kurkle, line 83: I think we should actually use `top, left, bottom, right` of the scale to determine the fallbacks. Because scales can be stacked. And we should consider the reverse option too.
So for x, we'd have start = left, end = right or if reversed start=right, end=left of the scale. 
And for y start = bottom, end = top or if reversed start=top, end=bottom of the scale.

stockiNail: OK! I agree. The check on reverse will be done in `getChartDimensionByScale`

stockiNail: done, adding test cases with reverse and stacked

kurkle: > I think we should actually use top, left, bottom, right **of the scale**

This part is not done. Should also add a test with stacked scales when this part is done.

Only testing reverse on x-scale, please add tests with y-scale too. Also hide legend etc not needed for the test. (but do not hide scale whose options we are testing against = the scale with `reverse: true` should be visible)

kurkle: By stacked scales I mean this feature:

image

stockiNail: ah ok. I have never used! let me try

stockiNail: have a look now. It should be better

test/fixtures/box/missingYMinMaxReverse.js
```diff
@@ -0,0 +1,45 @@
...
+          min: -10,
+          max: 50
+        }
+      },
+      plugins: {
+        annotation: {
+          annotations: {
+            first: {
+              type: 'box',
+              yMax: 1,
```
kurkle, line 23: Did you need to change the order of annotations for reversed scale? I would not expect that.

stockiNail: No, I didn't. I had tested both sequence and I left the last one. I'm committing having the same config.

stockiNail: it should be done

stockiNail:
@kurkle the stacked scales support is forcing us to change all functions used to get the location of annotations (not only the box ones) because the default dimension is not the chart area anymore but scale one.
Furthermore, the chart area is also used in some drawing phase to check if inside or not and also this part must be re-evaluated.
Therefore this PR needs additional updates

stockiNail:
Working on line annotation, I have recognized that the line annotation is working slightly different comparing with the other annotation when the location of the annotation is using `x/yMin` and `x/yMax`.
I wanted to use the helpers to get a rectangle on the chart (as used for all other annotations) but I have seen that it could work but this changes the current behavior of the line annotation.

Currently the `x,y` properties of line annotation element couldn't reflect the `top,left` of the box. The same for x2,y2 which couldn't be the `bottom,right`. This is because the properties are not checked and the line can be drawn correctly as well.

But `x,y` (and `x2,y2`) element properties are used by the sub elements of the annotation for the their position (label, arrow heads).

When #532 will be implemented, this could be taken care, because this could change significantly the chart rendering.

docs/guide/migrationV2.md
```diff
@@ -8,3 +8,4 @@ A number of changes were made to the configuration options passed to the plugin
 
  * `xScaleID` option default has been changed, now set to `undefined`. If the option is missing, the plugin will try to use the first scale of the chart, configured as `'x'` axis. If more than one scale has been defined in the chart as `'x'` axis, the option is mandatory to select the right scale.
  * `yScaleID` option default has been changed, now set to `undefined`. If the option is missing, the plugin will try to use the first scale of the chart, configured as `'y'` axis. If more than one scale has been defined in the chart as `'y'` axis, the option is mandatory to select the right scale.
+ * The scale dimension, not chart area anymore, is used to set `xMin`, `xMax`, `yMin` and `yMax` options when missing. This enables the stacked scale management.
```
kurkle, line 11: Should add a link to the chart.js docs for "stacked scale", because it might be confused with normal stacking.
Also could be worth mentioning that this only affects charts with stacked scales.

stockiNail: have a look if sounds better now

docs/guide/migrationV2.md
```diff
@@ -8,3 +8,5 @@ A number of changes were made to the configuration options passed to the plugin
 
  * `xScaleID` option default has been changed, now set to `undefined`. If the option is missing, the plugin will try to use the first scale of the chart, configured as `'x'` axis. If more than one scale has been defined in the chart as `'x'` axis, the option is mandatory to select the right scale.
  * `yScaleID` option default has been changed, now set to `undefined`. If the option is missing, the plugin will try to use the first scale of the chart, configured as `'y'` axis. If more than one scale has been defined in the chart as `'y'` axis, the option is mandatory to select the right scale.
+ * When [stacked scales](https://www.chartjs.org/docs/latest/axes/cartesian/#common-options-to-all-cartesian-axes) are used and  `xMin`, `xMax`, `yMin` and `yMax` options are missing, the scale dimension is used instead of chart area to set the annotation dimension.
+ * When [stacked scales](https://www.chartjs.org/docs/latest/axes/cartesian/#common-options-to-all-cartesian-axes) are used and  `xValue` and `yValue` options are missing, the scale dimension is used instead of chart area to set the annotation location.
```
kurkle, line 12:
```suggestion
* When [stacked scales](https://www.chartjs.org/docs/latest/axes/cartesian/#common-options-to-all-cartesian-axes) are used, instead of the whole chart area, the designated scale area is used as fallback for `xMin`, `xMax`, `yMin`, `yMax`, `xValue` or `yValue` options.
```

Not sure if this is any better. @etimberg what do you think?

etimberg: That makes sense to me

PR #657, opened 2022-01-24
Line: change event tests to inRange tests

stockiNail:


kurkle (commented):
Looks ok, but I don't really like that we have to be one pixel inside the element before the tests work. Also should test different borderWidths of the line (1 for the other edge case, or should it actually be 0.1?). Also without label.

kurkle:
> Coverage decreased (-1.08%) to 95.274%

stockiNail:
> Looks ok, but I don't really like that we have to be one pixel inside the element before the tests work. Also should test different borderWidths of the line (1 for the other edge case, or should it actually be 0.1?). Also without label.

Without 1 inRange return false. I need to go in deep and debug it. I have the feeling it's related to intersect function and the epsilon. Let me debug it and I'll come back with the result.

stockiNail:
> > Coverage decreased (-1.08%) to 95.274%

It's correct because the events part has been removed but it's present in the other PR.

kurkle:
> > Looks ok, but I don't really like that we have to be one pixel inside the element before the tests work. Also should test different borderWidths of the line (1 for the other edge case, or should it actually be 0.1?). Also without label.
> 
> Without 1 inRange return false. I need to go in deep and debug it. I have the feeling it's related to intersect function and the epsilon. Let me debug it and I'll come back with the result.

I think it would be fine if the test passes with `borderWidth: 1`. If it does not, then I think its worth taking a closer look.

stockiNail:
> > > Looks ok, but I don't really like that we have to be one pixel inside the element before the tests work. Also should test different borderWidths of the line (1 for the other edge case, or should it actually be 0.1?). Also without label.
> > 
> > 
> > Without 1 inRange return false. I need to go in deep and debug it. I have the feeling it's related to intersect function and the epsilon. Let me debug it and I'll come back with the result.
> 
> I think it would be fine if the test passes with `borderWidth: 1`. If it does not, then I think its worth taking a closer look.

I found the bug in intersect... Let me test again (with borderWidth: 1 and others)

stockiNail:
Ok the bug was here:

https://github.com/chartjs/chartjs-plugin-annotation/blob/c04e02ef2ff171cd878a308a34154c19632c00af/src/types/line.js#L66

It must be:

```javascript
return (sqr(x - xx) + sqr(y - yy)) <= epsilon;
```

stockiNail:
I'm committing the changes and then I'll rebase it.

kurkle:
I think after this we are good to go with 1.3.0?

So the version bump would be next. The we could give it a week or 2 before starting to merge v2 things.
The week or 2 so there is room to fix any new bugs found in 1.3 without creating any additional branches and stuff.

stockiNail:
> I think after this we are good to go with 1.3.0?
> 
> So the version bump would be next. The we could give it a week or 2 before starting to merge v2 things. The week or 2 so there is room to fix any new bugs found in 1.3 without creating any additional branches and stuff.

Completely agree with you!

test/specs/line.spec.js
```diff
@@ -1,3 +1,101 @@
...
+        }
+      };
+      const chart = window.scatterChart(10, 10, {annotation});
+      const element = window.getAnnotationElements(chart)[0];
+      it('should return true inside label of element', function() {
+        for (const borderWidth of [0, 10]) {
+          const halfBorder = borderWidth / 2;
+          element.options.label.borderWidth = borderWidth;
+          const rad = rotation / 180 * Math.PI;
+          for (const ax of [element.labelX - element.labelWidth / 2 - halfBorder + 1, element.labelX + element.labelWidth / 2 + halfBorder - 1]) {
```
kurkle, line 76: Similar issue (+-1) with the label `inRange`?

stockiNail: it depends on the decimal places.

DEBUG: 

Object{x: 186.9892578125, y: 237.79999999999998}, <-- rotated event
Object{labelX: 256, labelY: 256, w2: 69.0107421875, h2: 18.2} <-- label

For instance, with 45 degrees,  the check that fail is `y >= labelY - h2` because

y = 237.79999999999998 and (labelY - h2) = 237.8

It needs to add a margin, adding 0.00001 or something like that.

stockiNail: I use the same default espilo value of intersect: const epsilon = 0.001;

Ok for you?

stockiNail: have a look if ok for you, using a constant 0.001 as default epsilon in intersect and isOnLabel methods

test/specs/line.spec.js
```diff
@@ -72,17 +72,19 @@ describe('Line annotation', function() {
         for (const borderWidth of [0, 10]) {
           const halfBorder = borderWidth / 2;
           element.options.label.borderWidth = borderWidth;
+          element.debug = true;
```
kurkle, line 75: what is this?

stockiNail: sorry.... my debug

stockiNail: removed

kurkle (approved):

PR #659, opened 2022-01-26
Enable interaction mode support for events

stockiNail:
This PR is addressing the discussion #658

This PR is enabling the interaction mode support, similar to [CHART.JS](https://www.chartjs.org/docs/latest/configuration/interactions.html#interactions).

```javascript
annotation: {
  interaction: {
    mode: 'nearest' | 'point' | 'x' | 'y',
    axis: 'x' | 'y' | 'xy',
    intersect: boolean
  },
  annotations : ....
}
```
Similar to CHART.JS because this PR is implementation  `'nearest'` , `'point'`, `'x'`  and `'y'` modes (not all the others, more related to dataset elements). The fallback is to `chart.options`.
For `axis` option, `'x'`, `'y'` and `'xy'` values are supported, if `'r'` is set (not supported), the plugin will use `'xy'`. 

### TODO

- [x] box annotation
- [x] ellipse annotation
- [x] label annotation
- [x] line annotation
- [x] point annotation
- [x] polygon annotation
- [x] interaction mode 'x'
- [x] interaction mode 'y'
- [x] types
- [x] documentation


kurkle:
Looks ok for that use case. Not sure if it is the only use case though, maybe we need interaction modes here too?

stockiNail:
> Looks ok for that use case. Not sure if it is the only use case though, maybe we need interaction modes here too?

make sense to enable the interaction mode. I think we could use `nearest` (as default) and `point` (as this PR is implementing), using the same labels of CHART.JS.
Let me know if you would like to implement a subnode (as in CHART.JS) or only a options (`interactions`). I think the single option could better if we don't add additional options (like in CHART.JS, i.e. `intersect` and `axis`).

Furthermore I think we should add this new option to `annotation` node (valid for all annotations) and not to each single annotation.

What do you think?

kurkle:
Yes, I think the option should apply to all annotations. It could also fall back to chart.js interaction options, so we'd use the same mode by default. Need to decide what mode is "default" and what mode(s) to implement.

So, I think it makes sense to have a sub-node, and probably would be good to implement the same options too. The other options can be delt with later, but maybe its just some copied code from chart.js that is needed. 

Maybe the end result can be generalized so it can be used in chart.js too and thus be available to other plugins.
https://github.com/chartjs/Chart.js/pull/10046 is maybe relevant, though the functions used here would probably be one step lower level than those exposed by that PR.

kurkle:
This could probably be in v1.4 if we keep the current behavior as default.

stockiNail:
> Yes, I think the option should apply to all annotations. It could also fall back to chart.js interaction options, so we'd use the same mode by default. Need to decide what mode is "default" and what mode(s) to implement.

I think we should leave the current behavior (nearest) as default, to avoid breaking change. Nevertheless I need to go more in deep in interactions code (and implementation) of CHART.JS.



> So, I think it makes sense to have a sub-node, and probably would be good to implement the same options too. The other options can be delt with later, but maybe its just some copied code from chart.js that is needed.

:+1:
I have to admit that I have to study how the other options are really working (intersect and axis) and the impact in this plugin as well.

> Maybe the end result can be generalized so it can be used in chart.js too and thus be available to other plugins.
> [chartjs/Chart.js#10046](https://github.com/chartjs/Chart.js/pull/10046) is maybe relevant, though the functions used here would probably be one step lower level than those exposed by that PR.

Let me take time to have a look to that PR. I need some days because busy in other stuff.
 
> This could probably be in v1.4 if we keep the current behavior as default.

Yes, we could but maybe it's not so important to release it soon therefore it could wait for version 2 where there are already important updates to do. Anyway, if you agree, let's wait for feedback about version 1.3 and if possible issues will come.

kurkle:
> Yes, we could but maybe it's not so important to release it soon therefore it could wait for version 2 where there are already important updates to do. Anyway, if you agree, let's wait for feedback about version 1.3 and if possible issues will come.

If we can release an enhancement as a non breaking change, we should do it before releasing the breaking changes. That way users can have that enhancement and not need to deal with the breaking changes just yet. Could also provide some feedback that requires another breaking change, which we can then do in v2 instead of v3.

Another thing is the number of changes, in 1.3 we already have quite a lot.

stockiNail:
> If we can release an enhancement as a non breaking change, we should do it before releasing the breaking changes. That way users can have that enhancement and not need to deal with the breaking changes just yet. Could also provide some feedback that requires another breaking change, which we can then do in v2 instead of v3.
> 
> Another thing is the number of changes, in 1.3 we already have quite a lot.

Ok for me!!

stockiNail:
I have started working on the interaction mode.
First feedback is that the default in CHART.JS is always `nearest` apart of scatter chart which is `point`.
To whom is using scatter charts, this PR could be a breaking change if we want to fallback to CHART.JS options.

src/events.js
```diff
@@ -121,6 +114,17 @@ function dispatchEvent(handler, element, event) {
   callback(handler, [element.$context, event]);
 }
 
+function getItems(state, event, options) {
+  if (options.interaction.mode === 'point') {
+    return state.visibleElements.filter((element) => element.inRange(event.x, event.y));
+  }
+  const element = getNearestItem(state.visibleElements, event);
```
kurkle, line 121: if removed the `[0]` from this line from getNearestItems:
> .slice(0, 1)[0]; // return only the top item

then could just return the result of `getNearestItem`

stockiNail: Ok
Honestly I left it as I found it.
Anyway it is a working in progress because I would like to implement all options pf interaction and then this function should be changed however 
What do you think?

stockiNail: @kurkle I had a look and I was starting to add other modes, axis and intersect options. Nevertheless for the other modes it could be a little complex without the normalization of element properties (I mean x, y, x2, y2, width, height, centerX and centerY, as planned with PR #532).
That said, I would propose to stay with this kind of first implementation of interaction, only with 'point' and 'nearest', ignoring axis and intersect.
When PR #532 will be implemented, we could create another PR adding all missing modes and options.
make sense?

stockiNail: > if removed the `[0]` from this line from getNearestItems:
> 
> > .slice(0, 1)[0]; // return only the top item
> 
> then could just return the result of `getNearestItem`

DONE

stockiNail: > @kurkle I had a look and I was starting to add other modes, axis and intersect options. Nevertheless for the other modes it could be a little complex without the normalization of element properties (I mean x, y, x2, y2, width, height, centerX and centerY, as planned with PR #532).
> That said, I would propose to stay with this kind of first implementation of interaction, only with 'point' and 'nearest', ignoring axis and intersect.
> When PR #532 will be implemented, we could create another PR adding all missing modes and options.
> make sense?

or I can add inYRange and inXRange in all annotations

src/interaction.js
```diff
@@ -0,0 +1,84 @@
...
+}
+
+function inRangeByAxis(element, event, axis) {
+  if (axis === 'x') {
+    return element.inXRange(event.x, event.y, true);
+  } else if (axis === 'y') {
+    return element.inYRange(event.x, event.y, true);
+  }
+  return element.inXRange(event.x, event.y, true) || element.inYRange(event.x, event.y, true);
+}
```
kurkle, line 51: I'd favor a single `inRange` function, by maybe adding the axis as parameter or setting unused coordinate to undefined.
Disclaimer: I did not check the inRange logics on how complicated it would be to ignore the coordinate that is undefined.

src/events.js
kurkle: Maybe the normalization should be done first?

src/interaction.js
stockiNail: Yes, I arrived to the same conclusion. Only 1 method, passing axis. About the to use only inRange, well, it could be. For sure we will add some complexity but I didn't do it because I was thinking that inRange should be a method of CHARTJS Element class, which is extended here but the signature should be the same (apologize my java approach). Adding axis as last argument (with a default) it could be done

src/events.js
stockiNail: good question... maybe for this PR is not strictly needed. I see the normalization could help the other PR, #655.
I would suggest to go with a single method for inRange (as suggested below).

src/interaction.js
stockiNail: everything in `inRange`

stockiNail:
@kurkle apart the review, there are 2 CC issues:

1. `File annotation.js has 253 lines of code (exceeds 250 allowed).`: for this issue, I'd suggest to move out the "elements" part in a dedicated file (as we did for scales). Another PR is needed.
2. `Identical blocks of code found in 2 locations.`: this issue is related to `box` and `label` annotations which have got the same `inRange` method. The solution could be to use the inheritance, creating a "baseBox" annotation (with only `inRange` method) which will be extended by both annotation.

kurkle:
> @kurkle apart the review, there are 2 CC issues:
> 
> 1. `File annotation.js has 253 lines of code (exceeds 250 allowed).`: for this issue, I'd suggest to move out the "elements" part in a dedicated file (as we did for scales). Another PR is needed.

This is fine, we can fix it in another PR.

> 2. `Identical blocks of code found in 2 locations.`: this issue is related to `box` and `label` annotations which have got the same `inRange` method. The solution could be to use the inheritance, creating a "baseBox" annotation (with only `inRange` method) which will be extended by both annotation.

This one can be ignored, reporting duplication of one (or 3) line(s) is just silly.

stockiNail:
> This is fine, we can fix it in another PR.

I'm doing

docs/guide/migrationV2.md
```diff
@@ -9,3 +9,9 @@ A number of changes were made to the configuration options passed to the plugin
  * `xScaleID` option default has been changed, now set to `undefined`. If the option is missing, the plugin will try to use the first scale of the chart, configured as `'x'` axis. If more than one scale has been defined in the chart as `'x'` axis, the option is mandatory to select the right scale.
  * `yScaleID` option default has been changed, now set to `undefined`. If the option is missing, the plugin will try to use the first scale of the chart, configured as `'y'` axis. If more than one scale has been defined in the chart as `'y'` axis, the option is mandatory to select the right scale.
 * When [stacked scales](https://www.chartjs.org/docs/latest/axes/cartesian/#common-options-to-all-cartesian-axes) are used, instead of the whole chart area, the designated scale area is used as fallback for `xMin`, `xMax`, `yMin`, `yMax`, `xValue` or `yValue` options.
+
+## Events
+
+`chartjs-plugin-annotation` plugin version 2 introduces the `interaction` options, to configure which events trigger annotation interactions. By default, the plugin uses the [chart interaction configuration](https://www.chartjs.org/docs/latest/configuration/interactions.html#interactions).
```
kurkle, line 15:
```suggestion
`chartjs-plugin-annotation` plugin version 2 introduces the [`interaction`](options#interaction) options, to configure which events trigger annotation interactions. By default, the plugin uses the [chart interaction configuration](https://www.chartjs.org/docs/latest/configuration/interactions.html#interactions).
```

kurkle (approved):

docs/guide/migrationV2.md
kurkle: maybe add a link to the options page here?

kurkle (approved):

PR #664, opened 2022-02-01
Update the warning logic for missing scale

### LINKED ISSUE #662: Revise warning if a non-existing scale ID is used?
I noticed that chartjs-plugin-annotation has started logging several warnings similar to the following for my charts:

> No scale found with id 'y' for annotation 'my-annotation-id'

This appears to be the result of #526. In my case, I'm drawing vertical lines, so I leave all of the Y-related properties (`yAxisID`,  `yMin`, `yMax`) blank. So the fact that a default `y` scale doesn't exist shouldn't matter: I want the default documented behavior of "if `yScaleId` is not resolved, the line will take the entire chart height." (And, without `yMin` or `yMax`, the annotation would take the entire chart dimension regardless.)

Could/should the warning logic from #526 be revised? Instead of warning whenever the configured scale doesn't exist, it could warn if the configured scale doesn't exist and either of the following are true:

* The configured scale is not the default value.
* The annotation has configured properties for the scale's associated values (`value` and `endValue` for `scaleID`; `xMin`, `xMax`, `xValue` for `xScaleID`; `yMin`, `yMax`, `yValue` for `yScaleID`).

stockiNail:
Fix #662

The warning message will be emitted on the console if the scaleID is missing and also if there is any options to locate the annotation (`x/yValue`, `x/yMin` and  `x/yMax`).

### DIFF
--- a/src/scale.js
+++ b/src/scale.js
@@ -29,10 +29,23 @@ function scaleLimitDefined(scaleOptions, limit, suggestedLimit) {
 
 function verifyScaleIDs(annotation, scales) {
   for (const key of ['scaleID', 'xScaleID', 'yScaleID']) {
-    if (annotation[key] && !scales[annotation[key]]) {
+    if (annotation[key] && !scales[annotation[key]] && verifyProperties(annotation, key)) {
       console.warn(`No scale found with id '${annotation[key]}' for annotation '${annotation.id}'`);
     }
   }
+}
+
+function verifyProperties(annotation, key) {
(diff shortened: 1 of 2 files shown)


kurkle (commented):
A regression test would be good.

kurkle (approved):

kurkle:
@stockiNail  
I'm ok to push 1.3.1 out with only this change, because there are no other issues to fix currently.
Could you do the version bump?

PR #668, opened 2022-02-04
Enable rotation support to box annotation

### LINKED ISSUE #666: Add rotation support to box annotation

stockiNail:
Fix #666

The "label" rotation will be addressed by another PR.


stockiNail:
@kurkle sorry but I need to ask to you a hint based on your huge experience on "canvas" javascript development.

Developing this PR, I discovered an interesting thing that I didn't know it.
As you can see from updated files, I had to change the test case where the `backgroundColor` callback was invoked to set a gradient as background of the box.
I discovered that if you change the translation matrix (`context.translate(point)` to enable the rotation) in the canvas, the gradient must be created with the relative positions on the canvas and not the absolute ones.
Therefore, using the element position (`x, y, x2, y2`), the gradient is not applied correctly in the element but the coordinates must be translated in relative ones.
This behavior is affecting all annotations/sub-elements which are translating the canvas: ellipse, box (here), label (when rotation will be implemented). The line label is not affected now because the `labelX` and `labelY` properties are providing the center of the label but, when the normalization will be applied, it will be affected as well.

That said, in your opinion, are the users aware about that? 
If not, how do you think we could mitigate it?
Having a poor experience on canvas/javascript, I'm not aware how much this behavior can annoy the users. For me, it's annoying...

I thought about 2 possible things which could mitigate this behavior:

1. when the rotation haven't to be applied (=0), we can not use the translate on the canvas
2. provide a method in the elements which can translate the absolute coordinates in relative ones.

Maybe you have got another solution. 

Last but not least, based on above thought, I have a doubt if this PR could be a breaking change (and also the rotation enablement on label annotation will be), therefore it should be moved to version 2.0.0 (after the normalization).  

Apologize for long comment...

kurkle:
@stockiNail the context can be translated back after rotation, so the drawing can be done using original coordinates (and thus gradient works as expected):
https://codepen.io/kurkle/pen/OJORwKO

stockiNail:
> @stockiNail the context can be translated back after rotation, so the drawing can be done using original coordinates (and thus gradient works as expected):
> https://codepen.io/kurkle/pen/OJORwKO

**Thank you!!** I didn't know it. We should apply it on ellipse too (another PR, maybe could be set a bug). Tomorrow I'll change this PR.

PR #673, opened 2022-02-07
Bump version to 1.3.1

stockiNail:


kurkle:
I don't see any updates to dependencies, which is good, because those should not be updated for minor versions.

kurkle (approved):

stockiNail:
> I don't see any updates to dependencies, which is good, because those should not be updated for minor versions.

Fine! In fact, I have seen that only the version is changed in package-lock.json.

kurkle:
It takes a hour or so before "latest" works properly (due to cloudflare caching)

PR #675, opened 2022-02-09
Add some missing element properties to annotation element type

stockiNail:
This PR is adding missing properties to annotation element type, waiting for annotation element normalization.

### DIFF
--- a/types/element.d.ts
+++ b/types/element.d.ts
@@ -3,7 +3,15 @@ import { AnnotationOptions } from './options';
 export interface AnnotationElement {
   x: number,
   y: number,
-  x2: number,
-  y2: number,
+  x2?: number,
+  y2?: number,
+  width: number,
+  height: number,
+  pointX?: number,
+  pointY?: number,
+  labelX?: number,
+  labelY?: number,
+  labelWidth?: number,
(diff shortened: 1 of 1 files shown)


kurkle:
Does not make sense to add the properties that don't yet work. Also I'd like to get rid of some of those in v2.

stockiNail:
> Does not make sense to add the properties that don't yet work. Also I'd like to get rid of some of those in v2.

I could agree with you about v2. Nevertheless without those properties, I think the users can not use the callbacks to provide a gradient as background color.

For sure, we can close it and wait for v2.

kurkle:
If we want to fix the types for v1, we should make the properties depend on the annotation type. Because labelX for example is only available in line and label annotations.

stockiNail:
> If we want to fix the types for v1, we should make the properties depend on the annotation type. Because labelX for example is only available in line and label annotations.

Yes but being a "optional" property, could be undefined where labelX is not used. I was thinking to wait for version 2 but maybe someone will raise an issue because it's able to use them in a callback. Anyway, we can wait and close it because in version 2, this must be reviewed however.

stockiNail:
Closed waiting for v2

PR #678, opened 2022-02-10
Use scale defined in the chart when x/y ScaleID options are not set

### LINKED ISSUE #676: Line is diagonal if yAxis options set
If no title is set for yAxis, it behaves correctly, there is an horizontal line

But if we set a title, the line becomes diagonal:

## To reproduce,

Go on https://www.chartjs.org/chartjs-plugin-annotation/master/guide/types/line.html#configuration and add in the options:

```javascript
scales: {
    yAxis: {
        title: {
            display: true,
            text: "My First Dataset"
        }
    }
},
```

stockiNail:
Fix #676
This  PR removes the `x/yScaleID` options defaults from annotation. It will use the scale (defined as x or y) in the chart and only when 1 scale for the coordinate is defined.

### TODO

- [x] documentation update


src/helpers/helpers.chart.js
```diff
@@ -21,6 +21,25 @@ export function scaleValue(scale, value, fallback) {
...
 
+/**
+ * @param {Object} scales - chartjs object with all scales
+ * @param {Object} options - plugin options
+ * @param {string} key - annotation plugin scale id option key
+ * @returns {string} the unique scale defined in chartjs or the key passed as argument
+ */
+export function retrieveScaleID(scales, options, key) {
+  const scaleID = options[key];
+  if (scaleID || key === 'scaleID') {
```
kurkle, line 32: What is this `key === 'scaleID'` case?

src/helpers/helpers.chart.js
```diff
@@ -21,6 +21,25 @@ export function scaleValue(scale, value, fallback) {
...
+    return scaleID;
+  }
+  const axis = key.charAt(0);
+  const axes = Object.values(scales).filter((scale) => scale.axis && scale.axis === axis);
+  if (axes.length === 1) {
```
kurkle, line 37: Why not just default to the first scale when there are multiple and user did not specify which one to use?

src/scale.js
```diff
@@ -29,8 +30,9 @@ function scaleLimitDefined(scaleOptions, limit, suggestedLimit) {
 
 function verifyScaleIDs(annotation, scales) {
   for (const key of ['scaleID', 'xScaleID', 'yScaleID']) {
-    if (annotation[key] && !scales[annotation[key]] && verifyProperties(annotation, key)) {
-      console.warn(`No scale found with id '${annotation[key]}' for annotation '${annotation.id}'`);
+    const scaleID = retrieveScaleID(scales, annotation, key);
+    if (scaleID && !scales[scaleID] && verifyProperties(annotation, key)) {
```
kurkle, line 34: Can `scaleID` still be undefined?

kurkle (commented):
Looks good overall, just couple of comments / questions,

src/helpers/helpers.chart.js
stockiNail: The check can be removed and returns always  `return axes[0].id;`. Anyway we must check if there is not any scale,

src/scale.js
stockiNail: uhm... you are right. it's not possible. The first check must be removed.

stockiNail:
> Looks good overall, just couple of comments / questions,

If we will start going to V2 and due to element normalization, maybe it could be helpful to define the sequence of PR merging in order to help the rebasing, where needed. My 2 cents.

kurkle:
> > Looks good overall, just couple of comments / questions,
> 
> If we will start going to V2 and due to element normalization, maybe it could be helpful to define the sequence of PR merging in order to help the rebasing, where needed. My 2 cents.

I think that would be your task, because you've done all the work. So maybe just keep things as draft and mark the next one ready after merge.

stockiNail:
> I think that would be your task, because you've done all the work. So maybe just keep things as draft and mark the next one ready after merge.

Fine for me. In fact, there are all in Draft, apart of the V2 migration page which must be the first one.

src/scale.js
stockiNail: I checked better and the `retrieveScaleID` function can return an undefined value when the `scaleID` options is not set. See above comment related to `scaleID` option.

kurkle (dismissed):

kurkle (approved):

PR #682, opened 2022-02-11
Fix callout drawing that happens even if disabled

### LINKED ISSUE #681: Label callout is drawn even if not enabled
Label callout is drawn even if not enabled

stockiNail:
Fix #681


src/types/label.js
```diff
@@ -127,7 +127,7 @@ function calculatePosition(start, size, adjust = 0, position) {
 
 function drawCallout(ctx, element) {
   const {pointX, pointY, calloutPosition, options} = element;
-  if (element.inRange(pointX, pointY)) {
+  if (element.inRange(pointX, pointY) || !calloutPosition) {
```
kurkle, line 130: Turn the checks around, `!calloutPosition` is a faster test and the inRange test is skipped when that is true.

kurkle: + add a test

stockiNail: YES!

stockiNail: done

kurkle:
Some kind of regression test would be good to have, as this went through. A simple fixture is probably best, maybe as additional annotation in an existing one testing callouts?

stockiNail:
> Some kind of regression test would be good to have, as this went through. A simple fixture is probably best, maybe as additional annotation in an existing one testing callouts?

Yes, I'll commit it in next minutes.

kurkle (approved):

PR #684, opened 2022-02-11
Add samples of image and canvas as label content

### LINKED ISSUE #680: Missing sample(s) of image/canvas as content
All of the annotations accepting a label can also draw image or canvas instead of text. This is documented, but there are no samples.

stockiNail:
Fix #680

### DIFF
--- a/docs/samples/box/canvas.md
+++ b/docs/samples/box/canvas.md
@@ -43,7 +43,7 @@ const annotation2 = {
   xMin: 10,
   label: {
     enabled: true,
-    content: Utils.getHouse,
+    content: Utils.getHouse(),
     position: {
       x: 'center',
       y: 'end'
--- a/docs/samples/box/canvas.md
+++ b/docs/samples/box/canvas.md
@@ -25,7 +25,7 @@ const annotation1 = {
   xMin: 5,
   label: {
     enabled: true,
-    content: Utils.getSpiral,
+    content: Utils.getSpiral(),
     position: {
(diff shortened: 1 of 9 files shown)


docs/samples/box/canvas.md
```diff
@@ -0,0 +1,93 @@
...
+// <block:annotation1:1>
+const annotation1 = {
+  type: 'box',
+  backgroundColor: 'rgba(165, 214, 167, 0.2)',
+  borderWidth: 0,
+  xMax: 2,
+  xMin: 5,
+  label: {
+    enabled: true,
+    content: Utils.getSpiral,
```
LeeLenaleee, line 28:
```suggestion
    content: Utils.getSpiral(),
```

While both syntaxes work I think this is a bit nicer looking, colors in the console look better and its a bit more clear its a function that gets called that can return something

LeeLenaleee (changes requested):
Add a tab for the utils file as done in the chart.js sample's so people can more easily see how we got the spiral and house, That it really is done with canvas and not just an image.

https://github.com/chartjs/Chart.js/blob/master/docs/samples/utils.md
https://www.chartjs.org/docs/master/samples/utils.html

stockiNail:
> Add a tab for the utils file as done in the chart.js sample's so people can more easily see how we got the spiral and house, That it really is done with canvas and not just an image.

Thank you @LeeLenaleee I have never seen that page. I'm going to add it.

PR #696, opened 2022-02-23
Add inner chart sample

stockiNail:
Add ne sample for label annotation.

innerChart


docs/scripts/utils.js
```diff
@@ -50,6 +51,42 @@ export function getImage() {
   return img;
 }
 
+var pieChart = null;
```
LeeLenaleee, line 54: We use const and var, think it's better to use const and let instead. Bit cleaner

LeeLenaleee (commented):
1 quick thing that might be out of scope but is a bit of an eyesore to me, will give it a better look later tonight

docs/scripts/utils.js
stockiNail: I agree with and I wanted to use `let` but, on the top of `utils`, there is `var _seed = Date.now();` therefore I assumed that in utils file the use of `var` is preferable.
Can I change both?

stockiNail: I have changed both to **let**

docs/samples/label/innerChart.md
```diff
@@ -0,0 +1,152 @@
...
+    data: Utils.numbers(numberCfg),
+    borderColor: innerDataset.backgroundColor[0]
+  }, {
+    data: Utils.numbers(numberCfg),
+    borderColor: innerDataset.backgroundColor[1],
+    hidden: true
+  }, {
+    data: Utils.numbers(numberCfg),
+    borderColor: innerDataset.backgroundColor[2],
+    hidden: true
```
LeeLenaleee, line 26: Add `backgroundColor: 'transparent'` here and remove `autocolors: false` in the plugin. Makes it a bit easyer for people who dont know much about other chart.js plugins.

docs/scripts/utils.js
```diff
@@ -50,6 +51,22 @@ export function getImage() {
   return img;
 }
 
+let pieChart = null;
+
+export function getChart() {
+  if (pieChart) {
+    pieChart.update();
```
LeeLenaleee, line 58: Is this update really necesarry because you never change the chart after you made it so you can just return it?

docs/samples/label/innerChart.md
```diff
@@ -0,0 +1,152 @@
...
+  xValue: (ctx) => point(ctx).x,
+  yAdjust: -60,
+  yValue: (ctx) => point(ctx).y
+};
+// </block:annotation2>
+
+// <block:annotation3:3>
+const annotation3 = {
+  type: 'label',
+  content: ['Select the data to see', 'clicking the inner chart data'],
```
LeeLenaleee, line 68:
```suggestion
  content: ['Click on a slice of the pie chart', 'to see the data in the main chart'],
```

LeeLenaleee (commented):
Nice sample, 3 more small things.

docs/scripts/utils.js
stockiNail: well... I have added at the beginning and I don't know why I left it. I'll remove it.
But I want to randomize the data of the inner chart too, when the main will be updated. In this way it will be clear that it's a chartjs instance and not an image.

stockiNail: done, the randomize of inner chart data too.

docs/samples/label/innerChart.md
stockiNail: done!

LeeLenaleee (approved):

kurkle (approved):

PR #709, opened 2022-03-07
Add migration guide v2 section to getting started page

stockiNail:
Add migration guide v2 section to getting started page

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -82,6 +82,7 @@ module.exports = {
     sidebar: {
       '/guide/': [
         '',
+        'migrationV2',
         'integration',
         'usage',
         'options',
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -82,7 +82,6 @@ module.exports = {
     sidebar: {
       '/guide/': [
         '',
-        'migrationV2',
         'integration',
         'usage',
         'options',
@@ -98,7 +97,8 @@ module.exports = {
             'types/point',
             'types/polygon'
           ]
-        }
+        },
+        'migrationV2'
       ],
       '/samples/': [
         'intro',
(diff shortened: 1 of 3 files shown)


kurkle:
While it might not get too long, I think a separate page would be good.

stockiNail:
> While it might not get too long, I think a separate page would be good.

Yes, I thought the same but in CHART.JS the migration guise is inside the Getting started session.
The Getting started plugin is a page and not a group.
 group? separated page from getting started? What would you prefer?

LeeLenaleee:
I think I would place at the bottom of the nav, and then when a V3 comes out we can make it into a seperate sub category where we list each migration

stockiNail:
> I think I would place at the bottom of the nav, and then when a V3 comes out we can make it into a seperate sub category where we list each migration

done!

kurkle (approved):

kurkle:
Are we ready to move towards v2?

stockiNail:
> Are we ready to move towards v2?

well, good question. I don't see any other enhancement (or bug) to address, in this moment. 
Personally I'm quite busy in this period for a couple of projects but maybe I could find time. Be patient.

And you, what do you think?

kurkle:
> > Are we ready to move towards v2?
> 
> well, good question. I don't see any other enhancement (or bug) to address, in this moment. Personally I'm quite busy in this period for a couple of projects but maybe I could find time. Be patient.
> 
> And you, what do you think?

Quite busy too still, so maybe we should wait for a little more then.

stockiNail:
> Quite busy too still, so maybe we should wait for a little more then.

Agree, I've submitted those PRs in order do not forget what I had in mind (getting old it happens) about implementation.

stockiNail:
> Are we ready to move towards v2?

Shall we start version 2.0 on Apr 4th? 4 weeks after version 1.4.0, make sense?

kurkle:
> > Are we ready to move towards v2?
> 
> Shall we start version 2.0 on Apr 4th? 4 weeks after version 1.4.0, make sense?

Sounds good to me!

PR #711, opened 2022-03-18
Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/src/helpers/helpers.core.js
+++ b/src/helpers/helpers.core.js
@@ -30,7 +30,7 @@ export function getElementCenterPoint(element, useFinalPosition) {
   return {x, y};
 }
 
-const isOlderPart = (act, req) => req > act || (act.length > req.length && act.substr(0, req.length) === req);
+const isOlderPart = (act, req) => req > act || (act.length > req.length && act.slice(0, req.length) === req);
 
 export function requireVersion(pkg, min, ver, strict = true) {
   const parts = ver.split('.');


LeeLenaleee (approved):

kurkle (approved):

kurkle:
Merged as this was not a breaking change. If we decide to release another minor for v1, the milestone of this pr needs to be updated.

stockiNail:
@kurkle I have seen that this PR created a release note 1.4.1 even if tagged as 2.0.0:  https://github.com/chartjs/chartjs-plugin-annotation/releases/tag/untagged-e554510da7259acef101

I don't know if this is correct and a normal behavior.

kurkle:
release-drafter determines the version based on version resolver config:
https://github.com/chartjs/chartjs-plugin-annotation/blob/67989c8164c3644523123cbfc098198be638e2db/.github/release-drafter.yml#L28-L40

so unless its a major or minor, its a patch :)

it will update after new commits are added.

kurkle:
https://en.wikipedia.org/wiki/No_such_thing_as_a_stupid_question

Issue #718, opened 2022-04-07
Interaction axis options default in the samples is not xy


stockiNail:
Default interaction options in the samples are set to 'x' instead of 'xy' as expected.

https://github.com/chartjs/chartjs-plugin-annotation/blob/70078ccbe49568f6a4106667ad008bef7edf5495/docs/scripts/defaults.js#L9-L13

kurkle:
Why is `xy` expected?

stockiNail:
Ok, doable. Nevertheless, my opinion as user, I'm expecting to go to the samples and to have the default chart options (in this case interaction is not) because if I'm adding the interaction to the sample, it seems the user should always do. But that's my point of view and maybe I'm wrong.  Anyway, let me change the PR following your hint.

kurkle:
I'm ok removing the defaults and setting in the samples where required. Its better that way. I think the current defaults are to make the charts work in an opinionated way, which is not good.

stockiNail:
> I'm ok removing the defaults and setting in the samples where required. Its better that way. I think the current defaults are to make the charts work in an opinionated way, which is not good.

Already done

PR #721, opened 2022-04-07
Add sample about interaction options

stockiNail:
This PR is adding a specific sample where the user can check the different behaviors with different interaction options.

https://github.com/chartjs/chartjs-plugin-annotation/issues/718#issuecomment-1091725296

### DIFF
--- a/docs/samples/interaction.md
+++ b/docs/samples/interaction.md
@@ -23,26 +23,6 @@ const data = {
 };
 // </block:setup>
 
-// <block:printer:4>
-const printer = {
-  id: 'printer',
-  afterDraw(chart) {
-    const options = chart.options.plugins.annotation.interaction;
-    const mode = options.mode || 'nearest';
-    const axis = options.axis || 'xy';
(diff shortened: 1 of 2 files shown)


docs/samples/interaction.md
```diff
@@ -0,0 +1,184 @@
...
+  yMax: 75,
+  yMin: 25,
+  yScaleID: 'y'
+};
+// </block:annotation2>
+
+/* <block:config:0> */
+const config = {
+  type: 'line',
+  plugins: [printer],
```
LeeLenaleee, line 117: Maby instead of a custom plugin to put the info on the chart you can use a scriptable function for the title. Don't know what is more clear so if this is looks better its also fine

stockiNail: good hint! Let me try

stockiNail: Plugin has been removed and it uses the title as suggested (really better, thanks a lot @LeeLenaleee).

LeeLenaleee (approved):

PR #723, opened 2022-04-08
Require at least chartjs v3.7

LeeLenaleee:
Remove todo for V2.

Although not sure if we want it to be minimal 3.7 because of the bugfix or 3.8 if we want to remove this todo as wel:
https://github.com/chartjs/chartjs-plugin-annotation/blob/70078ccbe49568f6a4106667ad008bef7edf5495/src/helpers/helpers.canvas.js#L07-L18


stockiNail (approved):

stockiNail:
> Although not sure if we want it to be minimal 3.7 because of the bugfix or 3.8 if we want to remove this todo as wel:

In my opinion, I will leave the function in the plugin till the function will be available and released in chartjs.

PR #724, opened 2022-04-11
Update min version req in readme

LeeLenaleee:
Addition on #723

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,12 +7,13 @@
 [![documentation](https://img.shields.io/static/v1?message=Documentation&color=informational)](https://www.chartjs.org/chartjs-plugin-annotation/index)
 <a href="https://github.com/chartjs/awesome"><img src="https://awesome.re/badge-flat2.svg" alt="Awesome"></a>
 
-An annotation plugin for Chart.js >= 3.0.0
+An annotation plugin for Chart.js >= 3.7.0
 
 ---
 > This plugin needs to be registered. It does not function as inline plugin.
 ---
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,12 +7,13 @@
...
 <a href="https://github.com/chartjs/awesome"><img src="https://awesome.re/badge-flat2.svg" alt="Awesome"></a>
 
-An annotation plugin for Chart.js >= 3.0.0
+An annotation plugin for Chart.js >= 3.7.0
 
 ---
 > This plugin needs to be registered. It does not function as inline plugin.
 ---
 
+For Chart.js 3.0.0 to 3.6.2 support, use [version 1.4.0 of this plugin](https://github.com/chartjs/chartjs-plugin-annotation/releases/tag/v1.4.0)
```
stockiNail, line 16: In theory, the version 1.4.0 can be used also with Chart.js version 3.7.x, due to the workaround.

LeeLenaleee: while it is true that so far you can use 1.4 with 3.7 and prob 3.8 and forward I think its more staightforward to put it like this so you also can't get in the future why isn't 1.4 working with 3.9 for example (because there might be a similar bugg that the workaround got used for). Guess I can put between brackets behind it that it might work above 3.6.2 but not guarantees or support if it breaks

kurkle: To me the requested text looks ok. Maybe make it `version 1.4` so it leaves room for possible bug fixes in that (1.4.1 etc).

kurkle (approved):

Issue #175, opened 2019-08-16
Feature request: Annotation z-index


michealjroberts:
I've noticed that the annotation line is somehow z-indexed on the Canvas higher than the hover tooltips for the normal Chart.js line tooltips.

johnmifit:
I agree, it appears that sometimes boxes are being layer in front of lines even though the lines are further back in the `annotations` array.

diegoiglesias:
+1

sheadawson:
+1

sheadawson:
I just had this issue with boxes drawing over my line graph lines. Quick fix: add this before adding your annotations

`chart.ctx.globalCompositeOperation='destination-over';`

sheadawson:
Still had a weird issue with the hover tooltips. Worked around with custom html tooltips https://www.chartjs.org/samples/latest/tooltips/custom-line.html

firebird631:
I have the same issue, the annotation line appears above the hover toolip and I find no option to control that.

kurkle:
@firebird631 what is your `drawTime` set to?

firebird631:
@kurkle I don't understand, all is ok today :) But I have no idea of what I have changed that fix it. I'll make a reply if I found what was the mistake.

shmkane:
Was there ever a solution to this?

elitastic:
Yes, drawTime should be set to "beforeDraw"!

stockiNail:
Even if the `drawTime` could solve some use cases, it could happen that more annotations must be drawn at the same `drawTime` (in order to be consistent with the other elements of the chart) but one annotation can not be visible because the another annotation is drawn on top.

See codepen: https://codepen.io/stockinail/pen/PoJoMbr

In the sample, the point annotation is not visible because the box one is drawn on the same "area".
The workaround is to put the annotations config in the order of drawing. 
Nevertheless it sounds to me just a workaround.
For this use case, I think it could be better to add `z` option to the annotations in order to draw the annotations in the desired order.

kurkle:
I agree, but the `z` should be global to the chart and thus be supported in the core too. I would not want to end up having inconsistent z option in every plugin.

I think one way of doing that is:

- collect all the z values in use
- call draw with each z value for each dataset/scale/plugin in order.

Probably a breaking change.

stockiNail:
@kurkle sounds good to me. 
[...]

PR #725, opened 2022-04-11
Enable box annotation label as label sub-element

stockiNail:
This PR moves label node of box annotation to sub element.

This PR enables the box label animation.

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -142,7 +142,10 @@ function draw(chart, caller, clip) {
     clipArea(ctx, chartArea);
   }
 
-  drawElements(ctx, visibleElements, caller);
+  const todraw = drawElements(ctx, visibleElements, caller);
+  for (const item of todraw) {
+    item.element.draw(ctx, item.box);
+  }
 
   if (clip) {
     unclipArea(ctx);
@@ -160,21 +163,23 @@ function draw(chart, caller, clip) {
 }
 
 function drawElements(ctx, elements, caller) {
+  const todraw = [];
   for (const el of elements) {
     if (el.options.drawTime === caller) {
-      el.draw(ctx);
+      todraw.push({element: el});
     }
     if (el.elements && el.elements.length) {
-      drawSubElements(ctx, el, caller);
+      drawSubElements(ctx, el, caller, todraw);
     }
   }
+  return todraw;
 }
 
-function drawSubElements(ctx, el, caller) {
+function drawSubElements(ctx, el, caller, todraw) {
   const box = 'getBoundingBox' in el ? el.getBoundingBox() : undefined;
   for (const sub of el.elements) {
     if (sub.options.drawTime === caller) {
-      sub.draw(ctx, box);
+      todraw.push({element: sub, box});
     }
   }
 }
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -143,7 +143,6 @@ function draw(chart, caller, clip) {
   }
 
   drawElements(ctx, visibleElements, caller);
-  drawSubElements(ctx, visibleElements, caller);
 
   if (clip) {
     unclipArea(ctx);
@@ -160,18 +159,18 @@ function draw(chart, caller, clip) {
   });
 }
 
-function drawElements(ctx, elements, caller, parent) {
+function drawElements(ctx, elements, caller) {
   for (const el of elements) {
     if (el.options.drawTime === caller) {
-      el.draw(ctx, parent);
+      el.draw(ctx);
     }
-  }
-}
-
-function drawSubElements(ctx, elements, caller) {
-  for (const el of elements) {
-    if (isArray(el.elements)) {
-      drawElements(ctx, el.elements, caller, el);
+    if (el.elements && el.elements.length) {
+      const box = 'getBoundingBox' in el ? el.getBoundingBox() : undefined;
+      for (const sub of el.elements) {
+        if (sub.options.drawTime === caller) {
+          sub.draw(ctx, box);
+        }
+      }
     }
   }
 }
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -164,12 +164,16 @@ function drawElements(ctx, elements, caller) {
     if (el.options.drawTime === caller) {
       el.draw(ctx);
     }
-    if (el.elements && el.elements.length) {
-      const box = 'getBoundingBox' in el ? el.getBoundingBox() : undefined;
-      for (const sub of el.elements) {
-        if (sub.options.drawTime === caller) {
-          sub.draw(ctx, box);
-        }
+    drawSubElements(ctx, el, caller);
+  }
+}
+
+function drawSubElements(ctx, el, caller) {
+  if (el.elements && el.elements.length) {
+    const box = 'getBoundingBox' in el ? el.getBoundingBox() : undefined;
+    for (const sub of el.elements) {
+      if (sub.options.drawTime === caller) {
+        sub.draw(ctx, box);
       }
     }
   }
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -160,10 +160,10 @@ function draw(chart, caller, clip) {
   });
 }
 
-function drawElements(ctx, elements, caller) {
+function drawElements(ctx, elements, caller, parent) {
   for (const el of elements) {
     if (el.options.drawTime === caller) {
-      el.draw(ctx);
+      el.draw(ctx, parent);
     }
   }
 }
@@ -171,7 +171,7 @@ function drawElements(ctx, elements, caller) {
 function drawSubElements(ctx, elements, caller) {
   for (const el of elements) {
     if (isArray(el.elements)) {
-      drawElements(ctx, el.elements, caller);
+      drawElements(ctx, el.elements, caller, el);
     }
   }
 }
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -164,17 +164,17 @@ function drawElements(ctx, elements, caller) {
     if (el.options.drawTime === caller) {
       el.draw(ctx);
     }
-    drawSubElements(ctx, el, caller);
+    if (el.elements && el.elements.length) {
+      drawSubElements(ctx, el, caller);
+    }
   }
 }
 
 function drawSubElements(ctx, el, caller) {
-  if (el.elements && el.elements.length) {
-    const box = 'getBoundingBox' in el ? el.getBoundingBox() : undefined;
-    for (const sub of el.elements) {
-      if (sub.options.drawTime === caller) {
-        sub.draw(ctx, box);
-      }
+  const box = 'getBoundingBox' in el ? el.getBoundingBox() : undefined;
+  for (const sub of el.elements) {
+    if (sub.options.drawTime === caller) {
+      sub.draw(ctx, box);
     }
   }
 }
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -135,17 +135,16 @@ export default {
 };
 
 function draw(chart, caller, clip) {
-  const {ctx, chartArea} = chart;
+  const {ctx, canvas, chartArea} = chart;
   const {visibleElements} = chartStates.get(chart);
+  let box = {x: 0, y: 0, width: canvas.width, height: canvas.height};
 
   if (clip) {
     clipArea(ctx, chartArea);
+    box = {x: chartArea.left, y: chartArea.top, width: chartArea.width, height: chartArea.height};
   }
 
-  const todraw = drawElements(ctx, visibleElements, caller);
-  for (const item of todraw) {
-    item.element.draw(ctx, item.box);
-  }
+  drawElements(ctx, visibleElements, caller, box);
 
   if (clip) {
     unclipArea(ctx);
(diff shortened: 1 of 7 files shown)


src/types/label.js
```diff
@@ -24,6 +24,13 @@ export default class LabelAnnotation extends Element {
     translate(ctx, this.getCenterPoint(), options.rotation);
     drawCallout(ctx, this);
     drawBox(ctx, this, options);
+    if (this.box) {
```
stockiNail, line 27: I needed to store the box annotation dimensions in order to apply to clip. I don't like it too much.

stockiNail: we cannot apply the clip to all annotations :(
Exceptions:
1. callout for LABEL annotation
2. arrowHeads for LINE annotation

Both above mentioned elements are NOT included in box model calculation.

src/types/box.js
```diff
@@ -109,31 +97,65 @@ BoxAnnotation.descriptors = {
...
+    y: y + halfBorder + padding.top,
+    width: width - borderWidth - padding.width,
+    height: height - borderWidth - padding.height
+  };
+}
+
+function resolveLabelElementProperties(chart, properties, options) {
+  const label = options.label;
+  label.backgroundColor = 'transparent';
+  label.callout.display = false;
```
stockiNail, line 141: added these properties settings to override whatever user configuration (box label should work without background and callout)

kurkle: Ref: https://github.com/chartjs/chartjs-plugin-annotation/pull/698#discussion_r815340691

The question is, do we want to (force) the box label to not have background or callout? there could be a use case for those too. I think setting the defaults would be enough, if user overrides, then fine!

src/types/label.js
kurkle: could pass chartArea to the parents?

kurkle: and then remove the outer clip, so annotations can draw outside chart area if needed...

src/types/box.js
stockiNail: Well, the box label wasn't thought to use callout and background color.
Let me think about... for callout I need to think about a point to link to the callout (could be the center of the label or top/left of the label). The box annotation (but the same for line) we don't have any option to locate a point (correctly).

src/types/label.js
stockiNail: I have added a method to box annotation which returns the bounding box.

stockiNail:
@kurkle the follwoing options are missing in the label of box annotation because the goal was to use the box rendering options.
But it makes sense to add the following missing options:

```
  backgroundColor
  backgroundShadowColor
  borderCapStyle
  borderDash
  borderDashOffset
  borderJoinStyle
  borderRadius
  borderShadowColor
  borderWidth
  shadowBlur
  shadowOffsetX
  shadowOffsetY
```

In my opionion this deserve a specific PR, with doc and new test cases.

About callout, I don't see any use case. The callout is thought to outline a point linking it to a label. I would like do not enable it in box annotation. What do you think?

test/fixtures/box/label-dynamic.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
                 content: 'This is dynamic!',
               },
               enter({element}) {
-                element.options.label.display = true;
+                element.elements[0].options.display = true;
```
stockiNail, line 34: @kurkle as reported https://github.com/chartjs/chartjs-plugin-annotation/pull/698#issuecomment-1054016151, to change the label options at runtime it's a little bit more complex.

src/types/label.js
```diff
@@ -15,15 +15,22 @@ export default class LabelAnnotation extends Element {
...
-    if (!options.content) {
+    if (!options.display || !options.content) {
       return;
     }
     ctx.save();
     translate(ctx, this.getCenterPoint(), options.rotation);
     drawCallout(ctx, this);
     drawBox(ctx, this, options);
+    if (parent && 'getBoundingBox' in parent) {
+      const {x, y, width, height} = parent.getBoundingBox();
```
kurkle, line 28: Why not just provide the bounding box (or the area where this element is allowed to draw to) directly?

This could potentially lead to detecting the type of parent and doing things based on that, which leads to bad maintainability.

src/annotation.js
```diff
@@ -160,18 +160,18 @@ function draw(chart, caller, clip) {
...
+      el.draw(ctx, parent);
     }
   }
 }
 
 function drawSubElements(ctx, elements, caller) {
   for (const el of elements) {
     if (isArray(el.elements)) {
-      drawElements(ctx, el.elements, caller);
+      drawElements(ctx, el.elements, caller, el);
```
kurkle, line 174: I'm not sure why I made it like this, where all of the sub elements are drawn after all of the annotations. If some of annotations overlap, all of the sub elements will be on top. This is another reason why I suggested the refactoring of the draw loop.

test/fixtures/box/label-dynamic.js
kurkle: We could help that use case by providing a `label` getter in box annotation
```js
get label() { return this.elements && this.elements[0]; }
```

kurkle:
> @kurkle the following options are missing in the label of box annotation because the goal was to use the box rendering options. But it makes sense to add the following missing options:
> 
> ```
>   backgroundColor
>   backgroundShadowColor
>   borderCapStyle
>   borderDash
>   borderDashOffset
>   borderJoinStyle
>   borderRadius
>   borderShadowColor
>   borderWidth
>   shadowBlur
>   shadowOffsetX
>   shadowOffsetY
> ```
> 
> In my opinion this deserves a specific PR, with doc and new test cases.
> 
> About callout, I don't see any use case. The callout is thought to outline a point linking it to a label. I would like do not enable it in box annotation. What do you think?

I don't see a use case for the callout either, but you never know what people come up with.
I think it the options should fall back to box options, but be overridable for the label. This would probably be easier if #630 was finished.

stockiNail:
> I think it the options should fall back to box options, but be overridable for the label. This would probably be easier if #630 was finished.

I agree for this reason I haven't started yet. All those options could be defined as commons.

stockiNail:
> get label() { return this.elements && this.elements[0]; }

Yes, a fast path. breaking change anyway!

src/annotation.js
stockiNail: > I'm not sure why I made it like this, where all of the sub elements are drawn after all of the annotations. If some of annotations overlap, all of the sub elements will be on top. This is another reason why I suggested the refactoring of the draw loop.

This is not 100% correct. Every sub-element has got own `drawTime` and therefore will be drawn based on that, like in the past. For instance, if the label is configured with `beforeDraw` and the element `afterDraw`, the element is drawn after the label.

stockiNail: See line `if (el.options.drawTime === caller) {`.

```javascript
function drawElements(ctx, elements, caller) {
  for (const el of elements) {
    if (el.options.drawTime === caller) {
      el.draw(ctx);
    }
  }
}
```

kurkle: Right. My analysis only applies if all have the same drawTime. But the behavior could be unexpected anyway. But this can be changed separately if needed.

src/types/label.js
stockiNail: > Why not just provide the bounding box (or the area where this element is allowed to draw to) directly?

I don't know the area where the sub element must be drawn in that point. The clip area, in this case, is NOT the element area but the element area with some adjustment (i.e. padding).

Furthermore, not all elements (and then annotations) needs the bounding box but in that point of code I can have only element instance (if we don't want to have additional structure to manage it). Therefore it should be a method of the element.
For sure we can add a `if` statement, checking if is a box annotation and then provide the bounding box only for that. In this case, but I could be wrong, it's less maintainable going to if statement.
In this way, the draw method can have also the "parent" (who is the owner of subelement) and can use for additional logic, like in this case. Maybe it could be needed for the label of the line as well.

Going to your suggestion, we have to check in the `annotation.js` if the element is a box one and then pass the bounding box. 
Or do you have another idea?

src/annotation.js
stockiNail: > My analysis only applies if all have the same drawTime

See issue #175. In my opionion we could add a `z` option where, when the `drawTime` are equals, you can define the z-index.
But, afaik, you would like to add this logic to all chart elements instead to apply only to this plugin.
Anyway we could get all visible elements and all their subelements, extracting by drawtime, and then draw them (1 cycle).

src/types/label.js
kurkle: No, but lets assume we have many elements that have many different kinds of sub-elements. Each of those elements that can be sub elements, need to check for `getBoundingBox` in parent vs it can be checked once in annotation.js

> In this way, the draw method can have also the "parent" (who is the owner of subelement) and can use for additional logic,

IMO this is just the kind of thing that leads to unmaintainable code, the element drawing starts to have branches based on the parent. It should be the parent that provides the options to the sub/child and child always does the same thing with the same options not knowing if it has a parent or not.

stockiNail: > No, but lets assume we have many elements that have many different kinds of sub-elements. Each of those elements that can be sub elements, need to check for `getBoundingBox` in parent vs it can be checked once in annotation.js

Ok, let me try to move it in the annotation.js.



> IMO this is just the kind of thing that leads to unmaintainable code, the element drawing starts to have branches based on the parent. It should be the parent that provides the options to the sub/child and child always does the same thing with the same options not knowing if it has a parent or not.

OK!!

src/annotation.js
kurkle: I think it can be done just by combining the two loops.

```js
for (const el of elements) {
  if (el.options.drawTime === caller) { 
    el.draw(ctx, chartArea); 
  }
  if (el.elements && el.elements.length) {
    const box = 'getBoundingBox' in el ? el.getBoundingBox() : chartArea;
    for (const sub of el.elements || []) {
      if (sub.options.drawTime === caller) { 
        sub.draw(ctx, box); 
      }
    }
  }
}

Comment (kurkle) [reply src/annotation.js:174]: To the z-index, we could add it here and adopt chart.js z-index when/if its available, but that should again be a separate pr.

**Comment by stockiNail** [comment]
> > get label() { return this.elements && this.elements[0]; }
> 
> Yes, a fast path. breaking change anyway!

done

File: src/annotation.js (hunk 5 above)
Comment (stockiNail) [reply src/annotation.js:174]: > I think it can be done just by combining the two loops.

I'm testing it, I have done it similar to your proposal.

Comment (stockiNail) [reply src/annotation.js:174]: > To the z-index, we could add it here and adopt chart.js z-index when/if its available, but that should again be a separate pr.

OK!

Comment (stockiNail) [reply src/annotation.js:174]: > I'm not sure why I made it like this, where all of the sub elements are drawn after all of the annotations.

I think you had done it to use recursively the drawElement function, that makes sense anyway.

Comment (stockiNail) [reply src/annotation.js:174]: I don't think we should use chartArea by default. The annotations could be set with `clip: true` and therefore the area is not correct, I guess. Maybe canvas area.

File: src/annotation.js [hunk 6]
```diff
@@ -165,13 +164,13 @@ function drawElements(ctx, elements, caller) {
...
     }
-  }
-}
-
-function drawSubElements(ctx, elements, caller) {
-  for (const el of elements) {
-    if (isArray(el.elements)) {
-      drawElements(ctx, el.elements, caller);
+    if (el.elements && el.elements.length) {
+      const box = 'getBoundingBox' in el ? el.getBoundingBox() : undefined;
```
stockiNail, line 168: @kurkle maybe you don't like it. I prefer to pass an undefined value instead of a default. Let me know.

src/annotation.js
```diff
@@ -165,13 +164,17 @@ function drawElements(ctx, elements, caller) {
...
   }
 }
 
-function drawSubElements(ctx, elements, caller) {
-  for (const el of elements) {
-    if (isArray(el.elements)) {
-      drawElements(ctx, el.elements, caller);
+function drawSubElements(ctx, el, caller) {
+  if (el.elements && el.elements.length) {
+    const box = 'getBoundingBox' in el ? el.getBoundingBox() : undefined;
```
stockiNail, line 173: @kurkle maybe you don't like it. I prefer to pass an undefined value instead of a default. Let me know.

src/annotation.js
```diff
@@ -142,8 +142,10 @@ function draw(chart, caller, clip) {
     clipArea(ctx, chartArea);
   }
 
-  drawElements(ctx, visibleElements, caller);
-  drawSubElements(ctx, visibleElements, caller);
+  const todraw = drawElements(ctx, visibleElements, caller);
+  for (const item of todraw) {
+    item.element.draw(ctx, item.box);
+  }
```
stockiNail, line 148: @kurkle I have changed the cycle to draw the elements in order to have an array with all elements (ans sub ones) to be drawn. This will enable also the z option in next PR. I can come back to previous one if you prefer

kurkle: Please try to separate things. It might be useful for z-index to create an array, but it has no use whatsoever in this PR.

src/annotation.js
kurkle: You are right, I don't like it. If we always passed an area (be it canvas area or chart area), it removes one `if` from all of the elements that need to do clipping (or whatever) based on the provided area.

src/annotation.js
stockiNail: ok, apologize

stockiNail: fall back to previous commit

kurkle (approved):

stockiNail:
@kurkle thanks for approval. What about CC issues?

src/annotation.js
stockiNail: > I'm not sure why I made it like this, where all of the sub elements are drawn after all of the annotations. If some of annotations overlap, all of the sub elements will be on top. This is another reason why I suggested the refactoring of the draw loop.

@kurkle coding the line label as element, maybe I have understood why you implemented in that way. With new cycle to draw elements and sub elements, the label of the one are drawn after every own line element, This have got a side effect that when the `drawTime` is the same, the labels are NOT on top anymore but they could be crosses by another annotation.
Before this update, the labels were always drawn after all elements and therefore the labels were never overlapped by any annotation. 
Now, I discovered, during the tests (https://github.com/chartjs/chartjs-plugin-annotation/blob/master/test/fixtures/line/label_drawTime.js), I see this new behavior.

PR #736, opened 2022-05-16
Move interaction object at top level of plugin options

stockiNail:
The `interaction` object has been moved at top level of plugin options, instead of common object as implemented by PR #630.

This is because the `interaction` is not configurable at annotation instance but is a plugin configuration, like `clip` one.


kurkle (approved):
Not sure if this has an impact on the resolution of options, but probably not,

PR #738, opened 2022-05-16
Add z option to all annotations

### LINKED ISSUE #175: Feature request: Annotation z-index

stockiNail:
Fix #175 

This PR is adding new option to all annotations and their label subelements in order to enable the user to define which annotations should be drawn on top, with the same `drawTime`.


src/annotation.js
```diff
@@ -150,25 +150,31 @@ function draw(chart, caller, clip) {
...
+    item.element.draw(chart.ctx, item.area);
+  }
 
   if (clip) {
     unclipArea(ctx);
   }
 }
 
-function drawElements(chart, elements, caller, area) {
+function drawElements(elements, caller, area) {
```
LeeLenaleee, line 164: Would rename the function to something like `getDrawableElements` since in its current implementation it does not do the actual drawing

docs/guide/types/box.md
```diff
@@ -92,6 +93,7 @@ If one of the axes does not match an axis in the chart, the box will take the en
 | `yMax` | Bottom edge of the box in units along the y axis.
 | `yMin` | Top edge of the box in units along the y axis.
 | `yScaleID` | ID of the Y scale to bind onto. If missing, the plugin will try to use the scale of the chart, configured as `'y'` axis. If more than one scale has been defined in the chart as `'y'` axis, the option is mandatory to select the right scale.
+| `z` | The `z` property determines the drawing stack level of the box annotation element. All visible elements will be drawn in crescent order of `z` option, with the same `drawTime` option.
```
LeeLenaleee, line 96: Don't think `cresent` is the correct wording, guess it should be `ascending`. Might be my lack of english but also when I putted it through translate it didn't give the results I would have thought.

Same comment for the other files

LeeLenaleee (changes requested):
2 small things

src/annotation.js
stockiNail: yeah, make a lot of sense

docs/guide/types/box.md
stockiNail: Using Google it is translated correctly but I agree to change in `ascending`

src/annotation.js
stockiNail: changed in getDrawableElements

docs/guide/types/box.md
stockiNail: Changed using `ascending`

stockiNail:
Changing the function name, we have re-enabled code climate issue about cognitive complexity. :worried:

docs/guide/types/box.md
LeeLenaleee: aah yeah seems I read over it, it was at bottom of verbs, cuz all other references google thought it could be where about the moon 😅

LeeLenaleee (dismissed):

stockiNail:
@LeeLenaleee apologize, I have tried to solve CC issue about complexity (I think better to have an exception as was before) and your approval is dismissed.

LeeLenaleee:
np

LeeLenaleee (approved):

LeeLenaleee:
Common MD file that gets imported would be better, would be more in line with chart.js itself and will make it more user frienndly because people will only go to the place they are using and prob won't see the common page

stockiNail:
> Common MD file that gets imported would be better, would be more in line with chart.js itself and will make it more user frienndly because people will only go to the place they are using and prob won't see the common page

OK! The content for the next PR is already decided! ;)

PR #740, opened 2022-05-19
Enable callout in the label of line annotation

stockiNail:
This PR is enabling the `callout` options in the label of line annotation.

### TODO

- [x] test cases
- [x] types
- [x] documentation
- [x] samples


stockiNail:
@kurkle I have a doubt about CC issue. We could need an object to create and add to element.defaults but where? Maybe a additional file "callout.js", which contains all functions to manage the callout?
Maybe this could be applied to the labels and arrowHeads (to reduce the size of the js files and close some CC issues). Just a proposal.

stockiNail:
@kurkle I solved CC duplication code issue using the `defaults` of the label annotation for `callout`.

kurkle:
> @kurkle I solved CC duplication code issue using the `defaults` of the label annotation for `callout`.

I think the `common` should now work for these shared defaults.

stockiNail:
> I think the `common` should now work for these shared defaults.

I wasn't sure. Let me try

stockiNail:
> > @kurkle I solved CC duplication code issue using the `defaults` of the label annotation for `callout`.
> 
> I think the `common` should now work for these shared defaults.

I have removed the callout options from label ( leaving ONLY the callout node) and adding them in the common, you don't have any fallback.
In my opinion, but maybe I'm wrong, when the options are loaded in the element, only defaults and defaultRoutes of the element are used (and not the common).

Therefore to work, we should maintain all callout options and setting to undefined to have the fallback but only to centralize the defaults, not the options.

kurkle (approved):

PR #742, opened 2022-05-19
Use margin option in the callout check for being shown

### LINKED ISSUE #741: Callout is shown even if box label and the margin is beyond the point
Currently the callout is not shown if the the point, which represents the annotation calculated location, is `inRange`.

The following picture shows how the `margin` is not considered in the check if the point is `inRange` or not (and it should) and then to draw callout or not.

Callout conf:

```javascript
callout: {
  display: true,
  margin: 20,
  position: 'bottom'
}
```

image

stockiNail:
Fix #741


kurkle:
Not sure what's up with coveralls, but 0% coverage is not acceptable :)

kurkle (approved):

PR #744, opened 2022-05-21
Add `beforeDraw` and `afterDraw` hooks to the annotations

### LINKED ISSUE #146: padding issue with drawing an average line

stockiNail:
This PR is adding before/after draw hooks in order to allow the user to apply whatever needed customization on out-of the-box annotations. Fix #146.

The hooks will be invoked once for every annotation for each drawing cycle.

```javascript
plugins: {
  annotation: {
    annotations: {
      line: {
        type: 'line',
        ....
        beforeDraw(context) {
          console.log('beforeDraw');
        },
        afterDraw(context) {
          console.log('afterDraw');
        },
      },
    }
  }
}
```

| Hook | Arguments | Return type | Default | Description
| :-- | :-- | :-- | :-- | :--
| beforeDraw | context | void | `annotation.beforeDraw` | invoked before the element drawing
| afterDraw | context | void | `annotation.afterDraw` | invoked after the element drawing

Common hooks for all annotations can be set in the annotation plugin options:

```javascript
plugins: {
  annotation: {
    beforeDraw(context) {
      console.log('Common beforeDraw');
    },
    afterDraw(context) {
      console.log('Common afterDraw');
    },
    annotations: {
      line: {
        type: 'line',
        ....
      },
    }
  }
}
```

### TODO

- [x] test cases
- [x] types
- [x] documentation
- [x] samples


kurkle (commented):
I still don't like the counters
How about making it simpler, just draw before/after the main element is drawn?
We could add `afterDrawChildren` later, if it's really needed.

stockiNail:
> I still don't like the counters
> How about making it simpler, just draw before/after the main element is drawn?
> We could add `afterDrawChildren` later, if it's really needed.

I think its reasonable 
But Let me take time to review the PR
My memory is not so persistent and I dont remember so well what I wrote

stockiNail:
> How about making it simpler, just draw before/after the main element is drawn?

done! review applied!

kurkle (approved):
Sorry for the huge delay on these. LGTM

PR #745, opened 2022-05-23
Enable fallback to common defaults

kurkle:
As reply to: https://github.com/chartjs/chartjs-plugin-annotation/pull/740#issuecomment-1134514169

I'm sure I missed some common options, but those can be added in subsequent pull requests.

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -1,11 +1,12 @@
 import {Chart} from 'chart.js';
-import {clipArea, unclipArea, isObject, isArray} from 'chart.js/helpers';
-import {handleEvent, hooks, updateListeners} from './events';
-import {adjustScaleRange, verifyScaleOptions} from './scale';
-import {updateElements, resolveType} from './elements';
-import {annotationTypes} from './types';
-import {requireVersion} from './helpers';
+import {clipArea, isArray, isObject, unclipArea} from 'chart.js/helpers';
 import {version} from '../package.json';
+import {commonDefaults} from './commonDefaults';
+import {resolveType, updateElements} from './elements';
+import {handleEvent, hooks, updateListeners} from './events';
+import {requireVersion} from './helpers';
+import {adjustScaleRange, verifyScaleOptions} from './scale';
+import {annotationTypes} from './types';
 
 const chartStates = new Map();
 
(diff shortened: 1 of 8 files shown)


stockiNail:
I was testing the same thing, passing the common to updateElements.

kurkle:
This is not really optimal, feel free to create better alternative:

| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/chartjs-plugin-annotation.esm.js` | 15.5 kB | +294 B (+2%) |
| `dist/chartjs-plugin-annotation.js` | 15.6 kB | +298 B (+2%) |
| `dist/chartjs-plugin-annotation.min.js` | 9.84 kB | +102 B (+1%) |

stockiNail:
> This is not really optimal, feel free to create better alternative:
> Filename 	Size 	Change
> `dist/chartjs-plugin-annotation.esm.js` 	15.5 kB 	+294 B (+2%)
> `dist/chartjs-plugin-annotation.js` 	15.6 kB 	+298 B (+2%)
> `dist/chartjs-plugin-annotation.min.js` 	9.84 kB 	+102 B (+1%)

I don't know how but I'm gonna think about. ;)

kurkle:
Maybe its tolerable now:

| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/chartjs-plugin-annotation.esm.js` | 15.4 kB | +246 B (+2%) |
| `dist/chartjs-plugin-annotation.js` | 15.5 kB | +251 B (+2%) |
| `dist/chartjs-plugin-annotation.min.js` | 9.8 kB | +67 B (+1%) |

Would need to do this somehow without `deepMerge` to make it smaller (and faster probably too)

stockiNail:
just thinking loud, could we use the proxies created by CHART,JS instead of the simple objects (which needs deep merge)?

kurkle:
> just thinking loud, could we use the proxies created by CHART,JS instead of the simple objects (which needs deep merge)?

I don't recall anymore why we are resolving the options to simple objects. But the proxy probably has a list of all the keys resolvable. Lets see.

stockiNail:
Having done some tests, this PR can address to centralized common options in the common but it doesn't solve the issues about duplication of code of callout (for instance).
This is because callout node is a subnode of label one (for line) instead for label is a subnode of root one, therefore fallback is not resolved anyway.

kurkle:
There are some issues still with the fallback definitions, I'll test some more to be sure what's wrong.

PR #749, opened 2022-05-24
Enable label on ellipse annotation

stockiNail:
This PR is enabling the labeling of the ellipse annotation.

### TODO

- [x] test cases
- [x] types
- [x] documentation
- [x] samples


stockiNail:
@kurkle I was thinking that maybe we could remove the bounding box to clip the label inside the box annotation, as did in this PR for ellipse. It would be a breaking change.

kurkle (approved):

PR #751, opened 2022-05-24
Add sample how to select annotations

stockiNail:
This PR is adding a sample to the doc to show how you can select an annotation.


stockiNail:
The sample was created to address the discussion #750.
How you can see, the events hooks are storing the status, about the selected elements, and changing the cursor on canvas, in "external" references. The  #750 discussion is proposing to have a share object in the context (the reference stored in the state of plugin) in order to share these additional info across the annotations (by event hooks).

LeeLenaleee (approved):

PR #752, opened 2022-05-25
Update line annotation diagram to describe new element structure

stockiNail:
Removes the wrong line label element diagram because not updated to new structure.


stockiNail:
> The line: The following diagram is showing the element properties about a `'line'` annotation: and the inner label:
> 
> Should now be: The following diagram is showing the element properties about a `'line'` annotation:
> 
> Can be found at line number 80

Yeah, forgot to remove it.

PR #754, opened 2022-05-30
Enable the line curve in the line annotation

stockiNail:
This PR is enabling the line curve representation by the line annotation.

curvesample

| Option | Type | Default | Description
| :-- | :-- | :-- | :--
| curve | boolean | false | if true, the line will be drawn as a curve
| controlPoint | string - number- {x: string - number, y: string - number}  | {y:'-50%'} | Distance (in px or percentage of the distance between the start and end point) from the center

It uses quadratic curve and `Path2D`.

### TODO 

- [x] review control point calculation 
- [x] test cases
- [x] types
- [x] documentation
- [x] sample


src/types/line.js
```diff
@@ -1,43 +1,58 @@
...
 
   inRange(mouseX, mouseY, axis, useFinalPosition) {
     const hBorderWidth = this.options.borderWidth / 2;
     if (axis !== 'x' && axis !== 'y') {
-      const epsilon = sqr(hBorderWidth);
       const point = {mouseX, mouseY};
+      const {path, ctx} = this;
+      if (path) {
+        setBorderStyle(ctx, this.options);
+        const result = ctx.isPointInStroke(path, mouseX, mouseY) || isOnLabel(this, point, useFinalPosition);
```
kurkle, line 25: I'm not 100% the mouseX/mouseY are canvas coordinates (when dpr !== 1), but they should be.

src/types/line.js
```diff
@@ -207,6 +206,11 @@ LineAnnotation.defaultRoutes = {
   borderColor: 'color'
 };
 
+function inYorXAxisRange(element, {mouseX, mouseY}, axis, {hBorderWidth, useFinalPosition}) {
```
kurkle, line 209:
```suggestion
function inAxisRange(element, {mouseX, mouseY}, axis, {hBorderWidth, useFinalPosition}) {
```

src/helpers/helpers.options.js
```diff
@@ -61,17 +61,18 @@ export function calculateTextAlignment(size, options) {
...
 
 /**
- * @param {LabelPositionObject|string} value
- * @returns {LabelPositionObject}
+ * @param {{x: number|string, y: number|string}|string|number} value
+ * @param {string|number} defaultValue
+ * @returns {{x: number|string, y: number|string}}
  */
-export function toPosition(value) {
+export function toPointOption(value, defaultValue) {
```
kurkle, line 68: to me the old name is better. could also have a default value for the defaultValue :)
```suggestion
export function toPosition(value, defaultValue = 'center') {
```

kurkle (commented):
couple of nits, I think this will be a very nice addition!

src/types/line.js
stockiNail: Apologize @kurkle but I don't understand. What  does `(when dpr !== 1),` mean?

kurkle: DevicePixelRatio, eg. Retina display

stockiNail: Ok, forgive my lack of knowledge... I can do some tests with different dpr and to see if it works. I'll check also the documentation of the `Path` to see if the topic is mentioned. But do you have any experience with `Path`s where dpr doesn't work?

stockiNail: @kurkle I don't know if I did a thorough test, changing Karma config, and enabling DPR to 2

```javascript
    customLaunchers: {
      chrome: {
        base: 'Chrome',
        flags: [
          '--disable-accelerated-2d-canvas',
          '--disable-background-timer-throttling',
          '--disable-backgrounding-occluded-windows',
          '--disable-renderer-backgrounding'
        ]
      },
      firefox: {
        base: 'Firefox',
        prefs: {
          'layers.acceleration.disabled': true,
          'layout.css.devPixelsPerPx': 2.0 // <--- change pixel ratio
        }
      }
    },
```

and running the test on FF (`npm run dev:ff`), no issue at all.

I don't see any other way to test it :(.

What do u think?

kurkle: Not sure if that works, but I think quickest test is to run a sample page in browser, zoom in and see if it works properly

stockiNail: thanks! I tested and you're right. It doesn't work. uhm.... without Path2D it's very difficult to implement a curve... Let me investigate or do you know any trick about that?

stockiNail: Found it!

stockiNail: image

kurkle: Should not be that simple, was you chart at the top left corner of the window? :)

kurkle: https://github.com/chartjs/Chart.js/blob/52cf8e8a94130d63b7618751fe0df871ca1af222/src/helpers/helpers.dom.js#L94-L123 that should do it

stockiNail: WOW!!! Great! I wasn't aware the getRelativePosition is also managing the DPR. Every time I'm learning some thing new with you!!!!! UR awesome! I really appreciate. I ca tell you only thank youu! 
But coming back to the topic, I need to find a way to get the chart, because is needed to getRelativePosition and I don't have it in inRange method... let me try... I'll find a way (that maybe you don't like it :) )

stockiNail: I have it, in the element context!!!

stockiNail: Uhm... it doesn't work. Inside the inRange I'm calling getRelativePosition but, tracking also in Chart.js, the original event was already "translated" by  getRelativePosition . Therefore applying it again, I'm quite sure the new {x, y} is not correct.

stockiNail: Code:

```javascript
        const {chart} = this.$context;
        const {x, y} = getRelativePosition({clientX: mouseX, clientY: mouseY}, chart);
        const result = ctx.isPointInStroke(path, x, y) || isOnLabel(this, point, useFinalPosition);
```

Instead, it's continuing working multiplying the mouseX and mouseY to DPR..

stockiNail: The original event: `clientX: 369, clientY: 71`
In range argument: `mouseX: 361,mouseY: 63` (after getRelativePosition invoked in CHART.js)
after my call to getRelativePosition: `x: 353,y: 55`

mouseX, mouseY multiplied by DPR: `mx: 481.3333333333333, my: 84, dpr: 1.3333333333333333` (it works)

stockiNail: Maybe I'm wrong but getRelativePosition is dividing by DPR (and this is what we are receiving as mouseX and mouseY). Therefore to come back the original position, we should multiply by dpr.... but, repeating myself, maybe I'm wrong.

kurkle: hmm, ok. Maybe your original solution is correct then. There must be something different in `Path`

stockiNail: yes, I think so :(

kurkle (dismissed):

kurkle:
@stockiNail I'm not merging these, so you can decide the order you want to go.

kurkle (approved):

PR #761, opened 2022-06-08
Added usage information

DarkTrick:
- import and register example for react
- fully functional example for react and typescript

### DIFF
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -2,7 +2,7 @@
 
 :::danger
 
-This plugin needs to be registered. It does not function as inline plugin.
+This plugin needs to be registered. It does not function as inline plugin (see below).
 
 :::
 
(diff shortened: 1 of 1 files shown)


DarkTrick:
Currently it's very difficult to get things working if you're a newcomer. This change is trying to improve that.
As I'm using react + typescript I could only add examples for this environment. Hopefully more examples will follow.

LeeLenaleee:
It is already documented [here](https://www.chartjs.org/chartjs-plugin-annotation/latest/guide/integration.html#bundlers-webpack-rollup-etc) how to integrate the annotation plugin within chart.js.

This information is for the zoom, annotation, deferred and datalabels plugin in the same place as it is within chart.js itself.

Also I don't think that different frameworks should all be documented. Since they all fall under the category of bundlers. Only when there is a tool that does it totally different should it be added in my oppinion

docs/guide/index.md
```diff
@@ -24,3 +24,71 @@ For Chart.js 2.4.0 to 2.9.x support, use [version 0.5.7 of this plugin](https://
...
+          },
+        ],
+      },
+    },
+  };
+
+   const data2 = {
+    labels:   [ 'a', 'b'],
+    datasets: [ { data: [1, 2] } ],
+   };
```
stockiNail, line 88: Indentation of 2 spaces, not 3

docs/guide/index.md
```diff
@@ -24,3 +24,71 @@ For Chart.js 2.4.0 to 2.9.x support, use [version 0.5.7 of this plugin](https://
...
+          {
+            id: 'a-line-1',
+            type: 'line' as const, // important, otherwise typescript complains
+            mode: 'horizontal',
+            scaleID: 'y',
+            value: 1.0,
+            borderColor: 'red',
+            borderWidth: 4,
+            label: {
+              enabled: true,
```
stockiNail, line 76: Going to version 2, `enabled` is not longer supported. `display` must be used.

docs/guide/index.md
```diff
@@ -24,3 +24,71 @@ For Chart.js 2.4.0 to 2.9.x support, use [version 0.5.7 of this plugin](https://
...
+      y: {} // <- "y" will be used as "scaleID" in annotations
+    }
+    plugins: {
+      legend: {
+        display: false,
+      },
+      annotation: {
+        annotations: [
+          {
+            id: 'a-line-1',
```
stockiNail, line 68: Not need. If yes, maybe it's better to use `annotations` as object, not array

stockiNail (commented):
I have the feeling that the sample is not based on "master" implementation because is using obsoleted options. Maybe I'm wrong.

LeeLenaleee:
@stockiNail as I described in my first comment do we want this, I have the feeling that current way of documentation is better. Since if we start going this route you could argue you need at least 3 react samples (1 functional component, 1 class and 1 with create element). While all the frameworks are basically the same with importing and that is well documented on the integration page.

stockiNail:
> as I described in my first comment do we want this, I have the feeling that current way of documentation is better. Since if we start going this route you could argue you need at least 3 react samples (1 functional component, 1 class and 1 with create element). While all the frameworks are basically the same with importing and that is well documented on the integration page.

I have applied the review anyway! 
But I agree with you. The risk is to have samples for every JS frameworks.

LeeLenaleee:
Closing this since the neccesarry information can be found in the [integration page](https://www.chartjs.org/chartjs-plugin-annotation/latest/guide/integration.html). 

Way of importing for frameworks is at the moment to my knowledge all the same and falls under the bundler category. 
If there is a framework that works in a different way we can add it.

But replicating the bunder import for all the used frameworks is something I don't see as usefull and only makes the docs too full so people will skip even more through it.

DarkTrick:
> It is already documented [here](https://www.chartjs.org/chartjs-plugin-annotation/latest/guide/integration.html#bundlers-webpack-rollup-etc) how to integrate the annotation plugin within chart.js.

This might indeed be already enough. Perhaps I'm viewing the website in a wrong way, but I feel this information is quite hidden (the tiny link on the bottom right). I would expect that information to be within the *getting started* section.

What would you think about merging ["getting started"](https://www.chartjs.org/chartjs-plugin-annotation/latest/guide/), ["integration"](https://www.chartjs.org/chartjs-plugin-annotation/latest/guide/integration.html#bundlers-webpack-rollup-etc) and ["usage"](https://www.chartjs.org/chartjs-plugin-annotation/latest/guide/usage.html) into one bigger "getting started" page.

PR #762, opened 2022-06-08
Add external links to release note/drafter

stockiNail:
Adds to release note/drafter the following links:

- documentation
- migration guide (v2)
- samples

to keep it aligned to Chart.js format.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -42,6 +42,9 @@ template: |
   # Essential Links
 
   * [npm](https://www.npmjs.com/package/chartjs-plugin-annotation)
+  * [Docs](https://www.chartjs.org/chartjs-plugin-annotation/)
+  * [Migration guide](https://www.chartjs.org/chartjs-plugin-annotation/latest/guide/migrationV2.html)
+  * [Samples](https://www.chartjs.org/chartjs-plugin-annotation/latest/samples/intro.html)
 
   $CHANGES
 
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -42,8 +42,8 @@ template: |
   # Essential Links
 
   * [npm](https://www.npmjs.com/package/chartjs-plugin-annotation)
-  * [Docs](https://www.chartjs.org/chartjs-plugin-annotation/)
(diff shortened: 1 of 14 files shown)


.github/release-drafter.yml
```diff
@@ -42,6 +42,9 @@ template: |
   # Essential Links
 
   * [npm](https://www.npmjs.com/package/chartjs-plugin-annotation)
+  * [Docs](https://www.chartjs.org/chartjs-plugin-annotation/)
+  * [Migration guide](https://www.chartjs.org/chartjs-plugin-annotation/latest/guide/migrationV2.html)
+  * [Samples](https://www.chartjs.org/chartjs-plugin-annotation/latest/samples/intro.html)
```
LeeLenaleee, line 47:
```suggestion
  * [Migration guide](https://www.chartjs.org/chartjs-plugin-annotation/latest/guide/migrationV2.html)
  * [Docs](https://www.chartjs.org/chartjs-plugin-annotation/)
  * [Samples](https://www.chartjs.org/chartjs-plugin-annotation/latest/samples/intro.html)
```

LeeLenaleee:
np, np :) 
Maby we should also add typedoc to the documentation. Then we have all the links chart.js also has in its release drafter. But that is something for another PR and low prio

LeeLenaleee (approved):

stockiNail:
> np, np :) Maby we should also add typedoc to the documentation. Then we have all the links chart.js also has in its release drafter. But that is something for another PR and low prio

@LeeLenaleee Yes, I agree 100%.

Off topic: I was thinking that maybe we should release version 2. in this https://github.com/chartjs/chartjs-plugin-annotation/pull/755#issuecomment-1146187396 I have shared my standpoint about missing things.

With this PR, remains:

1. [#749](https://github.com/chartjs/chartjs-plugin-annotation/pull/749#issuecomment-1136057177 ) raises a doubt about the clip on box for label. If the proposal will be accepted, we have a breaking change and therefore another PR to add it in the migration guide is required
2. currently events options, commons for all annotation, are managed at root level of plugin options. Now with `common` node, maybe it makes sense to move over there, like all other annotation options. This would be a breaking change.

Item 2 can be ignored and we will see in version 3. 

What do you think?

PR #763, opened 2022-06-08
Bump version to 2.0.0

stockiNail:
As https://github.com/chartjs/chartjs-plugin-annotation/pull/755#issuecomment-1146187396 I see still 1 open point and a decision is needed:

1. [#749](https://github.com/chartjs/chartjs-plugin-annotation/pull/749#issuecomment-1136057177 ) raises a doubt about the clip on box for label. If the proposal will be accepted, we have a breaking change and therefore another PR to add it in the migration guide is required

I vote to remove the clip (breaking change) in order to have the same label management on box, ellipse, polygon (not yet prepared).

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -42,6 +42,9 @@ template: |
   # Essential Links
 
   * [npm](https://www.npmjs.com/package/chartjs-plugin-annotation)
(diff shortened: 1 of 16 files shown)


stockiNail:
Closed to avoid conflict with #771. It will be receated

PR #769, opened 2022-07-09
Change display to enabled at label options

rafaelmarreca:
The property display doesn't exists, the correct property is enabled


LeeLenaleee:
Current docs are correct. The docs in master are for the upcomming V2 release and not the current V1. In [the migration guide](https://www.chartjs.org/chartjs-plugin-annotation/master/guide/migrationV2.html#options) thats also in master this is stated: 

> enabled option was replaced by display in the callout configuration of label annotation, in the label configuration of line and box annotations and in the arrow heads configuration of line annotation to have the same option on all elements.

PR #771, opened 2022-07-20
Add API docs

LeeLenaleee:


stockiNail:
@LeeLenaleee just for my knowledge. Do you think we should enrich in the future the type defs with the explanation of classes, properties and methods? Any guideline?

stockiNail (approved):

LeeLenaleee:
I think the better choice would be to switch to ts all together since the plugin is not that big it might be doable, tried switching chart.js helpers but that grew painfully real fast

PR #808, opened 2022-11-15
Change CHART.JS dependency to version 4

### LINKED ISSUE #807: Update Peer dependency
We need to check if everything works fine with V4 and then update the peer dependency

stockiNail:
Fix #807


stockiNail:
Thank you @kurkle for the example to fix docs build issue.
The version in package.json has been set to 2.1.0 if we want to create new version. If yes, I'll change all other PR milestones.

LeeLenaleee (approved):

test/fixtures/box/noDefaultScales.js
```diff
@@ -3,7 +3,8 @@ module.exports = {
     type: 'bar',
     data: {
       datasets: [{
-        data: [0, 5, 10, 15, 20, 22]
+        data: [0, 5, 10, 15, 20, 22],
+        yAxisID: 'y1'
```
kurkle, line 7: Maybe these test are an unsupported use case with v4? If so, maybe just remove the tests.

kurkle (commented):
Looks good, except if I understand the "noDefaultScales" tests right and those are useless tests now?

test/fixtures/box/noDefaultScales.js
stockiNail: Not sure because I think it's still valid. Let me do a test and then I'll come back to you. My doubt it' related to index axis.

stockiNail: Done some tests and these tests could be removed. But with these tests we can tests that the `retrieveScaleID` function is working well with no defaults scale ids as well. Let me know!

stockiNail: I think I can try to move the tests to a unit tests (no fixture) even if I have to think a little about what to test.

kurkle: For me removing the tests from this pr and adding new tests later for `retrieveScaleID` is ok.

stockiNail: I have already done. Removed noDeault fixture, added retrieveScaleID test cases in helpers spec.

stockiNail: Have a look if looks good

stockiNail: It's better to acquire a chart and use the scales of the chart instead of a custom scales structure

LeeLenaleee (approved):

kurkle (approved):
Actually for unit tests its better to ise mocked data, so the situation does not change when chart.js changes. But I think this is fine and we can call it integration or functional test :)

stockiNail:
Ok @kurkle 
I did in that way the first impl but I thought you preferred this one.
New version? Bump is already here 
If yes I can do it later

PR #815, opened 2022-12-05
Update packaging to match main chart.js library

### LINKED ISSUE #814: Server Error when using annotation plugin with Chart.js 4.0.1 in NextJS app
Hey guys,

First of all, thanks for making this super helpful library!

I am currently working on a NextJS and have implemented ChartJS v4.0.1. All works well, but when I want to install the annotation plugin and follow the steps of https://www.chartjs.org/chartjs-plugin-annotation/latest/guide/integration.html I get the following error:
Schermafbeelding 2022-12-01 om 11 25 58

My React component is: (only relevant is shown)
```
import {useEffect, useState} from "react";
import {BarElement, CategoryScale, Chart as ChartJS, Legend, LinearScale, Title, Tooltip} from 'chart.js';
import Annotation from 'chartjs-plugin-annotation';
import {Bar} from 'react-chartjs-2';

ChartJS.register(
    CategoryScale,
    LinearScale,
    BarElement,
    Title,
    Tooltip,
    Legend,
    Annotation,
);

.........
```

Do you guys have any idea how to solve this? Is it an issue because of ChartJS v4?

Thanks a lot!

Cheers,
Ferre

stockiNail:
Fix #814


karma.conf.cjs
```diff
@@ -15,7 +16,7 @@ module.exports = function(karma) {
   // we will prefer the unminified build which is easier to browse and works
   // better with source mapping. In other cases, pick the minified build to
   // make sure that the minification process (terser) doesn't break anything.
-  const regex = karma.autoWatch ? /chartjs-plugin-annotation\.js$/ : /chartjs-plugin-annotation\.min\.js$/;
+  const regex = karma.autoWatch ? /chartjs-plugin-annotation\.umd\.js$/ : /chartjs-plugin-annotation\.umd\.min\.js$/;
```
kurkle, line 19: I'm not sure if we should repeat this breaking change?

kurkle (commented):
Looks ok, but donwe really want to rename the umd files?

karma.conf.cjs
stockiNail: But I'm not so expert on these things (JS packaging). Let me know.

LeeLenaleee: the esm already has an `esm` part in the name so we are not using `chartjs-plugin-annotation.js` right now with this change as far as I can see so it should be usable instead of adding the umd syntax 🤔

kurkle: It is just a file name.
@etimberg what do you think, should we make the org repositories consistent?

stockiNail: @kurkle @LeeLenaleee @etimberg I would suggest do not add UMD in order to avoid a breaking change. When we will define if it must contain UMD, a major version will be bump.
Wdyt?

stockiNail: I can not make any change now because I am reinstalling my laptop right now. I am going to do asap

stockiNail: I had a look to chartjs-plugin-gradient, what @kurkle did. I'll do the same.

stockiNail: done

stockiNail:
@kurkle I wen back to the original version, without using UMD in the dist files, as shared. Let me know if it's ok and if yes, let me know if you think we should publish new version 2.1.1 (bumping).

package.json
```diff
@@ -5,8 +5,15 @@
   "version": "2.1.0",
   "author": "Evert Timberg <evert.timberg@gmail.com>",
   "license": "MIT",
-  "main": "dist/chartjs-plugin-annotation.js",
-  "module": "dist/chartjs-plugin-annotation.esm.js",
+  "type": "module",
+  "main": "dist/chartjs-plugin-annotation.esm.js",
```
kurkle, line 9: If we want to support chart.js v3 on node/commonjs with these later annotation plugin versions, this needs to be umd build with .cjs extension

Something like I did in autocolors: https://github.com/kurkle/chartjs-plugin-autocolors/blob/main/package.json

kurkle (approved):
Looks good to me, except it won't work in node/commonjs.

@etimberg is it ok for you to drop cjs support for plugins too? (main repo already did)

etimberg:
> Looks good to me, except it won't work in node/commonjs.
> 
> @etimberg is it ok for you to drop cjs support for plugins too? (main repo already did)

Dropping is fine by me, but should definitely be a major version release

package.json
stockiNail: Ok Let me update in next hours

stockiNail:
> > Looks good to me, except it won't work in node/commonjs.
> > @etimberg is it ok for you to drop cjs support for plugins too? (main repo already did)
> 
> Dropping is fine by me, but should definitely be a major version release

@etimberg ok that means all plugins/controllers in chartjs org must publish a major version for chartjs 4 compatibility, doesn’t it?

package.json
stockiNail: @kurkle updated as suggested

etimberg:
> > > Looks good to me, except it won't work in node/commonjs.
> > > @etimberg is it ok for you to drop cjs support for plugins too? (main repo already did)
> > 
> > 
> > Dropping is fine by me, but should definitely be a major version release
> 
> @etimberg ok that means all plugins/controllers in chartjs org must publish a major version for chartjs 4 compatibility, doesn’t it?

I don't think it's mandatory if the plugin packaging doesn't change

stockiNail:
@etimberg @kurkle with CHART.JS 4.1.1 this PR sounds to me useless because the issue should be fixed.

kurkle:
There could be some benefits to from `type: 'module'`, but maybe nothing worth the time before someone requests something explicit.

PR #822, opened 2022-12-22
Bump to version 2.1.1 

stockiNail:
Update dependencies


stockiNail:
@LeeLenaleee @kurkle what do you think if I will change this PR to bump new version 2.1.1? If yes I will do this evening

kurkle:
> @LeeLenaleee @kurkle what do you think if I will change this PR to bump new version 2.1.1? If yes I will do this evening

Sounds good to me, quick releases is always better.
Maybe it should still include the destroy hook for v3 compatibility though?

stockiNail:
@kurkle @LeeLenaleee not needed because the plugin has the dependency with chartjs 3.7.0 and afterDestroy has been introduced in version 3.7.0

stockiNail:
> Sounds good to me, quick releases is always better.

I have changed version and title of this PR.

kurkle (approved):

LeeLenaleee (approved):

PR #824, opened 2022-12-28
Remove destroy plugin hook and replaced with afterDestroy

stockiNail:
With CHART.JS 4, PR https://github.com/chartjs/Chart.js/pull/10549, the `destroy` plugin hook has been removed and replaced with `afterDestroy`.
Currently, the version 2.1.0 is not cleaning the states map.

### DIFF
--- a/src/annotation.js
+++ b/src/annotation.js
@@ -92,7 +92,7 @@ export default {
     }
   },
 
-  destroy(chart) {
+  afterDestroy(chart) {
     chartStates.delete(chart);
   },
 


LeeLenaleee (approved):

src/annotation.js
```diff
@@ -92,7 +92,7 @@ export default {
     }
   },
 
-  destroy(chart) {
+  afterDestroy(chart) {
```
kurkle, line 95: As we have not made a major version jump yet and thus should be compatible with Chart.js v3, maybe we should just keep both for now=

LeeLenaleee: Aah yes totally forgot about that 😓

stockiNail: @kurkle @LeeLenaleee not needed because the plugin has the dependency with chartjs 3.7.0 and afterDestroy has been introduced in version 3.7.0

PR #827, opened 2023-01-04
Upgrade dependencies and implementation of the integration test

stockiNail:
This PR is changing the dependency with CHART.JS, going to version4. It's also adding the test on TS 4.7.x

### DIFF
--- a/test/integration/integration-test.js
+++ b/test/integration/integration-test.js
@@ -33,10 +33,12 @@ function getChartJsVersions(projectPath) {
 describe('Integration Tests', () => {
   const distDir = path.resolve('./');
   const chartjsVersions = getChartJsVersions(distDir);
+  const baseTmpDir = path.join(os.tmpdir(), 'chartjs-plugin-annotation-tmp');
+  fs.rmSync(baseTmpDir, {recursive: true, force: true});
+  fs.mkdirSync(baseTmpDir);
+
   chartjsVersions.forEach(function(chartjs) {
-    const baseTmpDir = path.join(os.tmpdir(), 'chartjs-plugin-annotation-tmp');
     const tmpDir = path.join(baseTmpDir, chartjs);
-    fs.rmSync(tmpDir, {recursive: true, force: true});
     fs.mkdirSync(tmpDir);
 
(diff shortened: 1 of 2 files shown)


kurkle (commented):
Maybe it should ve testing both?

stockiNail:
> Maybe it should ve testing both?

Chart.js 3 and 4?

kurkle:
Yes, or was v3 support already droped?

stockiNail:
V3 should be supported, version 3.7.0

stockiNail:
I will work tomorrow on that. I think we need a package.json template to update at runtime, settimg v3 and then v4. I need to see how the fs api in js are working

kurkle:
Its already running for each of the ts versions, should not be hard to add another loop for that. Or just split to separate folders (and maybe test only some of the later ts versions with v4)

stockiNail:
> Its already running for each of the ts versions, should not be hard to add another loop for that. Or just split to separate folders (and maybe test only some of the later ts versions with v4)

@kurkle I have done! Have a look if ok.
Let me say that the implementation is getting the devDep and peerDep of the plugin to chart.js and based on that it calculates the amount of majors between them. And then, configure the package.json for testing to use the last version in the major, by [major].x.x. 
In this way we don't have to change anything next time, when the plugin will change the dependencies to Chart.js

stockiNail:
Must be rebased when #838 will be merged. Now in draft

kurkle:
Many lines of code to avoid adding a chart.js major once a year or two? :)
Less maintenance is less maintenance, so why not.

kurkle:
Another thought, maybe this one should be converted to pnpm too, so it could use workspace dependencies and avoid the temporary directory stuff. Like its done in chart.js now.

stockiNail:
> Many lines of code to avoid adding a chart.js major once a year or two? :)
> Less maintenance is less maintenance, so why not.

I could agree. For my mindset, I'm never afraid about maintenance or LoC ;) 

> Another thought, maybe this one should be converted to pnpm too, so it could use workspace dependencies and avoid the temporary directory stuff. Like its done in chart.js now.

OK! for this one, I need more time to have a look how it works with more details that I have got now.

stockiNail:
Close for now

Issue #829, opened 2023-01-05
Wrong documentation about label annotation and its positioning


stockiNail:
https://github.com/chartjs/chartjs-plugin-annotation/issues/828#issue-1520832754

stockiNail:
Close by https://github.com/chartjs/chartjs-plugin-annotation/pull/830

PR #830, opened 2023-01-05
Fix documentation about label annotation and its positioning

### LINKED ISSUE #828: Width for Label annotation doesn't work
I want `label` annotation to take the whole width of the axis, but for some reason, `xMin` and `xMax` values are ignored.
I also tried to use `width` parameter, but it doesn't have any effect either.

In the docs, [there's a line](https://www.chartjs.org/chartjs-plugin-annotation/latest/guide/types/label.html#general) saying that _"To enable to use the box positioning, the `radius` must be set to `0` or `NaN`."_, but could you specify what `radius` it refers to? I wasn't able to find anything for both this plugin or ChartJS itself.

In the example below I used 2 annotations - `box` that can take the whole width of the axis and `label` that doesn't work.

Thank you!

https://jsfiddle.net/ntyv2cdb/17/

stockiNail:
Fix #828 #829

### DIFF
--- a/docs/guide/types/label.md
+++ b/docs/guide/types/label.md
@@ -83,7 +83,7 @@ The following options are available for label annotations.
 
 If one of the axes does not match an axis in the chart, the content will be rendered in the center of the chart. The 2 coordinates, xValue, yValue are optional. If not specified, the content will be rendered in the center of the scale dimension.
 
-The 4 coordinates, xMin, xMax, yMin, yMax are optional. If not specified, the box is expanded out to the edges in the respective direction and the box size is used to calculated the center of the point. To enable to use the box positioning, the `radius` must be set to `0` or `NaN`.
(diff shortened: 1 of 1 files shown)


LeeLenaleee (approved):

LeeLenaleee:
2.1.2 since there are no enhancements

stockiNail:
> 2.1.2 since there are no enhancements, but I dont really see a need for a bump 🤔

Do you know if release-drafter will change the draft release that it created if in the next days a PR with a enhancement will be merged? I don't release-drafter so well :(

LeeLenaleee:
Yes if a enhancement gets merged it will since the enhancement will also be released

stockiNail:
> Yes if a enhancement gets merged it will since the enhancement will also be released

THANK YOUU!

Issue #786, opened 2022-08-15
Cannot set properties of undefined (setting 'display') at resolveLabelElementProperties


dbauszus-glx:
We seem to be stuck with plugin-annotation v1.3.0 and chartjs v3.7.0 where following example for a line chart with multiple datasets works with multiple annotations.

https://codepen.io/dbauszus-glx/pen/MWOjGzw

Upgrading plugin-annotation to v2.0.0 causes following TypeError: `Cannot set properties of undefined (setting 'display') at resolveLabelElementProperties`.

Upgrading chartjs to v3.9.1 causes a different TypeError: `Cannot read properties of undefined (reading 'lastIndexOf')`

The same error persists with chartjs on v3.9.1 and plugin-aanotation on v2.0.0

stockiNail:
@dbauszus-glx having a look to the codepen, I have seen errors on chart configuration (`responsive` and `tooltip` options were in wrong position). 
In the annotation plugin, you are suing `enabled` option which has been remove in version 2.0.0 (see [migration guide](https://www.chartjs.org/chartjs-plugin-annotation/latest/guide/migrationV2.html#options)), where `display` option is used.

I have created a codepen using you code (removing async load of JS, and usinf chart,js 3.9.1 and annotation 2.0.0): https://codepen.io/stockinail/pen/PoRyMpv
It seems working.

The issue seems to be related to async import and registration of plugin.

dbauszus-glx:
Interesting. Thanks for pointing out the differences in the migration guide.

I can see that async import from https://cdn.skypack.dev/chartjs-plugin-annotation@2.0.0 fails because the options argument is a proxy with callout.display undefined. 

```js
function resolveLabelElementProperties(chart, properties, options) {
  options.callout.display = false;

```

Is the use of a proxy new in version 2 or can you think of any other changes which may cause the registration to fail from a dynamic import?

stockiNail:
@dbauszus-glx the problem is not related to the undefined `callout` because otherwise it wouldn't work with importing javascript by `<script>` tag. And the issue is a missing registrations of annotation elements and their defaults (for this reason is undefined).

Anyway, I think I found the issue.
There is something weird  when we are importing chart.js 3.9.1 from `cdn.skypack.dev`.

This is the imported file from skypack:

```javascript
*
 * Skypack CDN - chart.js@3.9.1
 *
 * Learn more:
 *   📙 Package Documentation: https://www.skypack.dev/view/chart.js
[...]

PR #826, opened 2023-01-04
Fix links with anchors adding md extension in the documentation

stockiNail:
Fix #818 

This PR is also removing the `autocolors` configuration plugin from the guide.

PR #832, opened 2023-01-09
Add exports to manage "import" and "require" inclusions

### LINKED ISSUE #83: Lines working in examples; but not on live code
Hello there,

I'm having a very strange issue where I am not able to draw any lines on any of my charts with our live code.

If I [run one of the examples](http://i.imgur.com/CVCtcqp.png) on its own, I can see the lines with no issues. 

However, if I copy and paste the exact same code onto my site, the [same chart will not display the lines](http://i.imgur.com/sDvXomL.png). 

I have ensured that I'm using exactly the same dependencies in both places (to the point where I copied the same files to their respective /js folders). I've ensured that we have no CSS touching any Canvas elements, so I am truly at a loss for why this is not working. 

Any tips for what else I should be looking at? Should I provide any code snippets?
Thank you!

stockiNail:
Should fix #83. @kurkle @LeeLenaleee can you confirm? My experience on packaging is not so huge.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,14 @@
   "main": "dist/chartjs-plugin-annotation.js",
   "module": "dist/chartjs-plugin-annotation.esm.js",
   "types": "types/index.d.ts",
+  "jsdelivr": "dist/chartjs-plugin-annotation.min.js",
+  "unpkg": "dist/chartjs-plugin-annotation.min.js",
+  "exports": {
+    "types": "./types/index.d.ts",
+    "import": "./dist/chartjs-plugin-annotation.esm.js",
+    "require": "./dist/chartjs-plugin-annotation.min.js",
+    "script": "./dist/chartjs-plugin-annotation.min.js"
+  },
   "repository": {
     "type": "git",
     "url": "https://github.com/chartjs/chartjs-plugin-annotation.git"


package.json
```diff
@@ -8,6 +8,14 @@
   "main": "dist/chartjs-plugin-annotation.js",
   "module": "dist/chartjs-plugin-annotation.esm.js",
   "types": "types/index.d.ts",
+  "jsdelivr": "dist/chartjs-plugin-annotation.min.js",
+  "unpkg": "dist/chartjs-plugin-annotation.min.js",
+  "exports": {
+    "types": "./types/index.d.ts",
```
LeeLenaleee, line 14: small thing, we don't specify it like this anywhere in the package.json, in my oppinion we should keep in consistent, so either here we need to remove the `./` or add it everywhere else (has my preference)

kurkle (approved):
looks ok to me

package.json
kurkle: in exports, it needs to start with a dot where referencing a file. it could be a package too.

LeeLenaleee (approved):

package.json
stockiNail: Apoligize... I don't understand if it's correct.... Im CHART.,JS there is the dot:

https://github.com/chartjs/Chart.js/blob/b51c52336624a24fe1ec814dc47f9a4cb7dd6a8a/package.json#L16-L32

stockiNail:
bug? version 2.1.2 or can it wait minor?

stockiNail:
if we go to 2.1.2 (in my opinion make sense), can we have a look to PR #826 in order to fix the doc as well?

kurkle:
212 seems correct to me

robbytx:
This PR appears to resolve #786 for me, as I've just described in a comment on that issue.

PR #838, opened 2023-01-25
Set the plugin as module

### LINKED ISSUE #837: import from chart.js cause "cannot use import statement outside a module"
Hello,
When I try to import annotation plugin at NodeJS as follows 
`import annotation from 'chartjs-plugin-annotation'`
it is causing an error like below 
annotation_bug

So What I have done to fix it is I downgrade to 2.0.0 and change **chartjs-plugin-annotation.esm** as follows 

`import { Element, defaults, Animations, Chart } from 'chart.js/auto'` and everything work fines.  (That fix doesn't work at ver 2.1.0)

annotation_bug_2

May I know what is wrong in here as I don't find similar issue mentioned or Is it only happening to me

stockiNail:
Fix #837


stockiNail:
@kurkle @LeeLenaleee not sure if this is the right way to address all these different imports from different platforms/frameworks. From your experiences, is there a list of "exports" to use for all frameworks? In my opinion it could be also helpfull to have a common approach for all extensions in chartjs org because these issues shuld happen on other plugins/controllers as well, I guess.

kurkle (commented):
node should use import/require of these automatically. This does not seem like a right way to go.

stockiNail:
> node should use import/require of these automatically. This does not seem like a right way to go.

In the repo, that I use to test, is using import, therefore esm but there is that error specify in the issue

stockiNail:
@kurkle I think the best could be to set the plugin as a module. This is also solving the issue but at the same time is creating some issue on other framework (like next). Unfortunately I don't have enough experience on that... sorry.

stockiNail:
I have to admit that I cannot understand why it's not working "type": "module" in next (v13).. I have tried to add treemap 2.3.0 in next app and it works. Then I have changed the plugin package.json file using the same of treemap but it's still not working... I need time to think about.

stockiNail:
Finally I was able to use the plugin (set to type: module) in NextJS, by dynamic import. It's still not clear to me way this is needed but I'm having a look. 
That said, I think the plugin should be marked as module, which solves the issue https://github.com/chartjs/chartjs-plugin-annotation/issues/837
Furthermore I thin we should need to extend the integration test cases with some other frameworks (like react, next, vue) in order to be sure that it's working on those platforms. Make sense?

kurkle:
For frameworks (react, vue, etc.), its essientally the bundler used (webpack/vite/etc.), that makes the difference, and it would probably be less work to make integration tests with bundlers instead of frameworks.

stockiNail:
> For frameworks (react, vue, etc.), its essientally the bundler used (webpack/vite/etc.), that makes the difference, and it would probably be less work to make integration tests with bundlers instead of frameworks.

Yes, agree. I will have a look to what we are doing in Chart.js where I think there some integration test cases but later. Before I want to understand if we need to go to type: module or not and prepare again (https://github.com/chartjs/chartjs-plugin-annotation/pull/815) a PR to go there :(

stockiNail:
@kurkle @LeeLenaleee I have done some tests and it seems working (I have used some repos I had related to some previous issues). I have used the same "config" of treemap. 

Just a doubt. I think that with these changes, the dependency with CHART.JS 4 should be mandatory. What do you think?

package.json
```diff
@@ -5,8 +5,8 @@
   "version": "2.1.2",
   "author": "Evert Timberg <evert.timberg@gmail.com>",
   "license": "MIT",
-  "main": "dist/chartjs-plugin-annotation.js",
-  "module": "dist/chartjs-plugin-annotation.esm.js",
```
kurkle, line 9: This is wrong, the main needs to point to a commonjs build and module to esm build for older node support.
The for some other platforms (I think at least vite), we need to have the `exports` section.

And from that thought, maybe we should add another integration test with different node versions.
Should also have an integration test for both flavors of node, module and non-module package.
for example: https://github.com/kurkle/chartjs-chart-matrix/tree/main/test/integration

stockiNail: OK, now it's clear!! I think we should change treemap chart as well!

> And from that thought, maybe we should add another integration test with different node versions.
> Should also have an integration test for both flavors of node, module and non-module package.
> for example: https://github.com/kurkle/chartjs-chart-matrix/tree/main/test/integration

about integration test, I think I'll need more time to have a look, changing the current PR. hopefully it's ok for you

stockiNail: @kurkle I have a question about rollup. In matrix, you create .cjs with `format: 'umd'`. Instead, in Chart.js, `format: 'commonjs'` is used. Why is there this difference?

stockiNail: Done!

kurkle: > @kurkle I have a question about rollup. In matrix, you create .cjs with `format: 'umd'`. Instead, in Chart.js, `format: 'commonjs'` is used. Why is there this difference?

umd works for commonjs and amd, so its a bit more generic. it adds a little wrapper to the file, so its able to work on both cases. not sure how much amd is used or if it is needed, but also don't see why not to support it.

kurkle: The some of the entry points are renamed so its a breaking change. I would not wait with it too long though, so maybe soon after 2.2 go for 3?

Or could just move all the 2.2 stuff to 3 instead and publish all at once.

stockiNail:
@kurkle about breaking change, do you think we could avoid it if we will create and publish the same files in NPM (`chartjs-plugin-annotation.js` is now missing)?

stockiNail:
@echo178 this PR is creating a breaking change (`chartjs-plugin.annotation.js --> chartjs-plugin.annotation.cjs`), therefore it has been planned for next major version. @kurkle @LeeLenaleee thoughts?

stockiNail:
@kurkle I have added the test cases for node module and commonjs (slightly different comparing with matrix project). I think we should take a decision if to release this PR in a minor or not. If the breaking change is only the missing `chartjs-plugin.annotation.js` dist file, I think we could create it as well (compatibility with the past). About minimum Chart.js dependencies, I think we had already done it with version1 where we changed that in a minor. 
After the decision, I think we should bump to version 2.2.0 (with or without this PR).
What do you think?

package.json
stockiNail: > The some of the entry points are renamed so its a breaking change. I would not wait with it too long though, so maybe soon after 2.2 go for 3?
> 
> Or could just move all the 2.2 stuff to 3 instead and publish all at once.

I have changed rollup config to create `chartjs-plugin-annotation.js` as well. From user perspective, I don't see breaking changes related to the renamed file now. 
I would like to go to next version anyway, including it. 
@kurkle just a hint: do you think it's better to go to version 3 or publish 2.2 with this one (remove breaking change, if there isn't)? For me no problem, just a matter to change the bump PR and some milestones, already assigned, before publishing.

stockiNail:
> The some of the entry points are renamed so its a breaking change. I would not wait with it too long though, so maybe soon after 2.2 go for 3?
> 
> Or could just move all the 2.2 stuff to 3 instead and publish all at once.

I have changed rollup config to create `chartjs-plugin-annotation.js` as well. From user perspective, I don't see breaking changes related to the renamed file now. 
I would like to go to next version anyway, including it. 
@kurkle just a hint: do you think it's better to go to version 3 or publish 2.2 with this one (remove breaking change, if there isn't)? For me no problem, just a matter to change the bump PR and some milestones, already assigned, before publishing.

donmccurdy:
Just a drive-by comment to say that I've been patching my chartjs-plugin-annotation module locally to use `"type": "module"`, and that fixed build errors in my build (SvelteKit + Vite) here. Looking forward to using the next release, and let me know if any help is needed.

I'm not aware that the `"type": "module"` and `.js` → `.cjs` updates would be a breaking change for package consumers using npm and bundlers. For consumers using CDNs it may be a breaking change, though.

rollup.config.js
```diff
@@ -38,6 +41,24 @@ module.exports = [
     },
     external
   },
+  // js
```
kurkle, line 44: this is not needed anymore, right?

stockiNail: yes @kurkle, I had added if we wanted to "mitigate" the break... I'll remove it. No hurry, right?  it needs major version, correct?

stockiNail: Removed it

kurkle (approved):

PR #846, opened 2023-02-04
Enable opacity option to the labels when the content is an image or canvas

### LINKED ISSUE #844: Image Opacity in LabelAnnotation
Hi all, I could not find a way to set the opacity of an image when using LabelAnnotation. When I have to use text, I can change the opacity by setting the opacity in the hex code in the `color` option.

So, is there an option to set the opacity?

stockiNail:
Fix #844
This PR is enabling the `opacity` option to the labels.
The option is overriding what is set by the user in the style of the element (image or canvas). If not set, the plugin is using the opacity of the style or 1 if not set.


LeeLenaleee (approved):

stockiNail:
Thank you @LeeLenaleee

PR #858, opened 2023-02-26
Remove label node from point annotations from init samples

stockiNail:
Remove label node from point annotations, because not managed by poitns, from init samples.


LeeLenaleee (approved):

stockiNail:
Thank youuu @LeeLenaleee

PR #859, opened 2023-02-27
Set maximum width to stroke and fill text when invoked to draw labels

stockiNail:
This PR is setting the max width when [`strokeText`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/strokeText#syntax) and [`fillText`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fillText#syntax) methods  of are invoked to draw the labels.

This enables a simple appearance animation on the labels based on the width of annotation element.

- [ ] wait for https://github.com/chartjs/chartjs-plugin-annotation/pull/825


stockiNail:
This could be helpful on animation but it could create some limitations when you need to draw the label, ignoring the box size. Better do not implement.

PR #860, opened 2023-02-28
Fix value stored in element options for indexable option

stockiNail:
The PR https://github.com/chartjs/chartjs-plugin-annotation/pull/801 introduced indexable options (`font` and `color`).

The current options resolution is checking if the options is indexable and if true, it stores the value of the options.

https://github.com/chartjs/chartjs-plugin-annotation/blob/bc38d6bfad27a052c199604e89d1338c07681d3f/src/elements.js#L134

In this way, it stores a `Proxy` object (or an array of `Proxy`) in element options, something that probably you don't want.

This PR is storing "plain" objects as `element.options`.

### DIFF
--- a/src/elements.js
+++ b/src/elements.js
@@ -1,5 +1,5 @@
 import {Animations} from 'chart.js';
-import {isObject, defined} from 'chart.js/helpers';
+import {isObject, isArray, defined} from 'chart.js/helpers';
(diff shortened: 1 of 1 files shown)


kurkle (approved):

stockiNail:
Thank youu @kurkle

PR #863, opened 2023-03-08
Move default init animation logic to each annotation element

stockiNail:


stockiNail:
Changing the signature of init callback from types stand point, I think it is a breaking change.
@kurkle do you think we could add to version 3 if the code it is ok for you, of course?

kurkle (approved):

PR #867, opened 2023-03-15
Fix init option type setting it as optional

### LINKED ISSUE #866: Issue with AnnotationOptions type in 2.2.0 version
With the latest version of AnnotationOptions type the library is requiring an 'init' function property. This function seems to be part of an internal class of some sort when creating a Box type annotation.

This same issue is not happening in earlier versions so i am stuck at version 2.1.2 until it is fixed.

stockiNail:
Fix #866


stockiNail:
@kurkle @LeeLenaleee if the PR is ok for you and then you will approve it, we could release a bug fixing release. What do you think? Too early?

LeeLenaleee (approved):

stockiNail:
Thank you @LeeLenaleee and apologize for the bug. I haven't recognized it due to my poor knowledge on TS :(.

kurkle:
I would suggest releasing bug fixes asap. There are plenty of version numbers to use, so no need to collect weeks or months of work to one (and keep the bug(s) unfixed meanwhile).

stockiNail:
> I would suggest releasing bug fixes asap. There are plenty of version numbers to use, so no need to collect weeks or months of work to one (and keep the bug(s) unfixed meanwhile).

Got it !! (in fact, I had already prepared the bump PR ;)). I'll not ask anymore!

stockiNail:
@kurkle do you want to release in the next days a version 3 with only PR https://github.com/chartjs/chartjs-plugin-annotation/pull/838 to fix the type=module issue?

kurkle:
I don't have any objections to that, sounds good.

PR #877, opened 2023-04-20
Fix compile errors when using multiple plugins

### LINKED ISSUE #854: Error TS2567 when using this plugin in version 2.1.2
When I install this plugin and import it I get an error:
```
Error: node_modules/chartjs-plugin-annotation/types/index.d.ts:6:13 - error TS2428: All declarations of 'PluginOptionsByType' must have identical type parameters.

6   interface PluginOptionsByType<TType extends ChartType> {
              ~~~~~~~~~~~~~~~~~~~
```
I am using chartjs 4.2.1.
When using version 2.0.1 of this plugin everything works. However I dont think thats a proper solution.

SebastiaanSafeguard:
Closes #854

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -3,7 +3,7 @@ import { AnnotationPluginOptions, BoxAnnotationOptions, EllipseAnnotationOptions
 
 declare module 'chart.js' {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  interface PluginOptionsByType<TType extends ChartType> {
+  interface PluginOptionsByType<TType extends ChartType = ChartType> {
     annotation: AnnotationPluginOptions;
   }
 


stockiNail (approved):

rfrancois:
Hi @stockiNail ,

I tried version 3.0.0 and I have a new error : 
` Object literal may only specify known properties, and 'annotation' does not exist in type '_DeepPartialObject<PluginOptionsByType<keyof ChartTypeRegistry, keyof ChartTypeRegistry>>'.`

With this code :

```
import { Chart } from "chart.js/auto";
import * as moment from 'moment';
import zoomPlugin from 'chartjs-plugin-zoom';
import annotationPlugin from 'chartjs-plugin-annotation';
import 'chartjs-adapter-moment';

Chart.register(annotationPlugin)

...

new Chart(this._chartRef.nativeElement, {
  type: 'bar',
  plugins: [zoomPlugin],
  data: {
	datasets: datasets
  },
  options: {
	responsive: true,
	maintainAspectRatio: false,
	plugins: {
	  zoom: {
		zoom: {
		  drag: {
			enabled: true
		  },
		  mode: 'x',
		},
	  },
	  annotation: { } // If I comment this line, error disappears. If I put content inside this object to draw my annotations, still have the error
	}
  },
})
```

Any suggestion ?

stockiNail:
@rfrancois created an issue: https://github.com/chartjs/chartjs-plugin-annotation/issues/886
I had a quick look and I was able to reproduce it. Also here, it doesn't depend on annotation plugin but unfortunately there is, in my opinion, something not well defined (at types level) in Chart.js. 
I say that because in my test case, I have the issue on zoom instead of annotation (depends on import order).

PR #885, opened 2023-05-11
Fix documentation after setting plugin as module

stockiNail:
This Pr should fix the issue related to the documentation where the chart instances are not available because the chart.js elements are not registered.

The issue sounds to be related to `vuepress-theme-chartjs` and how imports are managed. 
Not clear to me why (apologize) but adding the import names where missing, the app built by webpack/vue seems to have the registration and is importing everything as it should.

### DIFF
--- a/docs/.vuepress/config.ts
+++ b/docs/.vuepress/config.ts
@@ -101,8 +101,8 @@ export default defineConfig({
     docsDir: 'docs',
     chart: {
       imports: [
-        ['scripts/register.js'],
-        ['scripts/defaults.js'],
+        ['scripts/register.js', 'Register'],
+        ['scripts/defaults.js', 'Defaults'],
         ['scripts/utils.js', 'Utils'],
       ]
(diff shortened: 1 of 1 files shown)


stockiNail:
@kurkle @LeeLenaleee let me know what you think. Not sure why this update should solve the issue but weirdly the vue app JS is not built correctly if we omit the names of the imports.

stockiNail:
@kurkle @LeeLenaleee I didn't relate the PR to the issue because I would like to see how it works on GHP before to close the issue.

LeeLenaleee (approved):

kurkle (approved):

stockiNail:
@kurkle @LeeLenaleee I can confirm you that the development version on doc (master) is now rendering the charts.... The problem is that version 3.0.0 is broken :(
Let me know if you think we can fix in same way, otherwise we could release version 3.0.1.

PR #889, opened 2023-05-17
Enable drawTime option as integer to relate the annotation drawing to a specific dataset drawing

### LINKED ISSUE #887: Draw line annotation on top of first dataset but under second dataset
Currently using the drawTime: beforeDatasetsDraw option to ensure the first dataset is drawn on top of the line annotation (I'm using plugin to draw a y = 0 line).

I'd like to add a second dataset that gets drawn under the line annotation, the use case is that I'm using the fill: 0 option for this dataset so the line annotation should take preference.

The z index order ASC should be:
dataset #2 (using fill up to dataset #1)
line annotation (y = 0)
dataset #1 (not filled, just a line)

Is this possible? I assume the z index option is only in relation to other annotations and nothing to do with datasets.
I'm using chart.js v4.3.0 and v2.2.1 of plugin.

stockiNail:
Fix #887 

This PR enables the feature to draw an annotation before a specific dataset.
This is implemented setting `drawTime` option as integer which is representing the dataset index. 
The annotation will draw always before the dataset specified by index. IF the index is not consistent (doesn't represent a valid dataset index) the annotation is not drawn.

drawTime

### TODO

- [x] test cases
- [x] types
- [x] documentation


LeeLenaleee (dismissed):

stockiNail:
@LeeLenaleee I cannot merge because the checks are not available anymore and therefore the button is disable. I can switch to draft and again to ready for review in order to execute again the GHAs.

LeeLenaleee (approved):

PR #894, opened 2023-05-25
Remove default type from module augmentation of PluginOptionsByType

### LINKED ISSUE #886: Object literal may only specify known properties, and 'annotation' does not exist in specified type
I tried version 3.0.0 and I have a new error : 
` Object literal may only specify known properties, and 'annotation' does not exist in type '_DeepPartialObject<PluginOptionsByType<keyof ChartTypeRegistry, keyof ChartTypeRegistry>>'.`

With this code :

```
import { Chart } from "chart.js/auto";
import * as moment from 'moment';
import zoomPlugin from 'chartjs-plugin-zoom';
import annotationPlugin from 'chartjs-plugin-annotation';
import 'chartjs-adapter-moment';

Chart.register(annotationPlugin)

...

new Chart(chartRef.nativeElement, {
  type: 'bar',
  plugins: [zoomPlugin],
  data: {
	datasets: datasets
  },
  options: {
	responsive: true,
	maintainAspectRatio: false,
	plugins: {
	  zoom: {
		zoom: {
		  drag: {
			enabled: true
		  },
		  mode: 'x',
		},
	  },
	  annotation: { } // If I comment this line, error disappears. If I put content inside this object to draw my annotations, still have the error
	}
  },
})
```

My versions :
    "chart.js": "^4.3.0",
    "chartjs-adapter-moment": "^1.0.1",
    "chartjs-plugin-annotation": "^3.0.0",
    "chartjs-plugin-zoom": "^2.0.1",

If I remove zoom plugin, it works though. But I would like both plugins.
Any suggestion ?

_Originally posted by @rfrancois in https://github.com/chartjs/chartjs-plugin-annotation/issues/877#issuecomment-1543957021_

stockiNail:
Fix #886 revert #877 because the the root cause of the bug was different and fixed in CHART.JS by https://github.com/chartjs/Chart.js/pull/11309

Waiting for new CHART.JS version here the mentioned above PR will be published.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -3,7 +3,7 @@ import { AnnotationPluginOptions, BoxAnnotationOptions, EllipseAnnotationOptions
 
 declare module 'chart.js' {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  interface PluginOptionsByType<TType extends ChartType = ChartType> {
+  interface PluginOptionsByType<TType extends ChartType> {
     annotation: AnnotationPluginOptions;
   }
 


types/index.d.ts
```diff
@@ -3,7 +3,7 @@ import { AnnotationPluginOptions, BoxAnnotationOptions, EllipseAnnotationOptions
 
 declare module 'chart.js' {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  interface PluginOptionsByType<TType extends ChartType = ChartType> {
+  interface PluginOptionsByType<TType extends ChartType> {
```
LeeLenaleee, line 6: I think this is now ready to be merged, since this is the way it is noted in the latest chart.js version:  
https://github.com/chartjs/Chart.js/blob/dd554e0d87fc525c2a4efcecead9cc3c280936f4/src/types/index.d.ts#L2936

stockiNail: switched ready for review

LeeLenaleee (approved):

PR #916, opened 2023-11-11
Replace chart.min.js by chart.umd.js

JimmyGouraud:
There is no `chart.min.js` file when you `npm install chart.js`. I supposed it has been replaced by `chart.umd.js` as indicated in the [Chart.js integration documentation](https://www.chartjs.org/docs/latest/getting-started/integration.html).

### DIFF
--- a/docs/guide/integration.md
+++ b/docs/guide/integration.md
@@ -7,7 +7,7 @@ title: Integration
 ## Script Tag
 
 ```html
-<script src="path/to/chartjs/dist/chart.min.js"></script>
+<script src="path/to/chartjs/dist/chart.umd.js"></script>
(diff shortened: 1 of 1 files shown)


stockiNail (commented):
I would suggest to user `chart.umd.min.js`

JimmyGouraud:
But there is no `chart.umd.min.js` available with `npm install chart.js`.
image

stockiNail:
Ok, accessing to CDN by script tag, you can use also umd.min. But that's ok.

stockiNail (approved):

stockiNail:
Thanks @JimmyGouraud

Issue #938, opened 2024-08-15
TypeError: lineCap must be a string


mlohbihler:
I am using chart.js in node and using 'skia-canvas' to raise windows and provide a canvas / graphics context. This context appears to be pickier about `lineCap` and `lineJoin` properties, which cannot be set to undefined or null. Happily this problem is easily fixed by adding their 'annotation' equivalents (i.e. `borderCapStyle`, and `borderJoinStyle`) to the default options of certain structures. So far i have only encountered the errors with `LineAnnotation.defaults` and `arrowHeadsDefaults`. Adding the following lines to each is the fix:

```
  borderCapStyle: 'butt',
  borderJoinStyle: 'miter',
```

I can create a PR for this if that is easier, but i noted that there are currently 12 outstanding PRs already, some going back years.

mlohbihler:
I've just confirmed that the same defaults are also needed for `EllipseAnnotation.defaults` and `PointAnnotation.defaults`

stockiNail:
@mlohbihler feel free to open a PR anyway. I am the author of 9 on 12 PR still pending and hopefully they will be managed. 

Pay attention to options fallback setting the defaults.

mlohbihler:
@stockiNail sounds good. For completeness, here is a code snip that i just run with `node` that reproduces the problem. It's mostly pulled from the annotations sample code.

```
import { Window } from 'skia-canvas'
import { Chart, registerables } from 'chart.js'
import annotationPlugin from 'chartjs-plugin-annotation'

Chart.register(...registerables, annotationPlugin)

let win = new Window(1000, 1000)
win.once('draw', e => {
  const ctx = e.target.canvas.getContext('2d')
  new Chart(ctx, {
    type: 'line',
    data: {
      labels: ['January', 'February', 'March', 'April', 'May', 'June', 'July'],
      datasets: [
        {
          label: 'My First Dataset',
          data: [65, 59, 80, 81, 56, 55, 40],
          fill: false,
          borderColor: 'rgb(75, 192, 192)',
          tension: 0.1,
        },
      ],
    },
    options: {
      plugins: {
        annotation: {
          annotations: {
            ellipse1: {
              type: 'ellipse',
              xMin: 1,
              xMax: 2,
              yMin: 50,
              yMax: 70,
              backgroundColor: 'rgba(255, 99, 132, 0.25)',
            },
            line1: {
              type: 'line',
              yMin: 60,
              yMax: 60,
[...]

PR #902, opened 2023-07-11
Add hitTolerance option to all annotations

stockiNail:
Fix  #900

This PR is add new options, `hitTolerance`, which is a number and represents the amount of pixels to add for interacting on annotation elements.

### TODO

- [x] test case
- [x] types
- [x] documentation

PR #939, opened 2024-08-16
Change borderCapStyle and borderJoinStyle defaults to be compatible with SKIA canvas

mlohbihler:
Setting these defaults prevents errors when using annotations in the skia canvas context. See Issue for more details: https://github.com/chartjs/chartjs-plugin-annotation/issues/938


LeeLenaleee (approved):

LeeLenaleee:
Change LGTM, but it seems the CI fails

mlohbihler:
> Change LGTM, but it seems the CI fails

I'll have a looko.

mlohbihler:
I think i need help with this. `test-integration` works locally for me, but i'm on mac. Not sure why it would be failing on ubuntu.

mlohbihler:
> I think i need help with this. `test-integration` works locally for me, but i'm on mac. Not sure why it would be failing on ubuntu.
> 
> Edit: Actually, it looks like `test-integration` succeeded, but then the process failed with no further messages.

Actually, i'm getting the same build failure on the master branch. Can you check?

LeeLenaleee:
I will look into it this weekend

src/types/line.js
```diff
@@ -107,9 +107,11 @@ LineAnnotation.id = 'lineAnnotation';
 const arrowHeadsDefaults = {
   backgroundColor: undefined,
   backgroundShadowColor: undefined,
+  borderCapStyle: 'butt',
```
stockiNail, line 110: Setting the default here, borderCapstyle doesn't fallback to the line options.

https://github.com/chartjs/chartjs-plugin-annotation/blob/e94475cc299fd77dd6c008b7f4bfd4a985df3a6d/src/types/line.js#L198-L208

src/types/line.js
```diff
@@ -107,9 +107,11 @@ LineAnnotation.id = 'lineAnnotation';
...
   borderColor: undefined,
   borderDash: undefined,
   borderDashOffset: undefined,
+  borderJoinStyle: 'miter',
```
stockiNail, line 114: Setting the default here, borderJoinStyle doesn't fallback to the line options.

https://github.com/chartjs/chartjs-plugin-annotation/blob/e94475cc299fd77dd6c008b7f4bfd4a985df3a6d/src/types/line.js#L198-L208

stockiNail (commented):
@mlohbihler @LeeLenaleee has written in the issue, we need to take care about options fallback.

As far as I remember, with this updates we are creating a breaking change because we are stopping the fallback of start and end config to arrowHeads configuration and not to the line options, as is today.

src/types/line.js
mlohbihler: Thanks @stockiNail . Is this showing somewhere, like in a build output or something? I can't find a message describing why the build is failing.

stockiNail: @mlohbihler my thoughts are not related to the failed build. as far as I remember, this could happen because the browser versions, used in Karma, are changed and they are managed the canvas differently comparing with the past. You need to run Karma in debug and check the differences between the rendering by browser and the fixture saved in the repo. Usually we need to increase the tolerance.

stockiNail:
@mlohbihler LGTM. May be you can try the branch against SKIA canvas to check if it is working as expected.

stockiNail:
@mlohbihler the failed test has ben fixed in https://github.com/chartjs/chartjs-plugin-annotation/pull/902. Therefore after merging of that, we could re-run the failed job

mlohbihler:
> @mlohbihler the failed test has been fixed in #902. Therefore after merging of that and rebase this branch, we could re-run the failed job

@stockiNail nice. Thanks for this.

stockiNail:
@mlohbihler if you will rebase and push again, we can go on. Take your time. Thanks a lot!

mlohbihler:
> @mlohbihler if you will rebase and push again, we can go on. Take your time. Thanks a lot!

Build checks are passing now, but i guess with the (net zero) playing around i did i lost my approval. Re-requested.

stockiNail (changes requested):
Apologize if I hadn't time before to have a look. 

The changes are adding options not used by those annotations therefore I would prefer to change only the `setBorderStyle` (reverting the current changes of PR) as following:

```javascript
export function setBorderStyle(ctx, options) {
  if (options && options.borderWidth) {
    ctx.lineCap = options.borderCapStyle || 'butt';
    ctx.setLineDash(options.borderDash);
    ctx.lineDashOffset = options.borderDashOffset;
    ctx.lineJoin = options.borderJoinStyle  || 'miter';
    ctx.lineWidth = options.borderWidth;
    ctx.strokeStyle = options.borderColor;
    return true;
  }
}
```

Let me know what you think

mlohbihler:
> Apologize if I hadn't time before to have a look.
> 
> The changes are adding options not used by those annotations therefore I would prefer to change only the `setBorderStyle` (reverting the current changes of PR) as following:
> 
> ```js
> export function setBorderStyle(ctx, options) {
>   if (options && options.borderWidth) {
>     ctx.lineCap = options.borderCapStyle || 'butt';
>     ctx.setLineDash(options.borderDash);
>     ctx.lineDashOffset = options.borderDashOffset;
>     ctx.lineJoin = options.borderJoinStyle  || 'miter';
>     ctx.lineWidth = options.borderWidth;
>     ctx.strokeStyle = options.borderColor;
>     return true;
>   }
> }
> ```
> 
> Let me know what you think

That change seems to work for me. Note that there are other instances of these properties getting set. Not sure if they can be removed too. E.g. see https://github.com/mlohbihler/chartjs-plugin-annotation/blob/0bce23d13a8db9aa9c58a80e5fc2929b4dfbab71/src/types/box.js#L40

stockiNail:
> That change seems to work for me. Note that there are other instances of these properties getting set. Not sure if they can be removed too. E.g. see https://github.com/mlohbihler/chartjs-plugin-annotation/blob/0bce23d13a8db9aa9c58a80e5fc2929b4dfbab71/src/types/box.js#L40


@mlohbihler the joinStyle and capstyle are used for setting on canvas only in the point that you have changed.
The other istances are only in the options defintions of the annotations in order to enable the options settings from the user.

PR #942, opened 2024-09-19
Fix init option (initial animation) type

fteamkr:
Fix "Initial animation" option (init) type by document (https://www.chartjs.org/chartjs-plugin-annotation/latest/guide/configuration.html#initial-animation).

And init's return type might be Partial.

I created this PR not a issue, because it might be a simple fix. Thank you!


stockiNail:
@fteamkr thank you! Can you rebase the branch?

stockiNail (approved):

PR #959, opened 2024-12-12
Add doughnutLabel type to AnnotationTypeRegistry

Tmccleerey:


stockiNail:
@Tmccleerey thank you!

stockiNail:
@kurkle @LeeLenaleee there is a code climate issue related "similar" code. In my opinion, it should be ignored but I cannot do it and your support is needed.

LeeLenaleee:
Its an optional step so it is not blocking for this change. I am not sure if it will still show up on following pr's. If so I will try and take a look at it

kurkle:
Looks like I don't have the permission to add ignores there anymore. But as @LeeLenaleee said, codeclimate is informative, not blocking. So you can do the ignoring while pressing merge 😄

stockiNail:
uhm... a bit weird because before @LeeLenaleee approval the merge button was disabled to me. Next time it will happen, I'll check more carefully.
THANKS GUYS!!

PR #972, opened 2025-04-16
Add support for per-element clip attributes

### LINKED ISSUE #874: Configurable clip for each annotation
### Feature Proposal

Would it be possible to have the clip property per annotation so we can control it independently ?
What would be a workaround for this ?
Thanks in advance.

### Possible Implementation

_No response_

joshkel:
Fixes #874

I see in #874 that there are potential backwards compatibility concerns. I'm not aware of any in this code as implemented, but I still feel like I don't necessarily understand Chart.js's options implementation, so I may be missing something.

If this feature and approach seems desirable, I'll update tests and docs. Thank you.


stockiNail:
@joshkel thank you very much! I don't recall why we had some doubts to implement `clip` option at element level (my memory is not so persistent but it could be related to performance issue?!?).
For me it makes sense.

Nevertheless, your PR sounds good even if it would be perfect to move the `clip` option in the annotation plugin to the `common` node.
Why? The `common` node is the fallback node for element defaults. Which means that `if` statements in your code will be useless because we could only check `item.element.options.clip`. If not set in the element, it will have the value of the `common`.

That was the idea at that time when we talked about that. But it is also clear, my proposal would be a breaking change and therefore it will need a major version.

joshkel:
The purpose of the `if` checks is to limit performance impact: if a portion of the list of annotation elements all has the same `clip` option, then the plugin can avoid continually applying then clearing clipping settings.

Regarding whether to put the option in `common`: I'm afraid I don't understand why adding it there would be a breaking change. But I'm happy to go whichever route you prefer.

stockiNail:
> Regarding whether to put the option in `common`: I'm afraid I don't understand why adding it there would be a breaking change. But I'm happy to go whichever route you prefer.

Because I mean "moving" and not "adding", because only adding it, you will have the `clip` on the root config node which is not used.
And moving it, whoever is using `clip` should change the own implementation.

stockiNail:
Nevertheless I think we could proceed step by step, and going on with this implementation and in the new major, change it.

kurkle (approved):

PR #978, opened 2025-07-27
Rework TypeScript .d.ts handling

### LINKED ISSUE #977: TypeScript type issues
I'm working on updating a Node.js TypeScript project to use TypeScript's [`"module": "nodenext"`](https://www.typescriptlang.org/tsconfig/#module) and [`"moduleResolution": "nodenext"`](https://www.typescriptlang.org/tsconfig/#moduleResolution) and am getting several compiler errors similar to the following:

```
error TS2614: Module '"chartjs-plugin-annotation"' has no exported member 'AnnotationOptions'.
Did you mean to use 'import AnnotationOptions from "chartjs-plugin-annotation"' instead?

import type { AnnotationOptions } from 'chartjs-plugin-annotation';
              ~~~~~~~~~~~~~~~~~
```

From what I can tell, the problem is caused by the relative imports within chartjs-plugin-annotation's `types` directory. Running tsc reveals that TypeScript expects to see, e.g., a `./options` file and doesn't check for `./options.d.ts`:

```
======== Resolving module './options' from '/Users/josh/src/app/node_modules/chartjs-plugin-annotation/types/index.d.ts'. ========
Explicitly specified module resolution kind: 'NodeNext'.
Resolving in ESM mode with conditions 'import', 'types', 'node'.
Loading module as file / folder, candidate module location '/Users/josh/src/app/node_modules/chartjs-plugin-annotation/types/options', target file types: TypeScript, JavaScript, Declaration, JSON.
Directory '/Users/josh/src/app/node_modules/chartjs-plugin-annotation/types/options' does not exist, skipping all lookups in it.
======== Module name './options' was not resolved. ========
```

The [Are the Types Wrong](https://arethetypeswrong.github.io/?p=chartjs-plugin-annotation%403.1.0) tool reports these as "internal resolution errors" and also points out issues with CJS versus ESM types.

joshkel:
This reworks .d.ts handling to address the issues in #977 and reported by [Are the Types Wrong](https://arethetypeswrong.github.io/?p=chartjs-plugin-annotation%403.1.0).

Background information:

* According to the [TypeScript Handbook](https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html#im-writing-a-library):

    > If you’re using a bundler to emit your library... You must ensure that your declaration files get bundled as well. Recall the [first rule of declaration files](https://www.typescriptlang.org/docs/handbook/modules/theory.html#the-role-of-declaration-files): every declaration file represents exactly one JavaScript file.

* According to [Are the Types Wrong](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseESM.md),

    > A golden rule of declaration files is that if they represent a module—that is, if they use import or export at the top level—they must represent exactly one JavaScript file. They especially cannot represent JavaScript files of two different module formats.

* chartjs-plugin-annotation's CJS build uses a CJS default export (see [Rollup's docs](https://rollupjs.org/configuration-options/#output-exports)). Properly representing that in a `.d.ts` file requires using `export = Annotation`, but the `export =` syntax both raises challenges for exporting types alongside the default export and is incompatible with ESM. (The [TypeScript Handbook](https://www.typescriptlang.org/docs/handbook/modules/appendices/esm-cjs-interop.html) has even more background information here. See also [tshy's documentation](https://github.com/isaacs/tshy?tab=readme-ov-file#handling-default-exports), "`export default` is the bane of hybrid TypeScript modules" - although I believe that the situation is [not quite as bad as it says](https://github.com/isaacs/tshy/pull/105).)

Pulling this together for chartjs-plugin-annotation:

* Use [rollup-plugin-dts](https://www.npmjs.com/package/rollup-plugin-dts) to bundle declaration files alongside the bundled library files (as recommended by the TypeScript Handbook).
* Invoke rollup-plugin-dts twice to create separate `.d.ts` files for the ESM and CJS builds (and update package.json to reference those).
* Although a CJS file with a default export can't export additional values, it is apparently possible to export types alongside the default value, if the types are placed in a namespace with the same name as the default value.
* I had trouble finding a solution for the `export =` default issue. (I couldn't get rollup-plugin-dts to create an `export =` style definition. Changing chartjs-plugin-annotation to not use a CJS default export may be the simplest long-term solution, but it would break backward compatibility.) ~~To solve this, I added @rollup/plugin-replace to apply the necessary namespace and `export =` statement via string replacement.~~ _(changed - see below)_

Fixes #977

### DIFF
--- a/karma.conf.cjs
+++ b/karma.conf.cjs
@@ -18,6 +18,11 @@ module.exports = async function(karma) {
   const regex = karma.autoWatch ? /chartjs-plugin-annotation\.cjs$/ : /chartjs-plugin-annotation\.min\.js$/;
   const build = builds.filter(v => v.output.file && v.output.file.match(regex))[0];
 
+  // Even if using the unminified build, use UMD format with default exports, so
+  // we can access it as a global variable.
+  build.output.format = 'umd';
+  delete build.output.exports;
+
   if (env === 'test') {
     build.plugins = [
       json(),
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,6 @@
       "devDependencies": {
         "@rollup/plugin-json": "^6.0.0",
         "@rollup/plugin-node-resolve": "^15.0.1",
-        "@rollup/plugin-replace": "^6.0.2",
         "@rollup/plugin-terser": "^0.4.0",
         "@simonbrunel/vuepress-plugin-versions": "^0.2.0",
         "@typescript-eslint/eslint-plugin": "^5.51.0",
@@ -2027,28 +2026,6 @@
       },
       "peerDependencies": {
         "rollup": "^2.78.0||^3.0.0"
-      },
-      "peerDependenciesMeta": {
-        "rollup": {
-          "optional": true
-        }
-      }
-    },
-    "node_modules/@rollup/plugin-replace": {
-      "version": "6.0.2",
-      "resolved": "https://registry.npmjs.org/@rollup/plugin-replace/-/plugin-replace-6.0.2.tgz",
-      "integrity": "sha512-7QaYCf8bqF04dOy7w/eHmJeNExxTYwvKAmlSAH/EaWWUzbT0h5sbF6bktFoX/0F/0qwng5/dWFMyf3gzaM8DsQ==",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "@rollup/pluginutils": "^5.0.1",
-        "magic-string": "^0.30.3"
-      },
-      "engines": {
-        "node": ">=14.0.0"
-      },
-      "peerDependencies": {
-        "rollup": "^1.20.0||^2.0.0||^3.0.0||^4.0.0"
       },
       "peerDependenciesMeta": {
         "rollup": {
@@ -22423,16 +22400,6 @@
         "is-builtin-module": "^3.2.0",
         "is-module": "^1.0.0",
         "resolve": "^1.22.1"
-      }
-    },
-    "@rollup/plugin-replace": {
-      "version": "6.0.2",
-      "resolved": "https://registry.npmjs.org/@rollup/plugin-replace/-/plugin-replace-6.0.2.tgz",
-      "integrity": "sha512-7QaYCf8bqF04dOy7w/eHmJeNExxTYwvKAmlSAH/EaWWUzbT0h5sbF6bktFoX/0F/0qwng5/dWFMyf3gzaM8DsQ==",
-      "dev": true,
-      "requires": {
-        "@rollup/pluginutils": "^5.0.1",
-        "magic-string": "^0.30.3"
       }
     },
     "@rollup/plugin-terser": {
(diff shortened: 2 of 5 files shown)


joshkel:
I did some further investigation into the `export =` versus `export default` issue and realized that my attempted solution didn't work. rollup-plugin-dts's output simply isn't amenable to string replacement like I thought it was. I see a couple of potential paths forward:

1. Add other code or tooling to generate a `.d.cts` file with `export =` and a namespace: Microsoft's [API Extractor](https://api-extractor.com/) could probably be made to work with some string replacements, or I could write my own tool (which admittedly is more than I was hoping to tackle 🙂 ).
2. Give up and use a CJS-ESM-interop style default export: instead of `module.export = AnnotationPlugin`, do something like `exports.default = Annotation; Object.defineProperty(exports, '__esModule', { value: true });`. **This is a potentially breaking change**: anyone who's using the plugin in a CommonJS Node environment will now have to do `const AnnotationPlugin = require('chartjs-plugin-annotation').default` instead of `const AnnotationPlugin = require('chartjs-plugin-annotation')`, and anyone using `chartjs-plugin-annotation.cjs` as a UMD build via `<script>` tag will be similarly affected, but anyone using the `.min.js` version or using chartjs-plugin-annotation via bundler (i.e., anyone following the documentation) will be unaffected.
3. Give up on fixing the errors from Are the Types Wrong: As [its docs explain](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseExportDefault.md), "the inaccuracy of the typings is not easily observable" (since TypeScript's tsc with the `esModuleInterop` option and typical bundlers will both cover over the issue).

For the purposes of this PR, I went with approach 2, but please let me know how you'd like to proceed.

If chartjs-plugin-annotation ends up making a breaking change, I wonder if it would be best to separate the CJS and unminified UMD builds, as Chart.js itself does, to make this a little more explicit.

joshkel:
A fourth approach is to create a separate .d.ts file for the CJS version that re-exports everything. See https://github.com/chartjs/chartjs-plugin-datalabels/pull/436. chartjs-plugin-annotation defines enough data types that this is a decent amount of duplication.

kurkle:
To me, aligning with chart.js style sounds like a good option in the long run. Would not be a huge breaking change, but a breaking change anyway.

joshkel:
@kurkle , do you want me to pursue that as part of this PR?

kurkle:
@joshkel not quite sure, I'll need to catch up first. @stockiNail, @etimberg thoughts?

etimberg:
Catching up as well. My gut says that we should align with the Chart.js types even if it's breaking now

stockiNail:
Apologize for my late reply. Well, I have to admit that during last 2 years I have never touched any javascript code in deep therefore I'm a bit out of the problem. Anyway

>> To me, aligning with chart.js style sounds like a good option in the long run. Would not be a huge breaking change, but a breaking change anyway.

and 

>> we should align with the Chart.js types even if it's breaking now

I fully agree. Working in the past on chart.js, this plugin and others, I often see this differences, on packaging as well. In my opinion could make sense to align all in chartjs org even if it could be a breaking change.

